Issue #6, opened 2019-11-29
NPM Package?


adrianschubek:
Is there going to be a NPM package?

sagalbot:
If #8 is merged, you can `npm i calebporzio/project-x` to install the package from GitHub.

calebporzio:
Merged.

PR #8, opened 2019-12-01
chore: add version to package.json

sagalbot:
Adds a version key to `package.json`. This will allows users to `npm i calebporzio/project-x` without requiring the package to be published to NPM.

### Sidenote

Originally I started this branch hoping to swap `mix` for `rollup` to provide an optional ESM bundle in addition to the UMD. The ESM bundle cuts out a lot of boilerplate that UMD requires, lowering the file size quite a bit. Really useful for those that'll consume the package via npm. 

Getting the babel config working was a total nightmare though, and I eventually gave up. May take another crack at it later. Still wanted to contribute _something_, so here ya go.

---

Related #6

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,7 @@
 {
   "main": "dist/project-x.js",
   "name": "project-x",
+  "version": "0.2.0",
   "scripts": {
     "test": "jest",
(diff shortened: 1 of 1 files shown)


calebporzio:
Awesome! Thanks @sagalbot - I'd be open to that PR in the future!

PR #11, opened 2019-12-04
Add hidden to boolean attributes

tlgreg:
Wanted to toggle the `hidden` [global boolean attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden), so I added it.
Was thinking that a boolean value should just toggle any attribute, but didn't want to jump that.

### DIFF
--- a/src/component.js
+++ b/src/component.js
@@ -271,7 +271,7 @@ export default class Component {
                     }
                 })
             }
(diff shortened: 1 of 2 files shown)


calebporzio:
Yeah, maybe - this is good for now though -good move.

Thanks!

PR #16, opened 2019-12-10
Window modifier for x-on directive

mhanberg:
This allows you to install an event listener on the window object and still have access to component state in the callback

This PR is unsolicited, feel free to close if it doesn't fit your vision.


calebporzio:
I dig it, thanks!

keyurshah:
this is very helpful. thanks!

would it also be beneficial to add a way to access `document` as well? 

Seems you can access `document` by using `window.document` but the following did not work.

`x-on:document.[eventName].window`

PR #18, opened 2019-12-11
Let x-model, x-on, x-text. etc. listen to nested objects.

thormeier:
Use multi-level proxies to achieve this.

Also added tests and examples to showcase the functionality. I profiled the examples a bit and didn't recognize any memory leak caused by this.

Solves https://github.com/calebporzio/project-x/issues/14

**EDIT:** Just rebased onto master to solve conflicts.


calebporzio:
This is fantastic @thormeier, thanks!

Guessing you didn't mean to commit the `index.html` with this PR?

thormeier:
@calebporzio Thank you for the positive feedback!

I actually meant to commit it, as it's already present in the repo and I found it to be a nice show case and tool for local development. I can revert these changes, though, if you want.

calebporzio:
Gotcha, sounds good. Thanks for the top-notch contribution @thormeier!

PR #25, opened 2019-12-17
added ie11 polyfill scripts to README

keyurshah:
added instructions to README for IE11 users

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,13 @@ Add the following script to the end of your `<head>` section.
 <script src="https://cdn.jsdelivr.net/gh/calebporzio/project-x@v0.4.0/dist/project-x.min.js" defer></script>
 ```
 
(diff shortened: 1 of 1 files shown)


calebporzio:
So, I'm going to merge this for now. But I think I'd like to have a `@alpinejs/polyfills` repo that people can require in the future.

Any opinions on that?

Also, @Gummibeer, can you clarify what you mean by that?

Gummibeer:
@calebporzio I'm not this deep in JS but so far I know webpack, babel and browserlist are meant to build your JS source to a defined set of browsers including all possibly required polyfills!? :/

calebporzio:
Yeah, that's true, I get all confusled by that stuff though. Wrangling babel is so frustrating - would love an export to just swoop in and make stuff like this happen - but they really have to know what they're doing to not arbitrarily increase the package size.

PR #28, opened 2019-12-19
Support adding/removing multiple classes for an element at once

ahallhognason:
Now, instead of only being able to toggle one class with x-bind (as with the README's  example 'active' class), we can now toggle multiple classes.

Like so:
```html
<div class="border rounded py-2 px-4 mt-4" x-data="{ tab: 'foo' }">
	<ul class="flex border-b">
		<li class="mr-1" x-bind:class="{ '-mb-px': tab === 'foo' }">
			<button class="bg-white inline-block py-2 px-4 text-blue-500 hover:text-blue-800 font-semibold" x-bind:class="{ 'border-l border-t border-r rounded-t text-blue-700': tab === 'foo' }" x-on:click="tab = 'foo'">Foo</button>
		</li>
		<li class="mr-1" x-bind:class="{ '-mb-px': tab === 'bar' }">
			<button class="bg-white inline-block py-2 px-4 text-blue-500 hover:text-blue-800 font-semibold" x-bind:class="{ 'border-l border-t border-r rounded-t text-blue-700': tab === 'bar' }" x-on:click="tab = 'bar'">Bar</button>
		</li>
	</ul>

	<div x-show="tab === 'foo'">Tab Foo</div>
	<div x-show="tab === 'bar'">Tab Bar</div>
</div>
```


calebporzio:
Thanks, great addition!

PR #41, opened 2019-12-30
Add document to listen target for custom events

pfrlv:


keyurshah:
this would be useful. it might help to add documentation to the README as well, similar to PR #16  for `.window`

PR #44, opened 2020-01-03
Added support for x-html directive

ryangjchandler:
This PR simply adds support for a new `x-html` directive. There are tests to accompany the changes and the implementation is very similar to that of the `x-text` directive.

Happy to make changes to this too. There isn't any CONTRIBUTING.md file or notes so I wasn't sure whether I should also commit the dist/ version of the changes.


calebporzio:
Wow, I literally didn't have to make a single change to this PR to pull it in. Great work! Thanks!

ryangjchandler:
Not a problem at all, happy to help!

Issue #49, opened 2020-01-07
Nested components cannot access external data


SimoTod:
Hi @calebporzio, 
thanks for the amazing work so far. I was having a go with alpinejs and I noticed that, when there are nested components, the internal component cannot access the scope of the external one. 

For example,
```html
<html>
  <head>
    <script src="https://cdn.jsdelivr.net/gh/alpinejs/alpine@v1.2.0/dist/alpine.js" defer></script>
  </head>
  <body>
    <div x-data="{ foo: 'bar', foo2: 'BAR' }">
      <span x-text="foo"></span>
      <div x-data="{ foo: 'bob' }">
        <span id="s1" x-text="foo"></span>
        <span id="s2" x-text="foo2"></span>
      </div>
    </div>
  </body>
</html>
```
I would expect `span#s2` to display 'BAR' or, in alternative, i would expect to be able to reference foo2 in the internal data structure. 

I'm happy to work on a PR for this but I just wanted to check with you first in case this behaviour is expected and you do not want components to access external scopes.

Thanks, 
Simone

calebporzio:
Hey @SimoTod,

Great question.

This is definitely something that should be on the radar. Inter-component communication is a common need.

I suppose I've wanted to nail the core before adding this type of feature because it will increase the complexity of the project and therefore the issues, etc...

I would love to keep this conversation going though and start discussing potential APIs for this.

Here are a couple routes we could go off the top of my head:
A) A "prop" system
```
<div x-data="{ foo: 'bar' }">
  <div x-data="{}" x-props="{ baz: foo}"></div>
</div>
```
or something like that
B) Accessing the parent scope from a magic `$parent` object or something:
```
<div x-data="{ foo: 'bar' }">
  <div x-data="{}">
    <span x-text="$parent.foo"></span>
  </div>
</div>
```
C) Accessing data through simply passing down scope like you mentioned:
```
<div x-data="{ foo: 'bar' }">
  <div x-data="{}">
    <span x-text="foo"></span>
  </div>
</div>
```

Let's keep the conversation going. Thanks!

SimoTod:
Good points @calebporzio.

Option A feels similar to the approach Vue and other frameworks take but, if I didn't have any experience, it would be an additional learning friction. It also forces a dev to pass down the variables if there are multiple nested levels, polluting the DOM. It probably makes easier to implement the "reactivity" part, though. 

[...]

Issue #21, opened 2019-12-12
Using the parent variable after nested components


keyurshah:
I'm have a component and then declaring nested components inside.

Once I use a data variable after a new nested component is declared, the variable doesn't work. Here's an example:

https://codepen.io/kjshah/pen/df81e5331e12140ac50ecb342dedfb61

```
<div x-data="values()">
  <div class="font-semibold">This Doesn't Work</div>
  
  <div x-data="{name: 'blue widget'}">
    Name: <span x-text='name'></span>
  </div>
  
  <hr>
  (total should show here)
  Total: <span x-text="total"></span>
</div>
  
--------------------------------------------------
  
<div x-data="values()">
  <div class="font-semibold">This Works</div>
  
  Total: <span x-text="total"></span>
  
  <div x-data="{name: 'blue widget'}">
    Name: <span x-text='name'></span>
  </div>
</div>
```

```
<script>
function values() {
  return {
    total: 1,
  }
}
</script>
```

calebporzio:
Good catch, fixed in v0.4.3

rosswintle:
Hi folks,

I've been playing with Alpine this evening for the first time and I THINK this is still an issue.

For starters, it's not documented whether referring to a variable in a parent scope is even possible, so I assumed it was. This should be made clear in the readme. It's entirely possible that referring to data declared higher up in the DOM is not allowed. It's also entirely possible that I just couldn't find where in the readme it tells me this.

If it is possible then it's still an issue as far as I can tell.

Example:

I have a collection of buttons and I want to apply a class to a button when that button is clicked.

Here's a simple example coded with button values hard-coded: https://codepen.io/magicroundabout/pen/jOEGXrd

This works and is fine. But I figured I could set the button ID as data on the button and use it in the directives, like this: https://codepen.io/magicroundabout/pen/KKwXbME

But in the second example here the button can't see either the data declared higher up in the DOM or its own data!

Here's a CodePen with both examples together: https://codepen.io/magicroundabout/pen/qBEPLqq

Is this a bug? Or a misunderstanding?

PR #55, opened 2020-01-08
Inter-component communication

SimoTod:
This PR will allow nested components to access properties defined in a parent scope via a magic variable `$parent`. It support multiple levels of nesting and triggers updates accordingly on any of the levels involved.

It fixes https://github.com/alpinejs/alpine/issues/49 and the last comment mentioned on https://github.com/alpinejs/alpine/issues/21.

Example: https://codepen.io/SimoTod/pen/OJPzrym


SimoTod:
Closing for now as it needs further consideration

PR #59, opened 2020-01-08
Update correct handlers

c0b41:


Mudassar045:
I think you didn't read between the lines. Try to point this heading: 
https://github.com/alpinejs/alpine#x-on **OR** `Note: You are free to use the shorter "@" syntax: @click="..."`
Moreover, check out the source code snippet below
```
export function replaceAtAndColonWithStandardSyntax(name) {
    if (name.startsWith('@')) {
        return name.replace('@', 'x-on:')
    } else if (name.startsWith(':')) {
        return name.replace(':', 'x-bind:')
    }

    return name
}
```

c0b41:
well i try short hand method and not working so?

SimoTod:
@c0b41 This feature was released with v1.5.0. 
You wrote your PR while the new version was being released so probably you were still using 1.4.0. 

If you uses npm, try to update the library with `npm update alpinejs`.
If you uses the CDN, make sure the link is up to date => `<script src="https://cdn.jsdelivr.net/gh/alpinejs/alpine@v1.6.0/dist/alpine.js" defer></script>`.

After the update, it should work for you as well. 👍

c0b41:
@SimoTod thanks this might be problem 👍

PR #62, opened 2020-01-09
update README.md

erikscheepers:
I'm not a native English speaker, but I think "retreive" should be "retrieve".


calebporzio:
Indeed, thanks!

PR #65, opened 2020-01-09
allow colons in event names

keyurshah:
allow for events that are namespaced with a colon 

unpoly emits the example event:  `up:fragment:inserted`


calebporzio:
Hmm, this wouldn't be "too" hard to allow for I don't think.

Might have to just message some regex. If it's as simple as changing a couple of lines of regex, and if you provide a test, I'd be down for a PR

calebporzio:
Lol, thought this was an issue - PERFECT! Thanks

keyurshah:
awesome! thanks so much @calebporzio keep up the amazing work!!

Issue #67, opened 2020-01-10
Reactivity lost on v1.5 → v1.6 when binding to a function returning boolean


johndoe75:
This was working in version smaller or equal to 1.5.0 and since 1.6.0 it's no longer working:

```html
<div x-data="test(true)">

  <button @click="toggle()">Press me</button>

  <input type="text" x-bind:value="isSet" />

  <input type="text" x-bind:value="getIsSet()" />

</div>
```

```javascript
function test(status) {

  return {
    isSet: !!status,
    toggle() {
      this.isSet = !this.isSet
    },
    getIsSet() {
      return this.isSet
    }
  }

}
```

Reading the `isSet` value from the data-set is reactive and updated correct.  But binding the return value from `getIsSet()`, it's not updated – even though is initialized correct.  When running the same code with v1.5, it's working as expected.

You can test it in this [fiddle](https://jsfiddle.net/johndoe75/7wpae53j/4/).

jreviews:
I think this may be a duplicate of #66

johndoe75:
You're right.  Looks like 66 was published while I was working on the fiddle. :)

calebporzio:
Should be good now in 1.8.0

PR #68, opened 2020-01-11
Restore reactivity

### LINKED ISSUE #66: x-show with function not evaluating return value
When using the dropdown component example from the docs, `x-show="isOpen()"` doesn't update the DOM, even though the isOpen function does run. However, using `x-show="show"` works correctly.

This works on previous versions, but not the latest one.

Example here https://codepen.io/jreviews/pen/povVRag

SimoTod:
Fixes #66 and #67 .

When printing `this` from a function included in x-data, its context is already set to the Proxy object.
Binding the context on lines 20-25 seems to break the reactivity. We should be able to safely remove those lines since, when printing `this` from a function included in x-data, the context is already set to the Proxy object.

I've also added some additional tests to cover this user case.


calebporzio:
Thanks for this. I made a pretty big core refactor that took away these issues as well. Good tests - thanks for those!

Issue #72, opened 2020-01-11
Stopped working on IE11


keyurshah:
Hello,

It seems that after the transition to rollup and newer features, alpinejs is not working in IE11 with the current polyfills. 

I am getting the error:

```
Expected ')'
```

at this line in `utils.js`

```
export function walkSkippingNestedComponents(el, callback, isRoot = true) {
```

I then tested on a branch after the transition to rollup (1.1.3) and got it to work by changing babel.config.js to the following

```
module.exports = {
  presets: [
    [
      "@babel/preset-env",
      {
        targets: {
          browsers: "> 0.5%, ie >= 11"
        },
        modules: false,
        spec: true,
        useBuiltIns: "usage",
        forceAllTransforms: true,
        corejs: {
          version: 3,
          proposals: false
        }
      }
    ]
  ]
};
```

this increases the build size, but gets it to work for testing purposes.

I then tested the same configuration on the current build (1.7) and I now get the following error

```
Uncaught (in promise) TypeError: Cannot create property for a non-extensible object
```

Seems like an error with the proxy polyfill when researching that error. 

```
<script src="https://cdn.jsdelivr.net/npm/proxy-polyfill@0.3.0/proxy.min.js"></script>
```

I also tested by adding Laravel Mix back and the same object error still occurs. 

Wondering if anyone else has gotten a recent build to work with IE11? 

Thanks.

keyurshah:
it seems that the proxy polyfill issue happened with commit https://github.com/alpinejs/alpine/commit/1ddb8941b6790c8e5639884cbf674497cba6cd71

Perhaps with the expansion of the `$refs` functionality

---

If the issue with `babel.config.js` is true as well, then not sure how we can output a separate file based on ie11 as a target for `babel.config.js`. The file size doubles in size to 6.68 KB gzipped with the change in target to 

```
        targets: {
          browsers: "> 0.5%, ie >= 11"
        },
```

TomS-:
I'm also struggling with this issue. I hope a solution get's worked out as I support IE11.

SimoTod:
@keyurshah Not sure if it helps (I can't try myself because I'm a Mac user) but I believe it's because the polyfill doesn't allow the definition of new properties for the Proxy once created.

In their github page, they say: `The polyfill supports just a limited number of proxy 'traps'. It also works by calling seal on the object passed to Proxy. This means that the properties you want to proxy must be known at creation time.`

[...]

PR #76, opened 2020-01-11
IE11 Issues

keyurshah:
Just wanted to make my testing branch available for issue  #72 

`Uncaught (in promise) TypeError: Cannot create property for a non-extensible object`

in case it helps in anyway. It's not to merge into master. My apologies if there's a better way to do this.

It contains a modified `rollup.config.js` which creates `alpine-ie11.js` 

It also contains `index-ie11.html` for testing. 

Thank You


keyurshah:
This is outdated from the current discussion on the issue so will close out

PR #82, opened 2020-01-13
Fix duplicate callbacks in transitionOut() and transitionIn()

### LINKED ISSUE #81: catch error while using x-if
toggle x-if expression the console will receive 
```
alpine.js:1 Uncaught TypeError: Cannot read property 'isSameNode' of null
    at MutationObserver.observe.childList (alpine.js:1)
MutationObserver.observe.childList @ alpine.js:1
alpine.js:1 Uncaught TypeError: Cannot read property 'remove' of null
    at alpine.js:1
    at alpine.js:1
```
Is this a bug ? Or it's not supposed to using like this?

SimoTod:
It closes #81 

When a x-if or x-show directive doesn't use transitions, the callback is called twice due to a missing early return (potentially three times if the call receives forceSkip = true). As additional side effects, v-if prints an error in console as pointed out in #81 and it removes the next sibling if there are any at the same level.

Demo of the issue: https://codepen.io/SimoTod/pen/JjoZaEv


calebporzio:
Perfect, thanks!

Issue #77, opened 2020-01-12
Undeclared model object keys default to undefined instead of empty string


jreviews:
Take a look at the following examples:

Alpine.js - https://codepen.io/jreviews/pen/NWPMVMz

Vue.js - https://codepen.io/jreviews/pen/wvBjbNX

In Alpine.js, undeclared model object keys default to undefined, while in Vue.js they default to an empty string. Is it possible to have a similar behavior as Vue.js? It's ok to add the model keys to the object if there's a handful, but if there are dozens or hundreds it gets more complicated.

goodallt:
You could also explicitly define the object attribute as an empty string. It would look something like this: 
```<form x-data="{ formData: { name: '' } }">```

jreviews:
@goodallt Thanks, that's what I mentioned before:

> It's ok to add the model keys to the object if there's a handful, but if there are dozens or hundreds it gets more complicated.

I believe it would be an improvement not to explicitly have to declare every input in the object. I am thinking of very large forms here.

calebporzio:
Hmmm... I guess I don't have a strong opinion on this either way. I would be open to a PR, if the implementation was simple enough.

calebporzio:
Should be fixed in the latest release- thanks for the PR!

simarlab:
It's really closed this issue? I still receive undefined instead of an empty string.
Let me know, and thanks a lot for your great work!

Henriquex25:
I still receive undefined instead of an empty string.

SimoTod:
This bug was raised for v2, v3 has a different implementation and, currently, it returns the real value.
I think you can work around that with something like `<span x-text="myobj.myprop ?? ''">`

PR #87, opened 2020-01-15
Set initial value of undefined nested model keys to empty string

jreviews:
Please check if this is good enough for #77 
Thanks!


calebporzio:
Yep, looks good to me. I made a fairly large refactor. If you wouldn't mind adjusting it to the new location of the x-model code, and adding a test, I'll pull it in. Thanks!

PR #89, opened 2020-01-16
Update the example import statement, fix eslint error

opdavies:
As the `Alpine` variable is not used in the installation example, eslint could throw a `'Alpine' is defined but never used` error which would cause a build to fail.

Simplifying the import statement and not setting a variable name fixes this issue.


calebporzio:
Good call, thanks!

Issue #90, opened 2020-01-16
Error when mutation target is removed


thormeier:
I'm trying to use Alpine together with the select-library Choices (https://github.com/jshjohnson/Choices). It works, but an error is thrown when a new value is selected:
```
component.js:208 Uncaught TypeError: Cannot read property 'isSameNode' of null
    at MutationObserver.observe.childList
```

This likely happens because Choices removes unused options from the underlying select, which changes the value of the select, and adds the newly selected option to the DOM.

Here's a jsbin to reproduce the issue: https://jsbin.com/dukogakequ/edit?html,console,output

I'll start working on a PR today or tomorrow.

thormeier:
Came around doing a PR earlier than I thought :)

PR #91, opened 2020-01-16
Check for existence of parent component first when filtering mutations

thormeier:
Suppresses an error caused by the parent component being null in some
cases.

Fixes https://github.com/alpinejs/alpine/issues/90.


calebporzio:
Hey @thormeier - what are the cases where it's null?

thormeier:
See https://github.com/alpinejs/alpine/issues/90 - I wasn't able to fully understand the interdependence between Choices and Alpine, but I'm almost sure it happens when the mutation target is removed from the DOM before the observer is triggered. That's at least what my observation yielded.

joedixon:
I should have read through the PRs first, but I just ran into this same problem and raised and issue #110 

I get a similar issue when I have the WhatFont Chrome extension enabled.

joedixon:
I should have read through the PRs first, but I just ran into this same problem and raised an issue #110 

I get a similar issue when I have the WhatFont Chrome extension enabled.

joedixon:
I should have read through the PRs first, but I just ran into this same problem and raised an issue #110 

I get a similar issue when I have the WhatFont Chrome extension enabled.

joedixon:
I should have read through the PRs first, but I just ran into this same problem and raised an issue #110 

I get a similar issue when I have the WhatFont Chrome extension enabled.

joedixon:
I should have read through the PRs first, but I just ran into this same problem and raised an issue #110 

I get a similar issue when I have the WhatFont Chrome extension enabled.

joedixon:
I should have read through the PRs first, but I just ran into this same problem and raised an issue #110 

I get a similar issue when I have the WhatFont Chrome extension enabled.

joedixon:
I should have read through the PRs first, but I just ran into this same problem and raised an issue #110 

I get a similar issue when I have the WhatFont Chrome extension enabled.

joedixon:
I should have read through the PRs first, but I just ran into this same problem and raised an issue #110 

I get a similar issue when I have the WhatFont Chrome extension enabled.

ryatkins:
+1 Get this error as well.  Running Chrome in Incognito Mode fixes the issue.

calebporzio:
Thanks @thormeier.

Any chance you can write a test that simulates this behavior? I merged it, but because no test will fail if I roll it back - it could be refactored out in the future and the error could crop back up again.

Thanks.

thormeier:
@calebporzio Sure thing, will do a follow-up PR on this.

thormeier:
@calebporzio Added a test here: https://github.com/alpinejs/alpine/pull/126

PR #93, opened 2020-01-16
Set default value of undefined nested model keys to empty string for x…

jreviews:
Addresses #77 and replaces the previous PR #87 and sets undefined model key output to empty string for x-model and x-text. A test is also included.


calebporzio:
Thanks!

PR #99, opened 2020-01-17
Use Tailwind transition and transform classes

sebastiandedeyne:
Updated the transition snippets to use the new Tailwind classes available in v1.2, so copy-pasters have an immediately working example.


calebporzio:
Siiiick... thanks!

Issue #73, opened 2020-01-11
Add x-else and x-unless conditions


austenc:
Are `x-else` and/or `x-unless`  planned? Would be great to have them! Let me know if you want some help on these, happy to pair up or submit a PR :)

calebporzio:
Ah, `x-else` would definitely be handy. May want to pair on it though. One of those things that could be really easy, but also could be annoying. Good call

austenc:
Yeah, my thought too -- could get tricky! Let me know if/when you want to carve out some time for this, would love to help. Thanks Caleb!

WebDevEtc:
I wanted to play with alpine for a small project, and wrote up a little implementation of x-else.

An initial rough version can be seen in this commit - https://github.com/WebDevEtc/alpine/commit/d374fe42f4bb7f0983f834b5a1165140df726065

I added a bit of logic to the switch - for `x-else` it checks the previous sibling, and determines if it was a `<template>` tag with `x-if`. (With a couple more if statements to help with debugging via a `console.warn()`).

I'm sure there are some edge cases that I haven't thought of. Feel free to completely disregard if either of you already worked on this (I know your replies in this thread were from not too long ago).

I haven't done a PR as I would want to test it a little more first, probably clean up the if statements a little.

calebporzio:
Another weird thing. With Vue, `x-show` and `x-else` work together. So, although an `x-if`/`x-else` implementation is pretty straightforward, I would want to support `x-show`, which would be less straightforward.

This feels like something that will add unnecessary complexity to the codebase for a fairly small yield. Thoughts?

austenc:
Agree with everything you've said. After using alpine to build more stuff I think it should stay simple!

`x-else` in most cases would mostly just be sugar for `x-if="!something"`. I think it's okay to have a few negation operators given the complexity the alternative (else) would bring. As such, I'm going to close this issue :) Thanks!

MarcelRusu:
This would be very nice to have

coderbhaai:
Isnt this suppose to be a basic requirement?

PR #102, opened 2020-01-17
Feature/x else

WebDevEtc:
Following on from https://github.com/alpinejs/alpine/issues/73 - here is an implementation of a x-else feature.

Usage would look like this:

```
<template v-if="some condition"><h1>Some content</h1></template>
<template v-else><h1>Other content</h1></template>
```


sebastiandedeyne:
In my experience it's best to leave version bumps out of PRs altogether. It's up to the maintainer to decide, and also can depend on the contents of other open PRs or upstream changes.

(cc @joshgrib)

PR #103, opened 2020-01-17
Add documentation link to x-if warning

sebastiandedeyne:
Since there's no reasoning in the console warning (would be to verbose anyway), a small documentation link could be useful for curious developers.


calebporzio:
Thanks @sebastiandedeyne!

Issue #97, opened 2020-01-17
x-for "removes all elements when array is empty" fails when items.length > 1


jreviews:
Per the subject, the following test for `removes all elements when array is empty` fails:

```
test('removes all elements when array is empty', async () => {
    document.body.innerHTML = `
        <div x-data="{ items: ['foo','bar'] }">
            <button x-on:click="items = []"></button>

            <template x-for="item in items">
                <span x-text="item"></span>
            </template>
        </div>
    `

    Alpine.start()

    expect(document.querySelectorAll('span').length).toEqual(2)

    document.querySelector('button').click()

    await wait(() => { expect(document.querySelectorAll('span').length).toEqual(0) })
})
```

 Expected: 0
 Received: 2

    > 45 |     expect(document.querySelectorAll('span').length).toEqual(0)

PR #104, opened 2020-01-17
Fix emptying for loop

donkooijman:
This closes: #97 

I added another test besides the existing one. The new test checks against a previous array with multiple items.


calebporzio:
Great! Thanks a bunch!

calebporzio:
Thanks!

PR #123, opened 2020-01-21
Restore support for Microsoft Edge (engine EdgeHTML 18)

### LINKED ISSUE #109: Alpine.js not working in Edge
Hello, 

I've been experimenting with Alpine.js today and it looks really nice, but when I tried to test it in Edge and Internet Explorer it didn't look like it was working. I tried the simplest example from the docs, but it didn't work. 

I have added polyfills as described in docs. 

Here is the message that I've got in the console. 

```
SCRIPT1028: SCRIPT1028: Expected identifier, string or number
alpine.js (1,5743)
```

Tested code example:

```
<!DOCTYPE html>
<html>
    <head>
        <meta charset="UTF-8">
        <meta name="viewport" content="width=device-width,initial-scale=1.0">
        <script src="https://polyfill.io/v3/polyfill.min.js?features=MutationObserver%2CArray.from%2CArray.prototype.forEach%2CMap%2CSet%2CArray.prototype.includes%2CString.prototype.includes%2CPromise%2CNodeList.prototype.forEach%2CObject.values%2CReflect%2CReflect.set"></script>
        <script src="https://cdn.jsdelivr.net/npm/proxy-polyfill@0.3.0/proxy.min.js"></script>
        <script src="https://cdn.jsdelivr.net/gh/alpinejs/alpine@v1.9.2/dist/alpine.js" defer></script>
        <link href="https://fonts.googleapis.com/css?family=Fira+Sans+Condensed:400,500,600,700&display=swap" rel="stylesheet">
    </head> 
    <body>
            <button @click="open = true">Open Dropdown</button>
        
            <ul
                x-show="open"
                @click.away="open = false"
            >
                Dropdown Body
            </ul>
        </div>
    </body>
</html>
```

SimoTod:
This PR could be a little controversial so feel free to reject it.

As a few users pointed out, alpine doesn't start up with ie11 and edge.. 

I'm not too worried about ie11 but we should probably support edge (I'm talking about the previous version which most of the users have installed, i think Microsoft released a new version based on chromium last week).

The reason why it's not working is that we use `node: current` as a compiling target which includes features and optimisations that Edge doesn't support (for example, the spread operator)

It would increase the size of the bundle from 12.7KB (~4.4KB gripped) to 14.3KB (~4.7KB gzipped). 

Closes #109


calebporzio:
I'm down - thanks for being on top of this @SimoTod!

Issue #110, opened 2020-01-19
Conflict with Chrome extension


joedixon:
I use a Chrome extension called [WhatFont](https://chrome.google.com/webstore/detail/whatfont/jabopobgcpjmedljpbcaablpmlmfcogm?hl=en)

This seems to be adding a spurious div to my page with no parent element. This causes an error in the console. It doesn't break anything, but it would be nice to get rid of the error.

<img width="470" alt="Screenshot 2020-01-19 at 20 52 33" src="https://user-images.githubusercontent.com/3438564/72688318-bec5ea80-3afd-11ea-8355-143cd7124610.png">

I have done some digging in the code and I think it can be solved by adding an extra check when setting up the `MutationObserver` as below:

```
if (node.nodeType !== 1) return

// ADD THIS EXTRA CHECK TO ENSURE THE NODE HAS A PARENT
// Check the node has a parent
if (!node.parentElement) return;

// Discard any changes happening within an existing component.
// They will take care of themselves.
if (node.parentElement.closest('[x-data]')) return
```

But I've never used this class before and only just started dabbling with the framework so wanted to run it by you first in case there is anything that this could break.

Lmk - happy to submit a PR if you think the solution works.

Thanks,
Joe

calebporzio:
Should be fixed in the next release. Re-open if not.

PR #126, opened 2020-01-22
Add test for PR #91 regarding disconnected nodes and fix same issue while listening for new components.

thormeier:
@calebporzio I investigated the specific behaviour a little further and came to the conclusion that it actually is some weird behaviour by done by Choices. Somehow they manage to trigger a mutation on a DOM Node that is not even connected anymore (`target.isConnected === false` in the moment where the MutationObserver kicks in).

For simplicity I'm not replicating the exact behaviour of Choices here, but just mocking it by passing not yet added nodes as mutation targets, which is basically what happens.


keyurshah:
hello, i'm receiving an error and it seems possibly related to flickity changing the DOM at the same time as alpine is looping through.  This might be related to issues #90 and #110 

I get the error

```
index.js:85 Uncaught TypeError: Cannot read property 'closest' of null
    at Object.<anonymous> (index.js:85)
    at NodeList.forEach (<anonymous>)
    at Object.<anonymous> (index.js:85)
```

Possibly a check needs to be made in line 63 of index.js to see if the node exists when running closest This might be similar to what you did in pull request #90 to check for the node?

```
                        if (node.parentElement.closest('[x-data]')) return
```

thank you

thormeier:
@keyurshah Sounds like it, at least. Could you maybe create a codepen/jsbin to reproduce the error? Then I could try to see if it actually is related to the MutationObserver issue.

keyurshah:
@thormeier  I sent a link via the email in your profile to my local dev instance.  I couldn’t get it working on jsbin due to the complexity. 

Thanks!

keyurshah:
I was able to get a demo setup with the error here

* errors when importing flickity and alpinejs
  * https://stackblitz.com/edit/alpine-flickity

```
Cannot read property 'closest' of null
```




* works correctly when importing only flickity
  * https://stackblitz.com/edit/flickity-only

thormeier:
@keyurshah Thank you for providing the setup, helped a lot to understand what was going on. No need to apologize, the fix actually fits in here pretty well! 

I managed to alter the test in this PR to reproduce the exact issue and also added a fix for that. Also renamed the PR.

@calebporzio I hope you're ok with me fixing another related issue in this PR.

keyurshah:
I tried out the new build and it seems to have fixed the `closest` issue with flickity. Thanks so much @thormeier !!

thormeier:
Ok, in my opinion, this is getting worrisome.

We now have several libs, a CMS and chrome extensions causing this. (Issues https://github.com/alpinejs/alpine/issues/136, https://github.com/alpinejs/alpine/issues/135, https://github.com/alpinejs/alpine/issues/110, https://github.com/alpinejs/alpine/issues/90, PRs https://github.com/alpinejs/alpine/pull/126 and https://github.com/alpinejs/alpine/pull/134)
But no actual minimal version of of code to reproduce this. I'm starting to think of this as either a Chromium bug (have to investigate there, if that's a known issue) or some really weird thing alpine is doing that poses an edge case of the MutationObserver.

The fix (i.e. checking if target node is even there) plus its test are more or less fighting symptoms and we don't really know what causes this and what side effects these fixes might have. Someone got an idea how to proceed with this?

SimoTod:
I believe the mutatorobserver is set on the body so any dom changes will trigger it regardless them being started from an alpine component or not. That includes other javascript libraries, chrome extensions, etc. Some of the element may not have a parent because they are moved  around so any element lookup which could return a null value (parent, closest, etc) should be cheched before chaining the next method. I think your second PR will fix all of the new bugs.

calebporzio:
I tried to merge this but ran into some tree-related problems. I'd love it if you re-PRed this without the force push.

Also, I don't have the capacity right now to fully wrap my head around this full discussion.

I'd also love it if someone could let me know if this PR fixes the issues everyone is referencing, or if they still exist. It's a little hard for me to keep track.

Thanks for the contribution!

fakingfantastic:
👋 Hey @calebporzio , just throwing in my $.02. I'm currently working on a WordPress site that uses a Page builder, and I'm getting the error.

I'm including alpine.js globally, so it's throwing the `closest of null` error on every page of the site, and no directives are working. However, I recreated the accordion tab on one page of the site and it's working there! Turns out when you add the `x-data={..}` it allows the inner directives to work. I'm guessing it's making an isolate scope or something?

I should also mention, @thormeier 's PR does fix the error, but not the issue I mentioned above (unless that's intentional ¯\_(ツ)_/¯ ) . So, I vote merge it!

Anyway, hope this helps. @calebporzio if you want to ping me directly, I can show you a link to the site live so you can see the issue I'm mentioning

calebporzio:
Thanks for the heads up @fakingfantastic.

The other issue you're referring to, are you saying that something like `x-show` doesn't work outside of a parent `x-data`?

fakingfantastic:
@calebporzio - yes, pretty much, I can't get any directive to work, unless it's inside a node that has `x-data` on it.

Issue #79, opened 2020-01-13
[Feature Request] x-watch


TomS-:
In this example; there is a slider (let's say flickity), we want to create a custom UI that will track the active slide. AlpineJS would be good to do this because we can minipulate the UI based on the active slide.

Let's say for example on each slide we have a UI that has the title of the slide, we can use `<div @click="activeSlide = 1; activeTitle = 'Projects'">Projects</div>`

Then we can use x-watch to watch the activeSlide and run a function when it changes, I'm unsure of the best implementation for this `x-watch:activeSlide="function"` or `x-watch="{ activeSlide: function }`  so for example we could do `x-watch="{ activeSlide: $el.select(value) }"` value would be the value of the variable. I know the syntax is completely wrong, but it's the best way I can think of demostrating my thinking.

TomS-:
Noticed @pfrlv responded with a confused emoji. Here is Vue's implementation - https://vuejs.org/v2/guide/computed.html#ad

calebporzio:
Yeah, I could see the use for this. I'd probably rather do it like Vue does:

```
this.$watch('activeSlide', funcion(value) {...})
```

Thinking this doesn't need to be a directive though right? (`x-watch`)

TomS-:
@calebporzio correct me if I'm wrong, but wouldn't this rely on taking the function approach (https://github.com/alpinejs/alpine#x-data)? I don't see any harm in that, it could be seen as advanced functionality. I just thought `x-watch` or `x-computed` would keep the inline/TailwindCSS nature.

SimoTod:
@TomS- 
I like the idea of having a `x-watch` but only allowed on the same tag where you define x-data (like `x-init` does).
I'm not sure if we need 'x-computed, would it be the same as calling a function directly when you use the variable? (https://codepen.io/SimoTod/pen/RwNBeJr)

TomS-:
@SimoTod I also don't believe we need x-computed for this and I do believe it would be called where you define `x-data` I'm just unsure of the syntax but I was thinking for example: `x-data="activeText: 'Hello'" x-watch="{ activeText: function(value) { this.$refs.text.innerText = value } }"`  so you can later define `@click="activeText = 'Goodbye'"` and it will auto-update the text refs innerText.

SimoTod:
@TomS- 
Yeah, I got that. In my opinion, to keep it consistent with the core, it should be something like `x-watch="{ activeText: '$refs.text.innerText = activeText' }" `
[...]

PR #127, opened 2020-01-22
x-watch directive

SimoTod:
First implementation of a x-watch directive similar to what Vue JS has.

Right now, it doesn't support deep watchers (if a watcher is observing a variable, changes on nested levels won't trigger the callback).

The syntax is slightly different from Vue and more consistent with the rest of Alpine.
I.e. `x-watch="{variable: '[expression]' }"`

Readme probably needs updating. 

Feel free to reject it if it's not a good fit for the core or ask for changes.
There's a discussion ongoing here (#79) but we weren't sure if this is something you would like to include or if you prefer a different syntax.


calebporzio:
It's a good idea and will probably end up in the core at some point, but going to hold off for now.

PR #129, opened 2020-01-23
Add "selected" attribute into boolean attrs list

rzenkov:
Fixes <option :selected="selected"></option> behavior


calebporzio:
Thanks!

PR #134, opened 2020-01-24
Adds extra check for parentElement on mutation

KevinBatdorf:
In the case where the parentElement is null this would otherwise throw an error.

(The mutation happening in my example is from another script)

Screen Shot 2020-01-24 at 7 41 56 PM

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -60,7 +60,7 @@ const Alpine = {
 
                         // Discard any changes happening within an existing component.
                         // They will take care of themselves.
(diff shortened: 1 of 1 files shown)


thormeier:
Hey @KevinBatdorf, thank you for your contribution - what exactly is the difference to https://github.com/alpinejs/alpine/pull/126 ?

calebporzio:
Thanks @KevinBatdorf for PRing this, however it's already been PR'd and merged.

PR #138, opened 2020-01-25
Fix count of Magic Properties

davidsneighbour:
It's 4 properties now, not 3, sorry, only one single character change ;)


calebporzio:
Hah, love it! Thanks.

PR #139, opened 2020-01-25
TypeError in strict mode

### LINKED ISSUE #132: TypeError: proxy set handler returned false for property
I'm unsure if this will be fixed, or ever could be fixed without a huge refactor. But I thought I'd just check to see if it's anything I'm doing wrong. I want to store a slider in the proxy so I can access it for further functions. In this example I want to select the next slide when I fire a click `@click="nextSlide()"`. To do this I've stored the slider as data. However, it's returning the error `TypeError: proxy set handler returned false for property`. The code is:
```
let flkty = new Flickity(el, {
                    wrapAround: true,
                    prevNextButtons: false,
                    pageDots: false,
                    setGallerySize: el == this.$refs.slider ? false : true,
                    selectedAttraction: 0.01,
                    friction: 0.15
});

this.slider = flkty;
```

Thanks

SimoTod:
Closes #132

When javascript runs in strict mode and there is an assignment when the reactivity is suspended (for example, using x-init), the proxy returns an error as per documentation:
```
The set method should return a boolean value. 
Return true to indicate that assignment succeeded. 
If the set method returns false, and the assignment happened 
in strict-mode code, a TypeError will be thrown.
```
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy/handler/set#Return_value

Here's a codepen to reproduce the issue: https://codepen.io/SimoTod/pen/ZEYPrdQ


calebporzio:
Great, thank you!

PR #141, opened 2020-01-26
Fix nested proxies issue

### LINKED ISSUE #140: Problems using x-for and sorting
I'm probably missing something here: https://codepen.io/jespr/pen/qBEvMQW

But if you click on Name, then age, then Name, then Age rapidly after each other the whole page ends up hanging and there's a message in the console about setTimeout internally in Alpine.js taking XXX amount of ms:

```
[Violation] 'setTimeout' handler took 149ms
```

What am I doing wrong? :)

SimoTod:
closes #140

Please see the thread for a description of the issue.


calebporzio:
Perfect @SimoTod, thanks so much!!

Issue #145, opened 2020-01-28
x-on debounce?


awcodes:
For starters, thank you so much for this Caleb. I freaking love alpine.

Just a quick question though. Is there no debounce on input methods like there is in Livewire.
I've tried:
x-on:input.debounce="function()"
x-on:input.debounce.500="function()"

And both always execute the function immediately.

I know I can just use setTimeout in the function. Just wanted to see if it was already built in and i'm just doing it wrong.

Thanks again for the awesome framework!!!

calebporzio:
Yeah, I would be down for adding this to the framework. If someone could PR this that would be great.

awcodes:
That’s awesome Caleb. Thank you. 

You’re making the dev world such a better place to work and live in.

jreviews:
I added a PR and you can see it running here https://codepen.io/jreviews/pen/oNgPgXG

the94air:
Dang it! I wanted to make this PR :laughing:

awcodes:
You guys rock. Wish I was better at JS so I could have PR'd it. :)

Works as I expected.

Thanks everybody.

ryangjchandler:
Going to close this issue since the PR got merged. Thanks for contributing guys! Just a little note, if you mention the issue number inside of the PR description using 'Fixes #[issue]', or 'Closes #[issue]' it will automatically close issue once merged. Thanks again!

PR #153, opened 2020-01-31
Added x-on debounce modifier

jreviews:
This relates to #145 and includes test. Default wait is 250.

Example:

```
<div x-data="{}">
    <input x-on:input.debounce.100="$refs.span.innerText = parseInt($refs.span.innerText,0) + 1" />
    <span x-ref="span" x-text="0"></span>
</div>
```


src/directives/on.js
```diff
@@ -24,9 +24,12 @@ export function registerListener(component, el, event, modifiers, expression, ex
         const listenerTarget = modifiers.includes('window')
             ? window : (modifiers.includes('document') ? document : el)
 
+        const useDebounce = modifiers.includes('debounce')
+        const wait = modifiers.length > 1 ? parseInt(modifiers[1]) : 250
+
         const handler = e => {
             if (isKeyEvent(event)) {
-                if (isListeningForASpecificKeyThatHasntBeenPressed(e, modifiers)) {
+                if (!useDebounce && isListeningForASpecificKeyThatHasntBeenPressed(e, modifiers)) {
```
SimoTod: I'm not sure this line needs changing. We always want to skip a key event if the pressed key doesn't match what we defined whether or not we added the 'debounce' modifier.

For example:
```html
        <div x-data="{foo: 0}">
            <input x-on:keydown.a.debounce="foo = foo + 1" />
            <span x-text="foo"></span>
        </div>
```

When we press any keys, the counter will increment but it should not since we only want to increment it when we type `a`.

src/directives/on.js
```diff
@@ -24,9 +24,12 @@ export function registerListener(component, el, event, modifiers, expression, ex
         const listenerTarget = modifiers.includes('window')
             ? window : (modifiers.includes('document') ? document : el)
 
+        const useDebounce = modifiers.includes('debounce')
+        const wait = modifiers.length > 1 ? parseInt(modifiers[1]) : 250
```
SimoTod: Just a suggestion, feel free to ignore it if you think it's pointless.

This line assumes that `wait` is always the second modifier which would remove flexibility from the feature (and I feel people will forget about that most of the time).
For example, `click.away.debounce.1000` or `keypress.esc.debounce.1000` would not work as expected.

Would it not be safer to use indexOf to get the position of `debounce` and check/parse the next modifier?

By the way, well done @jreviews.

src/directives/on.js
jreviews: @SimoTod Thanks for your feedback! You are right. For some reason while working on this, that was the only way I was able to make it work, but checked again and it works without that, so I removed it.

src/directives/on.js
jreviews: Good catch and it makes perfect sense. I went ahead and updated the code. Thanks!

src/directives/on.js
SimoTod: You probably added it because, without that check, when you add 'debounce' the handler is always going to return without executing the code since `isListeningForASpecificKeyThatHasntBeenPressed` returns true. 

I think you need to handle that case inside `isListeningForASpecificKeyThatHasntBeenPressed`.

👍

jreviews: I am not sure now because I removed the check and both the automated test and the codepen with the updated code are working. Can you check here? https://codepen.io/jreviews/pen/XWJLoLZ

SimoTod: Sorry, I typed the previous message quickly and I didn't explain myself really well.
The implementation is good but there is still an edge case which is not fully covered.

For example, check out this pen: https://codepen.io/SimoTod/pen/JjoQxXw
The first input doesn't have the debounce modifier, the second one does.
Both or them are supposed to increase the counter when the key 'a' is pressed (but they should not increase the counter when another key is pressed).

At the moment, the debounced one does not increment the counter. With the previous implementation, the counter was updated with any key.

I hope it makes sense.

jreviews: Ah yes! I remember now. The issue was with the regular inputs, no the debounced ones. Your demo is so much better. I'll have another go at it. Thanks.

jreviews: Can you check again? the issue with the key modifier should be fixed and I updated the tests.

jreviews:
@SimoTod @calebporzio I noticed that livewire uses the 'ms' string for the wait time in the debounce modifier. Should we use the same syntax here? `keyup.a.debounce.500ms` instead of `keyup.a.debounce.500` ? That may also help to better identify if the modifier after debounce is an actual wait time.

SimoTod:
I think it would be a good idea but let's see what Caleb thinks.

ryangjchandler:
Definitely agree on the 'ms' suffix, fluent syntax between the libraries definitely makes it easier when combining the 2

calebporzio:
Love this, great PR - yeah, if we could keep it in sync with Livewire's api (offering ms and s suffix), that'd be great.

jreviews:
@calebporzio @SimoTod Updated the PR with support for ms and s suffixes.

SimoTod:
Is there a small risk that https://github.com/alpinejs/alpine/pull/153/files#diff-9d25c174874ef105b3c21a7aad59a637R70 will remove a different modifier in the future if debounce is in the middle of the chain and it doesn't have a time interval?

I wonder if it would be safer to only support Xms and Xs and match them using a regExp as Livewire does.

Apart form that (it's only a personal observation), it looks great to me. 👍

jreviews:
That's what I thought before and for sure that would be useful if there us another modifier that could be used after debounce. Is there one right now? 

I have code ready to add this:
```
    // Need to remove debounce and wait modifiers if present
    if ( modifiers.includes('debounce') ) {
        const debounceModifierIndex = modifiers.indexOf('debounce')
        let debounceIndexes = [debounceModifierIndex]
        if ( /^[0-9]+(ms|s)$/.test(modifiers[debounceModifierIndex+1]) ) {
            debounceIndexes.push(debounceModifierIndex+1)
        }
        keyModifiers = modifiers.filter( (modifier, index) => {
            return !debounceIndexes.includes(index)
        })
    }
```

Let me know if you guys prefer this and I'll push the PR.

SimoTod:
I don't think there is one from a logical point of view but other modifiers
are flexible so they work regardless their position. It's not my call
though, let's wait for Caleb to review. Thanks again.

calebporzio:
This is looking great BTW, but yeah, I'd like to keep the API flexible to have something like `@click.debounce.window` work if that makes sense or even more likely `@keydown.debounce.enter`.

I'm not as worried about being able to put the `Xms/Xs` not directly after `debounce` I think most people would alwasy do `debounce.Xms` if they were specifying a time.

Make sense?

Thanks!

jreviews:
Done, with tests. Wait times must specify an 'ms' or 's' suffix to be valid. All of these should be working:

- `@click.window.debounce`
- `@click.debounce.window`
- `@click.window.debounce.1s`
- `@click.debounce.1s.window` 

- `@keydown.enter.debounce`
- `@keydown.debounce.enter`
- `@keydown.enter.debounce.1s`
- `@keydown.debounce.1s.enter`

jnstq:
This would be awesome! When can we except this to be merged?

calebporzio:
Thanks for your contribution. I made some pretty major changes and refactors, check em out if you care to know.

Thanks again! This will be in the next release.

PR #160, opened 2020-02-03
Skip transitions when visibility does not change

### LINKED ISSUE #159: x-transition fires even if x-show/x-if state didn't change.
x-transition will always fire when any of the proxy data changes. This can cause for some unwanted animation being fired when updating other parts of the proxy data (UI).

I've put together a quick pen to show you, if you click "toggle" you will see "Hello" shake as it's trying to apply the classes even though it's data was not effected: https://codepen.io/TomS-/pen/JjogrGa

This is happening to me in Firefox, I have not tested Chrome.

SimoTod:
Closes #159 

Every time something changes,  Alpine reevaluates each x-show directive and applies the transition classes again, causing a quick flickering in some browsers (I.e. Firefox). 
This PR update Alpine so it will skip the transition calls when the visibility of an element hasn't changed.


calebporzio:
Thanks @SimoTod!!

PR #162, opened 2020-02-04
Prevent default action when an event handler returns false

### LINKED ISSUE #150: [Bug?] Javascript confirm dialog cancel not working
```html
<form action="/documents/1" 
    method="POST"
    x-data
    onclick="return confirm('Move to trash?')"
    x-on:click="return confirm('Move to trash?')"
>
    <button type="submit">Delete</button>
</form>
```

The OG `onclick=""` is working as expected. OK continues the request, and cancel aborts the request.

The Alpine version however always continues the request, on both OK and cancel.

Might be me doing something wrong, but I can't seem to get it working. Any ideas?

SimoTod:
Closes #150 
It emulates the native behaviour so `@click="return confirm('...')"` will stop the default action but `@click="confirm('...')" won't.

https://codepen.io/SimoTod/pen/gOpYoBQ


calebporzio:
Great! Thanks!

Issue #83, opened 2020-01-15
Issues with setTimeout


rusya1313:
I tried to create a simple TODO app and got my browser freezing.
[Violation] 'setTimeout' handler took <N>ms
Please have a look here: https://codesandbox.io/s/priceless-hofstadter-ijk3t

calebporzio:
Hmm... I don't see that problem at all when I load it in the sandbox. Do you get the same results in a CodePen?

Rusya13:
@calebporzio  Have you tried to add more than 8 tasks? It doesn't matter where. On my local environments the same result..

calebporzio:
I see the performance issue now. You are doing some pretty crazy things in that component. The problem is somehow the Proxy is getting nested an exponential number of times and creating this issue.

I can't pinpoint the exact problem, but here's a better way to write it that can handle large amounts of tasks:
```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="UTF-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
    <meta http-equiv="X-UA-Compatible" content="ie=edge" />
    <script
      src="dist/alpine.js"
      defer
    ></script>

    <title>Document</title>
  </head>
  <body>
    <div x-data="pageData()">
        <section id="add-task-container">
          <form x-on:submit.prevent>
            <input type="text" x-model="newLabel" />
            <button type="submit" x-on:click="addTask()">
              Add task
            </button>
          </form>
        </section>
        <div x-html="renderTasks()"></div>
    </div>
    <script>
      var store = {};

      function registerStore(data, name) {
        store[name] = data;
      }

      function pageData() {
        return {
            tasks: [],
            newLabel: '',
            addTask() {
                this.tasks.push({
                    id: Math.random().toString(36).substr(2, 9),
                    label: this.newLabel,
                    done: false,
                })

                this.newLabel = ''
            },
            renderTasks() {
                return this.tasks.length === 0 ? '' : this.tasks.map(task => {
                        return `
                            <div>
                                ${task.label}
                            </div>
                        `;
                    })
                    .join("");

            }
        };
      }
    </script>
  </body>
</html>
```

If you can pin-point the culprit in your code and isolate it to a simple code sample, we can open up this box again, but for now, I'm closing this.

Rusya13:
I saw another example on Twitter:
https://twitter.com/alexfwulf/status/1221191875379507200
[...]

PR #163, opened 2020-02-05
Proxy refactoring

SimoTod:
As highlighted by #83, there are still performance issues with the proxy.
When a proxy is created, even if it's not directly assigned to a variable, it stays in memory and it keeps observing the variable. When accessing multiple time the same variable, the proxyHandler creates a new proxy object every time and they remain active after the initial use.
It means then, if we read 10 time a property containing an object, there will be 10 proxy watching the variable which will trigger 10 refreshed every time it changes. 

Proposed new approach: rather than creating a new proxy every time we access a nested variable, we create a deep proxy at the beginning so we don't proxy only the first level but we do recursively for all level (objects would generate a proxies as soon as a directive use them any way so resource wise it's pretty much the same). The get trap just return true if property is '$isAlpineProxy' otherwise it returns the property value: we don't create new proxy here so it should resolve both the issues where proxy were nested or duplicated.
The set trap, when we add a new object, will try to create a deep proxy before inserting it to keep the property reactive.

The deepProxy utility will just create a proxy for valid objects and will ignore null values, DomNodes, existing proxies, etc.

I tried myself and I saw performance improvements, old tests still pass and examples looks okay.
I've asked one of the user to stress test it as well.

I'm conscious that it touches a low-level part of the core so I understand if you want to take a bit of time to test it or you want to changes/improves anything.

Hopefully it should fix all this proxy drama. :)


calebporzio:
I LOVE this. Thank you so much!

PR #165, opened 2020-02-07
Improve x-for reactivity

SimoTod:
Alternative solution for #144 using defineProperty + setter and getter (Ref. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty#Custom_Setters_and_Getters) instead of another proxy item.

The same considerations I made for #151 about the spread operator applies to this PR: the spread operator doesn't copy properties into the new object, only normal variables.

Performance (time and memory) will improve slightly as well since we are not going to create the same copy of an object 2-3 times.


SimoTod:
Updated to support the last release. Feel free to close it if not useful.

I think all the `if (typeof extraVarsData === 'undefined') extraVarsData = {}` lines can be deleted using proper default, let me know if you want me to look into it and update the PR.

calebporzio:
Going to hold off on this for now. However, I will add this PR to the v3 roadmap so we can use the contributed test to test against the new reactivity engine. Thanks!

PR #174, opened 2020-02-12
Unbinding global events when element is removed from the page

### LINKED ISSUE #167: Unbinding global event listeners w/ Turbolinks
The `window` and `document` modifier for `x-on` doesn't work well with turbolinks. The event listener gets duplicated each time the element is initialized (see [this minimal project](https://github.com/luizpvas/alpine-window-event-with-turbolinks) that reproduces the issue).

I think this could be solved by watching removed nodes in mutation observer, but Alpine would need to keep track of the global handlers associated with each element. I'm not sure this is a good path.

I'm glad to work on a PR if there is a good solution for this issue :+1:

luizpvas:
Based on the suggestion from @SimoTod, this implementation unbinds global events on the event handler itself when the element is removed from the body.

This fixes #167.


calebporzio:
Thanks @luizpvas!

Issue #173, opened 2020-02-12
Register multiple listeners with x-on


gwleuverink:
I recently came across a situation where I had to do something to the effect of:

```html
<button 
    x-on:click="doSomething()" 
    x-on:touchend="doSomething()"
> 
    Do something!
</button>
```

It would be great if we could register multiple listeners with one directive. For example:

```html
<button  x-on:[click, touchend]="doSomething()" > 
    Do something!
</button>
```

I'm not sure syntax like that is preferred or even possible, but I'm sure there are other use-cases where something like this could come in handy. Would you consider adding something like this in the future?

gwleuverink:
Just wrote up a small proof of concept :) Would love to hear your thoughts

SimoTod:
It's a good concept but I'm a bit unsure about modifiers: It seems like it applies the same ones to each event in the list, am i right?

How do we cover
`@click` + `@keypress.escape.window` (something that somebody could use for closing a modal)?

TomS-:
@SimoTod What about a pipe seperator like what is often used in arrays? 

`@click|keypress.escape.window="doSomething()"`

gwleuverink:
That is a good question. Yes, in the proof of concept right now it applies all modifiers to all handlers at once.
 
```html

<!--
Here the `prevent` modifier would be applied to both handlers
-->

<button 
    x-on:[click,touchend].prevent="doSomething()" > 
    Do something
</button>
```

The `keypress` and `escape` modifiers are a little bit of an odd duck since they have special behaviour.

1. The keypress modifier has it's own modifiers only applicable on keypress events
2. The window modifier registers a listener on a different node than the element itself

If you were to combine them you'd end up with something like this: 
Please note these are nonsensical examples, but you get the gist :)
```html

<!--
This would register both the click and keydown events on the window object (not what you'd want)
Furthermore the escape modifier would be applied to the click event. (won't do anything)
-->

<button
    x-on:[click,keypress].escape.window="doSomething()" > 
    Do something
</button>
```

I'm pondering the following approaches:
``` html

<!--
Group listeners together based on intent
-->

<button href="#some-id"  
    x-on:[click,touchend].prevent="doSomething()" 
    x-on:keypress.escape.window="doSomething()"
> 
    Do something
</button>


<!--
Register modifiers on the event level. 
I don't really like this approach since it drastically affects readability. But it is an idea we can iterate on
-->

<button  
[...]

PR #175, opened 2020-02-13
Feature/x on registers multiple listeners

gwleuverink:
Addresses feature request #173 

This PR adds support for registering multiple listeners with one x-on directive using the following syntax:

```html
<button  x-on:[click,touchend]="doSomething()" > 
    Do something!
</button>
```

I was not sure if this is a feature you'd like to have, either way it was an interesting source dive :)

## Backwards compatibility
This change won't have any impact I am aware of. Normal `x-on` attributes stil work the way they used to.

## Caveats
In this proof of concept it is currently not possible to separate event names with spaces. Might be able to add that, but this might have some impact on allowed attribute names that alpine parses. Not sure if this could lead to any side effects.

## Thoughts
I was wondering how Vue would do it, but it turns out they don't support this (as far as i'm aware)
Syntax therefore is wide open. Maybe losing the brackets makes it an easier read and keeps the filtering of allowed attribute characters more strict?

If you are open to the idea i'm happy to iterate on this and add some tests

### DIFF
--- a/src/component.js
+++ b/src/component.js
@@ -165,7 +165,7 @@ export default class Component {
 
     getEventNamesFromXAttrValue(value) {
         const firstCharacter = value[0]
-        const lastCharacter = value[value.length]
+        const lastCharacter = value[value.length - 1]
 
         if(firstCharacter !== '[' && lastCharacter !== ']') {
             return [value]
--- a/src/component.js
+++ b/src/component.js
@@ -163,11 +163,27 @@ export default class Component {
         this.resolveBoundAttributes(el, false, extraVars)
     }
 
+    getEventNamesFromXAttrValue(value) {
+        const firstCharacter = value[0]
+        const lastCharacter = value[value.length]
+
+        if(firstCharacter !== '[' && lastCharacter !== ']') {
+            return [value]
+        }
+  
+        eventNames = value.split(/[,\[\]]+/)
+        eventNames.shift()
+        eventNames.pop()
+
+        return eventNames
+    }
+
     registerListeners(el, extraVars) {
         getXAttrs(el).forEach(({ type, value, modifiers, expression }) => {
             switch (type) {
                 case 'on':
-                    registerListener(this, el, value, modifiers, expression, extraVars)
+                    let events = this.getEventNamesFromXAttrValue(value)
+                    events.forEach(event => registerListener(this, el, event, modifiers, expression, extraVars))
                     break;
 
                 case 'model':
--- a/src/component.js
+++ b/src/component.js
@@ -171,7 +171,7 @@ export default class Component {
             return [value]
         }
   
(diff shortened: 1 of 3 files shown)


src/component.js
```diff
@@ -163,11 +163,27 @@ export default class Component {
         this.resolveBoundAttributes(el, false, extraVars)
     }
 
+    getEventNamesFromXAttrValue(value) {
+        const firstCharacter = value[0]
+        const lastCharacter = value[value.length]
```
philippbosch: shouldn't this be 
```js
const lastCharacter = value[value.length - 1]
```
?

gwleuverink: Good catch. It definitely seems like I was sleeping there. I’ll give it another look

gwleuverink: Fixed with commit [464460d](https://github.com/alpinejs/alpine/commit/464460d).

philippbosch: 👍
Thanks for the great PR btw. Looking forward to seeing it land in master.

gwleuverink: Thanks! Happy to contribute

calebporzio:
For now, I'd like to keep the API as minimal as possible.

`<button  @click="doSomething()" @touchend="doSomething()"> 
    Do something!
</button>`

Should do it right?

Closing unless I missed something. Thanks for being willing to PR though!

gwleuverink:
Yep that would work for sure. It is just a shorthand aiming to make those cases a little bit cleaner to read.
Anyway, too bad! Keep up the good work!

deanzod:
> For now, I'd like to keep the API as minimal as possible.
> 
> `<button @click="doSomething()" @touchend="doSomething()"> Do something! </button>`
> 
> Should do it right?
> 
> Closing unless I missed something. Thanks for being willing to PR though!

Just testing this in Chrome with mobile/touch turned on in dev tools. The function gets called twice, once for the touch and once for the click. When I switch off mobile view and click it fires the function once (as you would expect - click only).

simonferndriger:
I would love to see this in AlpineJS, because it is a common case that you want to react the same way on several events. For example you want to save a value from an input field on `@click.outside`, but also on `@keydown.enter`.  Since you have the logic in the html code, you would need to duplicate it which makes it redundent and harder to maintain, and also uglier to be honest here.

SimoTod:
Not the exact feature, but you can also use x-bind and bind multiple directives at once

simonferndriger:
> Not the exact feature, but you can also use x-bind and bind multiple directives at once

Thank you. What is the syntax for that?

SimoTod:
> > Not the exact feature, but you can also use x-bind and bind multiple directives at once
> 
> Thank you. What is the syntax for that?

https://alpinejs.dev/directives/bind#bind-directives

Issue #155, opened 2020-01-31
[Feature] init from data object


shadoWalker89:
Hi,

Currently we can add an `x-init` to do something on init.
@calebporzio are ok with using a function named `init` in the data object as the hook for the `init` ?

```js
function myComponent() {
    return {
        init() {
            // Some init stuff
        }
    ];
}
```

I could make a PR

danfascia:
I think it's good to keep it versatile and allow a user defined function call in `x-init` rather than get too opinionated even though I nearly always call the function `init()` in the data object

luizpvas:
I've used this pattern in a couple different components. For example:

```html
<input class="field__input"
    placeholder="E-mail"
    name="new_users_emails[]"
    x-data="components.autocomplete()"
    x-init="init()"
/>
```

An automatic `init` function (maybe `connect`, similar to stimulus?) would make the HTML for reusable components a bit nicer and less error-prone.

shadoWalker89:
> I think it's good to keep it versatile and allow a user defined function call in `x-init` rather than get too opinionated even though I nearly always call the function `init()` in the data object

I never said that `x-init` should be removed, i'm talking about adding the `init` function hook. 😄 



> I've used this pattern in a couple different components. For example:
> 
> ```
> <input class="field__input"
>     placeholder="E-mail"
>     name="new_users_emails[]"
>     x-data="components.autocomplete()"
>     x-init="init()"
> />
> ```
> 
> An automatic `init` function (maybe `connect`, similar to stimulus?) would make the HTML for reusable components a bit nicer and less error-prone.

That is what i'm doing now, but i want to use the `init()` function without defining `x-init`. To me that feels a lot cleaner. I personally think that an `init` behavior should be coupled to the component and i shouldn't have to use an additional step `x-init` to tell the framework how to do it. But i'm cool with `x-init` as an option for the people that like it.

gsempe:
From my user point of view, I do prefer the explicit usage of x-init over an implicit behavior.

I think that Alpine changes should follow a direction that keep as low as possible the learning curve. Implicit behaviors make the onboarding of new developers more difficult and so I would say that this idea is not good for Alpine itself.

One more time, that's only my personal vision of Alpine.

gsempe:
I think as well to provide alternatives like proposed here, has the same effect to make more difficult the usage of the library for new users

shadoWalker89:
[...]

PR #178, opened 2020-02-13
Automatically call `init` method

nedmas:
Solves #155 

Updates the component class to automatically call an `init` method if one exists on the `x-data` object. Always prefers the method from `x-init` over automatic method.


calebporzio:
There are plans on adding something like this in the future. But not at the moment.  Thanks for contributing!

PR #181, opened 2020-02-16
Fix version number in CDN link for IE11 version

philippbosch:


philippbosch:
It has now been fixed for the 2.0 release.

PR #183, opened 2020-02-17
Nested x-for loops

### LINKED ISSUE #158: Access nested items in x-for
This issue is to open a discussion about on the problem found here: https://twitter.com/devgummibeer/status/1223415287447461889?s=21

I replied to this tweet with an alternative syntax for this too, see here: https://twitter.com/ryangjchandler/status/1223664594784264194?s=21

I personally think this syntax would be pretty clean and takes a PHP-approach with the syntax too, using the 'as' keyword. Removed the need for nested x-fors, when you only need to do the second loop on a single nested item. In theory, this could be taken a lot further with deeply nested items.

I am playing around with the x-for code currently to find a good way of achieving this syntax, and if it's something people would find useful, I'll throw a PR in for Caleb to ponder over. Have a good one guys! 🤙🏼

SimoTod:
Fixes #158 

Please double check the added loops to make sure they make sense.

The `extraVars` variable wasn't passed to the handleForDirective function call.
Passing the variable fixes the simple case where a nested x-for is not at the same level of its parent loop  in the DOM.
For example
```html
<template x-for="item in items">
    <div>
        <template x-for="tag in item.tags">
            <span x-text="tag"></span>
       </template>
    </div>
</template>
```

When 2 loops append elements to the same DOM node, though, the 2 cursors conflict and the second x-for tries to update/delete elements created by the loop.
For example
```html
<template x-for="item in items">
    <template x-for="tag in item.tags">
        <span x-text="tag"></span>
   </template>
</template>
```
To prevent this behaviour, I've introduced a new variable ($depth) so we can skip/ignore elements created by a different x-for loop.

I've also changed a couple of default values where I believe we want to set them to an empty object so I don't need to perform additional check to make sure the variable is defined.

---
Note that, at the moment, given
```html
<template x-for="item in items">
    <template x-for="tag in item.tags">
        <span x-text="tag"></span>
        <span x-text="item.title"></span> <!-- error -->
   </template>
</template>
```
the nested item doesn't have access to the parent variable (item). 
If we are happy with this implementation, we can add that improvement in a separate PR but I didn't want to make this one too heavy to review.

Thanks to @ryangjchandler for giving a couple of useful hints on how to tackle this issue.


SimoTod:
Hi @calebporzio when you have a sec, let me know if this PR is useful and I'll update it otherwise we can close it, it's not a problem. 👍

samadadi:
Hi. I am curious about "x-for" directive. Is there a reason that alpine developers did not use "x-for" directive in parent tags instead of "template" tag. I appreciate if somebody explain to me, thanks. exmaple: 
`
<ul x-for="user in users">
    <li x-text="user.name" x-bind:key="user.id"></li>
</ul>
`

jkr06:
FWIW, I am also interested in this. I know this is not a full Vue replacement, but in a project I am working on, there was 1 component which was using Vue where the data comes as json, with some nested objects. Without nested x-for, we cannot replace that part with Alpine. 
I feel like with nested x-for most of the basic use cases will be covered by this library, although I can understand "basic use cases" will mean different things for different people.

HugoDF:
The reason I believe `template` is used is because it supports DocumentFragment on its content property, which means a lot less wrangling of createElement calls and so on.

SimoTod:
@samadadi It's a design choice.
Alpine doesn't have a Virtual DOM so it needs to save in your light DOM all the info.
If it used a normal item, the content would be rendered and visible so it means that, on DOM ready, Alpine would need to detach the html, create a documentFragment, save it somewhere (probably on the element having x-for) and then doing exactly the same thing as now (but with more overhead).
Being an open source project, choices need to be made on design topics like that.

src/directives/for.js
```diff
@@ -1,10 +1,13 @@
 import { transitionIn, transitionOut, getXAttrs } from '../utils'
 
-export function handleForDirective(component, el, expression, initialUpdate) {
+export function handleForDirective(component, el, expression, initialUpdate, extraVars) {
     if (el.tagName.toLowerCase() !== 'template') console.warn('Alpine: [x-for] directive should only be added to <template> tags.')
 
     const { single, bunch, iterator1, iterator2 } = parseFor(expression)
 
+    // Get the depth level for the current iteration
```
SimoTod: required when we use
```html
<template x-for="...">
    <template x-for="...">
         ...
    </template>
</template>
```
because elements from different loops end on the same DOM level and we need to differentiate apart.

DanielDarrenJones:
Really hope this get merged the lack of nested for loops is what’s holding us back from jumping into alpine across more projects

calebporzio:
Thanks so much for the implementation. Instead of supporting directly nesting `<template>` tags, we are just going to force a wrapping element around each template.

This will be working and available in the next release.

Here's the new PR for the fix: #316

Issue #187, opened 2020-02-20
Unnecessary implementation


flash-me:
https://github.com/alpinejs/alpine/blob/f550d081e774841a8f97a496d3586bc2f57ec760/src/utils.js#L14-L24

Are there any valuable reasons for this? 

To obtain only unique values for a given array `arr` you can just do `arr = Array.from(new Set(arr))`. 

Supported by all major browsers + IE11

cheers 
flash :zap:

yet-another-dev:
Looks to me like `new Set(iterable)` is _not_ supported by IE11 (unfortunately!): 

![image](https://user-images.githubusercontent.com/56443891/75031947-4041cd00-545c-11ea-8cb7-e79f151b6e0b.png)

Source: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/Set#Browser_compatibility

flash-me:
Ouh. But still
 `const set = new Set(); 
 arr.forEach(v => set.add(v));
return Array.from(set);`

SimoTod:
The solution using Set should be faster, too. 

[...]

PR #188, opened 2020-02-21
improve array unique values snippet

julioocamargoo:
Improvement based on [#187](https://github.com/alpinejs/alpine/issues/187).

Thanks, Flash!

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -377,3 +377,7 @@ export function deepProxy(target, proxyHandler) {
 function isNumeric(subject){
     return ! isNaN(subject)
 }
+
+export default {
+    arrayUnique,
+}
--- a/src/utils.js
+++ b/src/utils.js
@@ -377,7 +377,3 @@ export function deepProxy(target, proxyHandler) {
 function isNumeric(subject){
     return ! isNaN(subject)
 }
-
-export default {
-    arrayUnique,
-}
--- a/src/utils.js
+++ b/src/utils.js
@@ -12,15 +12,7 @@ export function domReady() {
 }
 
 export function arrayUnique(array) {
-    var a = array.concat();
-    for(var i=0; i<a.length; ++i) {
-        for(var j=i+1; j<a.length; ++j) {
-            if(a[i] === a[j])
-                a.splice(j--, 1);
-        }
-    }
-
-    return a;
+    return Array.from(new Set(array))
 }
 
 export function isTesting() {
--- a/test/utils.spec.js
+++ b/test/utils.spec.js
@@ -1,0 +1,7 @@
+import utils from '../src/utils'
+
+test('utils/arrayUnique', () => {
+    const arrayMock = [1, 1, 2, 3, 1, 'a', 'b', 'c', 'b']
+    const expected = utils.arrayUnique(arrayMock)
+    expect(expected).toEqual([1, 2, 3, 'a', 'b', 'c'])
+})
--- a/test/utils.spec.js
+++ b/test/utils.spec.js
@@ -1,7 +1,7 @@
-import utils from '../src/utils'
+import { arrayUnique } from '../src/utils'
 
 test('utils/arrayUnique', () => {
     const arrayMock = [1, 1, 2, 3, 1, 'a', 'b', 'c', 'b']
-    const expected = utils.arrayUnique(arrayMock)
+    const expected = arrayUnique(arrayMock)
     expect(expected).toEqual([1, 2, 3, 'a', 'b', 'c'])
 })


src/utils.js
```diff
@@ -12,15 +12,7 @@ export function domReady() {
...
-    var a = array.concat();
-    for(var i=0; i<a.length; ++i) {
-        for(var j=i+1; j<a.length; ++j) {
-            if(a[i] === a[j])
-                a.splice(j--, 1);
-        }
-    }
-
-    return a;
+    return Array.from(new Set(array))
```
SimoTod: this probably doesn't work in IE11. Please see the thread.

src/utils.js
```diff
@@ -385,3 +377,7 @@ export function deepProxy(target, proxyHandler) {
 function isNumeric(subject){
     return ! isNaN(subject)
 }
+
+export default {
+    arrayUnique,
+}
```
SimoTod: We are already exporting that function and I'm not sure that arrayUnique should be the default for this file. It's probably tidier if we don't add these lines.
You can use `import { utils } from '../src/utils'` in your test

julioocamargoo: I actually wanted to export a object with all named exports, as we can see on many projects. However my changes affect only arrayUnique and so I didn't export all... I still find more interesting export like I did because the test is an example where I (should) use all methods and in this case the import will be tidier, but I can change the import in test as well if you prefer

src/utils.js
julioocamargoo: I didn't really pay attention on that. I'll update PR with the thread suggestion. Thanks!

src/utils.js
SimoTod: It's more down to what Caleb prefers, but I think we just need to keep it consistent (one way or the other one).
This is the kind of thing that falls in the "opinion" category, another dev could say that they really love the other syntax but, at the end, there's really little difference between them. 
I personally don't have strong feeling pro or against both ways.

src/utils.js
flash-me: I'm not sure about not using the `new Set(array)` syntax. 
There is more code non-compatible to IE11. e.g. all the usage of [spread syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax). <br>
- [here](https://github.com/alpinejs/alpine/blob/master/src/utils.js#L67)
- [here](https://github.com/alpinejs/alpine/blob/master/src/utils.js#L68) 

or [Array.includes](https://github.com/alpinejs/alpine/blob/master/src/utils.js#L127), which is also not supported in IE11
- [here](https://github.com/alpinejs/alpine/blob/master/src/utils.js#L127)
- [here](https://github.com/alpinejs/alpine/blob/master/src/utils.js#L151)

You may ensure compatibility for IE11 either **everywhere** or **nowhere**. 

cheers
flashe :zap:

SimoTod: @flash-me
Babel convert the spread syntax when compiling but it doesn't convert `new Set(iterator)`.
If you look at the compiled IE11 code in the dist folder, you won't find any trace of the spread operator.

For `Array.includes`, Alpine uses a polyfill. We could probably use another polyfill for `new Set(iterator)` if ii exists but it has to be part of this PR, imo.

Since `const set = new Set(); arr.forEach(v => set.add(v)); return Array.from(set);` is already supported by IE11, I think we should opt for this syntax, to be fair.

flash-me: tbh I didn't look into the dist folder. Shame on me 😇

And the opt-out shouldn't be that different much of what the browsers will do internally.

Since all I do was to help, I'm totally OK with this. 👍🏽 

cheers 
flash :zap:

julioocamargoo: Now that you commented I realize that we were already using core-js polyfill and aparently It do the job about Set. I don't know how Alpine is using core-js but maybe we already can use new Set(array)

SimoTod: @julioocamargoo If the core-js supports it, we are good. You don't need to add anything else, rollup should already include the polyfill for you.

julioocamargoo: I verified and new Set([]) works fine on IE with core-js. I've never used rollup but looking in the docs and [rollup-ie11.config.js](https://github.com/alpinejs/alpine/blob/master/rollup-ie11.config.js) I think that problem was solved as @SimoTod said, so I'll mantain this code.

src/utils.js
julioocamargoo: @calebporzio can you help here? I agree with @SimoTod about it, so what you think? We'll avoid export default on the project?

calebporzio: yeah, let's avoid it. We can stick with the current pattern we've established in utils.js. Thanks!

test/utils.spec.js
```diff
@@ -0,0 +1,7 @@
+import utils from '../src/utils'
```
HugoDF: Without the default export in utils, you could use
```suggestion
import { arrayUnique } from '../src/utils'
```

julioocamargoo: Yes, I were waiting for Caleb's confirmation. Thanks!

calebporzio:
I like it! Thanks!

Issue #161, opened 2020-02-04
Proxies break Promises


overheadhunter:
I don't quite know why there are proxies everywhere, but I think it is to register observers to properties that are relevant for bindings.

However, this seems to cause problems when invoking functions on objects that are proxied, which rely on promises:

```
Method Promise.prototype.then called on incompatible receiver [object Object]
    at Proxy.then (<anonymous>)
```

To better illustrate this problem, I've created [this minimal example on jsfiddle](https://jsfiddle.net/anzobhmj/).

So far, the only workaround I know of, is to only call functions on unscoped (i.e. globally scoped) objects such as `window.obj.fnUsingPromises()` instead of `obj.fnUsingPromises()`, which comes at the price of losing all possibilities to use data bindings.

Pixelrobin:
For the record, returning promises from within an object's method works. For me most use cases (api request, etc) wouldn't inline a promise but use a method anyways: https://jsfiddle.net/pgyodjw0/1/

```javascript
class MyObj {
  getBarValue() {
    return new Promise((resolve, reject) => {
      resolve(42);
    });
  }

  bar() {
    this.getBarValue().then(val => {
      console.log("bar", val);
    });
  }
}
```

overheadhunter:
> For the record, returning promises from within an object's method works. For me most use cases (api request, etc) wouldn't inline a promise but use a method anyways:

Interesting, but I don't think it covers all use cases. Sometime it makes sense to run a Promise once and reuse it if it is already resolved. Any idea how to do this without using fields (cause it doesn't seem to work with alpine) or globally-scoped variables (because it is kind of an ugly workaround)?

SimoTod:
@overheadhunter Probably not the answer that you wanted to hear...

It's fixable just adding something like in component.js at line 109
```javascript
                // If property is a Promise, just return it.
                if (target[key] && Promise.resolve(target[key]) === target[key]) return target[key]
```
 but I'm not sure that a promise should be assigned to a variable in the alpine context: It wouldn't be reactive and would give to other user the false impression that it could do something magical.

It seems more sensible to create Promises when you actually use it, either in a different function as suggested by @Pixelrobin or directly in your bar function.

```javascript
  bar() {
    const barValue = new Promise((resolve, reject) => {
      resolve(42);
    });
    barValue.then(val => {
      console.log("bar", val);
    });
  }
```

overheadhunter:
[...]

PR #192, opened 2020-02-22
Escaping RegExp from Proxy

amrikasir:
when I using regex test or regex exec `this.condition.regex.test('example'),` I've got `RegExp.prototype.exec called on incompatible receiver [object Object]`

if anyone have a better idea, I hope for some advices


SimoTod:
Good point @amrikasir , but we can't bind target for every function because the component would partially lose its reactivity.
Your best option is to check if target is an instance of RegExp and in that case applies your patch.
Instead of doing it in the getter, it probably makes more sense to do something in util > deepProxy (so if the type is RegExp you won't wrap it in a Proxy).

amrikasir:
@SimoTod when do it in utils > deepProxy, it wont wrap in Proxy. it mean, it won't reactive when we want to change patterns. what if we check the instanceof target in component > wrapDataInObservable ?

SimoTod:
@amrikasir
Possibly but I believe that RegExp is immutable so you are actually creating a new instance and assigning it to the parent object which is still a proxy.
I'm typing it from the top of my head so I might be wrong, maybe an example would help. 
You probably need to test a few solutions and see which one works without bugs.

amrikasir:
sure, I'll. I hope that a better idea for this issue. thanks @SimoTod

amrikasir:
@SimoTod it works in Util or in Component, so I put it in Util like you said.

it works :smiley:

SimoTod:
Sorry, I know it was my suggestion but we probably need a better and more generic solution since Date objects have the same problem (see #206) and there was an old issues about Promises (#161) which is the same issue again. Given that it will likely apply to other native object types, I don't think it's a good idea to add an if for each type any more. 
Also, Date objects have setters method (i.e. setSeconds()) which should probably be reactive.
Any thoughts?

tlgreg:
What should be deep proxied btw?

Currently, some native objects are blacklisted, but otherwise, the conditions letting through any object.

My first thought was to do a plain object check, but that also excludes arrays, which are deep-proxied atm (could be an exception).

Also, the current conditions letting through things like `Map` and `Set` instances too, but their items are not available as enumerable props so the `for ... in` loop does nothing.

SimoTod:
@tlgreg It's an option, we could use Proxies only for plain objects `{}` and Arrays.
The only issue here is that if the object has a setXXX method and you call it, it won't update the component which may or may not be confusing.

For example, assuming that only plain objects are wrapped in a proxy
```html
<div x-data="{ foo: new Date() }">
  <span x-text="foo"></span>
  <button @click="foo.setDate(foo.getDate() + 1)">+</button>
  <button @click="foo.setDate(foo.getDate() - 1)">-</button>
</div>
```
clicking + and - won't update the content of `<span>`.

Good point about Map and Set, they should probably work as Array does. 
I think we need a different issue/PR for this but it's probably doable with checking if an object is iterable and using `for ... of` or some similar approach.

SimoTod:
An alternative would be to use the 'apply' Proxy trap to catch the method call, bind it to the right context and trigger a component refresh. (something like https://github.com/alpinejs/alpine/compare/master...SimoTod:feature/native-objects?expand=1 https://codepen.io/SimoTod/pen/jOPmoeN).
There is a small performance impact because methods that don't change the object state will trigger a pointless refresh but I'm not sure we can avoid it.

amrikasir:
I think we can make it generic..

we should escaping every object when it is instanceof it's contructor

```javascript
// checking RegExp
/[a-z]+/ instanceof /[a-z]+/.constructor // true

// checking Date Type
var dateWithNew = new Date()
var datewithoutNew = Date()

datewithoutNew instanceof datewithoutNew.constructor // false
dateWithNew instanceof dateWithNew.constructor // true

// checking Promise
var ObjectPromise = new Promise(() => {})
ObjectPromise instanceof ObjectPromise.constructor // true
`

what should we do just escaping it from Proxy ??
how about that ?

amrikasir:
btw, I just test instanceof checking in browser console.

calebporzio:
Thanks for this, but it should be fixed with the new reactivity core in v2.1.0

PR #194, opened 2020-02-23
Fix mutation observer reinitializing element(s) in x-if within x-for usecase

### LINKED ISSUE #189: ReferenceError xx is not defined with x-for and x-if combination and dynamic content
An exception is thrown when the `x-data` is updated and re-evaluated.

Using this structure:

``` html
<div x-data="{ items: [...] }">
    <div>
      <template x-if="...">
        ...
      </template>
      <template x-if="...">
        ...
      </template>
    </div>
  </template>
</div>
```

When a new item is added and the `x-data` updated, it will render OK but the exception `Uncaught ReferenceError: item is not defined` is seen in the console.

Here's an example: https://codepen.io/bvalentino/pen/BaNQKZo

jokarz:
Fixes #189.

Node(s) handled by the if-directive triggers mutation observer to reinitialize the handled node(s). 

In the case of ```x-if``` within ```x-for```, the node is reinitialized without additional value (provided by ```x-for```) resulting in eval reporting undefined object even though it shouldn't be

### DIFF
--- a/src/component.js
+++ b/src/component.js
@@ -322,7 +322,7 @@ export default class Component {
 
                 if (mutations[i].addedNodes.length > 0) {
                     mutations[i].addedNodes.forEach(node => {
-                        if (node.nodeType !== 1) return
+                        if (node.nodeType !== 1 || node.__x_inserted_me) return
 
                         if (node.matches('[x-data]')) {
                             node.__x = new Component(node)


merthurturk:
@calebporzio any plans to merge this PR? `x-if` directives inside `x-for` directives break the variable context.

calebporzio:
Perfect, thank you @jokarz!

PR #202, opened 2020-02-27
Fix a few typos

Chris1904:
Thanks for Alpine! Love using it!

I went through the source to finally understand how it works underneath and came across a few typos. Figured I could send in a PR for this.

Thanks again 👍🏼


calebporzio:
Thanks!

PR #204, opened 2020-02-27
Added issue template

nainardev:


calebporzio:
Thanks for the effort, going to close this for now though. May use later.

PR #205, opened 2020-02-27
Added CodePen exampes

nainardev:


README.md
```diff
@@ -14,7 +15,7 @@ Think of it like [Tailwind](https://tailwindcss.com/) for JavaScript.
 
 **From CDN:** Add the following script to the end of your `<head>` section.
 ```html
-<script src="https://cdn.jsdelivr.net/gh/alpinejs/alpine@v2.0.1/dist/alpine.js" defer></script>
+<script src="https://cdn.jsdelivr.net/gh/alpinejs/alpine@latest/dist/alpine.js" defer></script>
```
SimoTod: My 2 pence: This is a bit dangerous if, at some point, there's a major release involving BC. 
People using this tag in production will suddenly have their website broken but they won't notice until someone reports it. Also, since the js is cached, when there is a new release any new feature won't be available in the first 24h.
I think it would confuse people and it would cause an increment of false issues.

nainardev: You can edit the PR. I'm now unable to

nainardev:
removed latest flag

README.md
the94air: I think this might work. It seems to be supported by JSdelivr very well.
```
https://cdn.jsdelivr.net/npm/alpinejs@2.x.x/dist/alpine.js
```

nainardev:
Everything seems good! Ready to merge

README.md
calebporzio: Thanks for the suggestion @the94air - will add this to the docs instead.

calebporzio:
Thanks for creating these. Going to hold off on adding them though.

nainardev:
Why??

README.md
SimoTod: Doesn't it have the same cache issue, though? The CDN updates only every 24h, do they have a public api to flush the cache?

PR #210, opened 2020-02-29
Typo: `tempate` -> `template` in x-for warning text

HugoDF:
I noticed that the warning would log out with `<tempate>` instead of `<template>`, so here's the fix 😄


HugoDF:
Not sure who apart from @calebporzio has merge permissions

calebporzio:
Thanks!

PR #215, opened 2020-03-01
refactor(component): code and style

nandordudas:


calebporzio:
Thanks for the effort.

Unfortunately, code style PRs like these are unideal. There are some style changes I agree with, and others I don't, so I would have to manually go through and filter out the ones I don't want.

Thanks again, but going to close.

PR #229, opened 2020-03-04
Feature: Support all HTML boolean attributes as per WHATWG HTML Standard

### LINKED ISSUE #223: Feature: support more/all boolean attributes
Follow up from https://github.com/alpinejs/alpine/issues/220, where the "open" boolean attribute was not implemented as such.

> A number of attributes are boolean attributes. The presence of a boolean attribute on an element represents the true value, and the absence of the attribute represents the false value.
>
> as per the HTML spec: https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attributes

The following seems to be a table of boolean attributes although it might be worth cross-referencing with the HTML spec: https://github.com/iandevlin/html-attributes/blob/master/boolean-attributes.json

HTML spec attributes table: https://html.spec.whatwg.org/multipage/indices.html#attributes-3:boolean-attribute

If I can get signoff from @calebporzio I'm happy to contribute this.

HugoDF:
Closes #223 

- Add `isBooleanAttr` helper (uses an Array of HTML boolean attributes as per https://html.spec.whatwg.org/multipage/indices.html#attributes-3:boolean-attribute)
- Update "bind" to use `isBooleanAttr` to check for Boolean Attributes
- Add relevant tests
- Update relevant documentation (README)

@calebporzio let me know your thoughts on RegEx vs array.includes. From the benchmarking I did, it seems `Array#includes` is marginally faster when the match is somewhere in the middle of the array but is slower when there is no match.

Benchmark output on Firefox:
```
RegExpInstance#test - match x 3,210,834 ops/sec ±0.81% (64 runs sampled)
Array#includes - match x 3,229,319 ops/sec ±0.80% (61 runs sampled)
RegExpLiteral#test - match x 3,158,156 ops/sec ±0.94% (63 runs sampled)
RegExpInstance#test - no match x 3,072,680 ops/sec ±0.88% (62 runs sampled)
Array#includes - no match x 2,794,591 ops/sec ±0.91% (62 runs sampled)
RegExpInstance#test - no match x 3,161,692 ops/sec ±0.41% (63 runs sampled)
```

See the code: https://github.com/HugoDF/alpinejs-playground/blob/master/pages/benchmark-array-includes-vs-regex.html#L56-L109


nandordudas:
My experience with regex is very confused. I love it, but sometimes i hate it. It this case there are a lot of words. Regex will try match one of them. But a more specific regex `(?:(a|b))` - with groups is a bit faster. More faster with lookbehind or lookahead. The `include` function needs one argument, hoe cool is that.
A good regex code is very fast, look for a password strength example - it stops when not enough length, missing chars, etc.

HugoDF:
@nandordudas yeah, I've implemented it both ways either way 🤷🏼‍♀️ just a reverted commit away

SimoTod:
If you go for a regular expression, you'll probably need to add `^(` after `/` at the beginning and `)$` at the end before `/` since you want only exact matches otherwise custom stuff like `x-bind:data-default`, which may or may not be boolean, would be treated as a boolean field.
And maybe the `i` flag for case insensitive.

Caleb's call but, for the sake of readability, I would use array.includes if the difference is minimal.

Great PR 👍

calebporzio:
Agreed on the readability of .includes

Also, please remove the commented out `expect()`s, and the parts of the test that drive them.

HugoDF:
@calebporzio fixed to use "includes" & good catch on the commented out assertions, for whatever reason JSDom needed a bit of a push to behave with attributes on non `input`-s

All fixed now 😄

calebporzio:
Thanks a bunch for this!

Issue #231, opened 2020-03-05
Checkboxes can't have a distinct `value`


philippbosch:
In plain HTML, `<input type="checkbox" />` elements _can_ have a `value` attribute:

```html
<input type="checkbox" name="foo" value="bar" />
```

If they do, the value that is sent to the server if the checkbox is checked will be the value of the `value` attribute:

```
foo=bar
```

You may also omit the `value` attribute:

```html
<input type="checkbox" name="foo" />
```

In that case the value sent to the server if the checkbox is checked will be `"on"`:

```
foo=on
```

So far, so good. The checkbox behaviour in Alpine is a little bit odd IMHO. Given this Alpine component:

```html
<form x-data="{value: 'bar'}">
    <input type="checkbox" name="foo" :value="value" />
</form>
```

The value for `foo` will be set to `"on"` instead of `"bar"` which I think is wrong.

I made a quick demo showing the differnt behaviours for Alpine, Vue and plain HTML:
https://ubiquitous-cautious-singer.glitch.me/

Vue.js and plain HTML set the value to the given `value` HTML attribute, and I think Alpine should do the same.

PR #232, opened 2020-03-05
Allow checkboxes to have distinct values

philippbosch:
This is a fix for #231 but I'm not happy with the `if (typeof value === 'string')` bit. Any feedback is welcome and appreciated.


src/directives/bind.js
```diff
@@ -27,6 +27,9 @@ export function handleAttributeBindingDirective(component, el, attrName, express
             } else {
                 el.checked = !! value
```
HugoDF: Pretty sure instead of doing the "if typeof string" you could just stick the value in here

HugoDF:
I don't think this only applies to "checked". We might need to extend to to all booleans where we're setting the value to `!! value`.

The HTML spec states that "presence" of "Boolean attributes" translates to their truthiness (the value actually doesn't matter).

I believe you can set any Boolean attribute to arbitrary values, they will be treated as "on"/"enabled" as far as HTML is concerned but in your case, their form submission value would be whatever you've set it to.

I've been doing some related work (for Boolean attributes) in https://github.com/alpinejs/alpine/pull/229

src/directives/bind.js
philippbosch: I think if the value is e.g. `true` we don't want to set the `value` attribute to `"true"` which is not what we want. But I think the `el checked = !! value` also needs to be changed.

HugoDF:
Ignore above comment

src/directives/bind.js
HugoDF: Sorry my bad I misread value vs checked

test/bind.spec.js
```diff
@@ -228,3 +228,19 @@ test('binding supports short syntax', async () => {
...
+            <input type="checkbox" name="stringCheckbox" :value="stringValue" />
+            <input type="checkbox" name="trueCheckbox" :value="trueValue" />
+            <input type="checkbox" name="falseCheckbox" :value="falseValue" />
+        </div>
+    `
+
+    Alpine.start()
+
+    expect(document.querySelector('input[name="trueCheckbox"]').value).toEqual('on')
+    expect(document.querySelector('input[name="falseCheckbox"]').value).toEqual('on')
```
HugoDF: Not sure I understand this.

So if we bind false to the checkbox's value, we expect it to equal "on"?

philippbosch: Yes, this is how it works in plain HTML. The value is `on` but it would not have a `checked` flag. If you submit the form to a server the browser will only send the values for checkboxes that are checked and omit those that aren‘t.

calebporzio:
Thanks!

PR #234, opened 2020-03-05
Improve x-attributes regexp

philippbosch:
For debugging reasons I tried to disable an `x-show` by changing it to `blah-x-show` and was surprised it was still parsed like `x-show`. 

The reason was that the regular expression used to filter all attributes was a bit too loose. My suggestion implemented in this PR is to:

1. Make sure the attribute _starts_ with `x-` and does not allow any extraneous characters before the `x`. This is done by adding a _start of string anchor_ to the beginning: `/^x-…` instead of `/x-…`
2. Make sure the directive type is matched exactly and does not allow any extraneous trailing characters  (e.g. `x-showblah`). This is done by adding a _word boundary_ (`\b`) to the end of the regexp. Using a _word boundary_ instead of an _end of string anchor_ (`$`) allows for values (`:click`) and modifiers (`.prevent`).

Any suggestions welcome and appreciated.


calebporzio:
Thanks!

PR #236, opened 2020-03-05
Feature: bind to style with object, array and string merging/override support

### LINKED ISSUE #213: Inline Styles
Is it possible to have inline styles in AlpineJS? I've tried something on CodePen, but nothing was working. Consider the following code:

```
<div class="box" :style="{ backgroundColor: 'red' }"></div>
```

Assuming `.box` has `width: 100px;`, `height: 100px`, and a `background-color: 'blue'`, the code above should make the box the color of `red`, but it does not.

I've tried the same code in VueJS and it did work with inline styles. Here is the same code in VueJS:

```
<div id="app">
  <div class="box" :style="{ backgroundColor: 'red' }"></div>
</div>
```

Is there an extra step to consider when working with inline styles in AlpineJS? I didn't see anything mentioned about inline styles in the [GitHub documentation](https://github.com/alpinejs/alpine).

Thanks for the help!

HugoDF:
Closes #213 Adds Vue-style binding support for the "style" attribute.

Style binding supports:
- string definition with merging/override support
  - eg. `<div style="display: inline" :style="'padding: 10px'"></div>` -> `display: inline; padding: 10px` applied
- array definition with merge/override support
  - eg. `<div style="display: inline" :style="['display: hidden; width: 200px;', 'height: 20px;']"></div>` -> `display: hidden; width: 200px; height: 20px;`
- object definition with override support
  - eg. `<div style="padding-top: 0" :style="{display: 'hidden', paddingTop: '15px', borderColor: 'teal'}></div>"` -> `display: hidden; padding-top: 15px; border-color: teal`

Update README with style object/array support.

~~Unrelated change (to tests): fixes missing assertion calls in `class` bind tests (see https://github.com/alpinejs/alpine/pull/236/commits/6d9e2a091a5d1c68debc15cb20c10afec3dcf197), I'm happy to submit this as a separate PR.~~ Submitted as https://github.com/alpinejs/alpine/pull/360.

### DIFF
--- a/README.md
+++ b/README.md
@@ -236,6 +236,24 @@ For example:
 
 In this example, the "hidden" class will only be applied when the value of the `foo` data attribute is `true`.
 
+**`x-bind` for style attributes**
+
+`x-bind` behaves a little differently when binding to the `style` attribute.
+
+For styles, you pass in an object who's keys are style names, and values are either CSS property values (strings) or falsy. If the value is falsy the style will not be applied.
+
+For example:
+`<div x-bind:class="{ display: foo && 'inline', fontSize: 20 + 'px' }"></div>`
+
+In this example, the "display" style will only be applied when the value of the `foo` data attribute is `true`.
+
+`style` bindings also support arrays and will merge/override rules.
+
+For example:
+`<div style="display: inline" x-bind:style="['display: hidden; width: 200px;', 'height: 20px;']"></div>`
+
+In this example, the "display" style will be set to "hidden", "width" will be "200px" and height will be "20px"
+
 **`x-bind` for boolean attributes**
 
 `x-bind` supports boolean attributes in the same way that value attributes, using a variable as the condition or any JavaScript expression that resolves to `true` or `false`.
--- a/dist/alpine-ie11.js
+++ b/dist/alpine-ie11.js
@@ -4468,10 +4468,48 @@
     }
   });
 
-  var HAS_SPECIES_SUPPORT$2 = arrayMethodHasSpeciesSupport('splice');
-  var USES_TO_LENGTH$7 = arrayMethodUsesToLength('splice', { ACCESSORS: true, 0: 0, 1: 2 });
+  var HAS_SPECIES_SUPPORT$2 = arrayMethodHasSpeciesSupport('slice');
+  var USES_TO_LENGTH$7 = arrayMethodUsesToLength('slice', { ACCESSORS: true, 0: 0, 1: 2 });
 
+  var SPECIES$6 = wellKnownSymbol('species');
+  var nativeSlice = [].slice;
   var max$1 = Math.max;
+
+  // `Array.prototype.slice` method
+  // https://tc39.github.io/ecma262/#sec-array.prototype.slice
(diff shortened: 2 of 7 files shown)


HugoDF:
@calebporzio this has been update with latest master changes and is ready for you to check out 👍

HugoDF:
@calebporzio any idea when this PR might be integrated? I'm aware it's a chunky one but it does bring support for a feature you mentioned just not having time to tackle.

atomgiant:
I would also really like to see this functionality in Alpine.  In my case I am using it for a color picker to dynamically adjust the selected color background.  Here is a CodePen which shows it not working if that helps https://codepen.io/atomgiant/pen/YzyzYxd

HugoDF:
I've cleaned this PR up for review 😄

calebporzio:
Thanks @HugoDF, but going to hold off on adding "object" syntax for the style attribute for now.

May remove the object syntax from the class attributes in v3 - undecided yet. I will put a note in the Roadmap and reference this PR.

Thanks for the contribution!

OwenMelbz:
@calebporzio would you be able to clear up the thoughts around removing object support from the `:class` ? it seems a massively common method to conditionally apply classes?

```html
<input :class="{'input-valid': isValid, 'input-invalid': !isValid && isDirty}" />
```

HugoDF:
@OwenMelbz the rationale is two fold, first is that it's achievable in userland with a util function that converts objects into an array (or string), second is that it's not implemented exactly in the same way as Vue and that has caused issues.

OwenMelbz:
Hey @HugoDF I appreciate the "issues caused" would add a complication.

So would that mean that if anybody wanted to use the syntax above they would need to create a global function that needs access to the data? Wonder if you could mockup what you mean for a real-world app, would it be something like...?


```html
<div x-data="{value: '', isValid: false, isDirty: false}">
    <input :class="mergeArrays(inputClasses($data), otherClassLogic())" />
</div>
```

```
window.mergeArrays = () => {
    let finalArray = [];

    arguments.forEach(arg => {
        finalArray = finalArray.concat(arg);
    });

    return finalArray;
}

window.otherClassLogic = () => {
    // blah do ajax lookup or something unique to this field
    return ['input-account-already-exists'];
}

window.inputClasses = ({isValid, isDirty}) => {
    let classes = [];

    if (isValid) {
        classes.push('input-valid')
    }

    if (!isValid && isDirty) {
        classes.push('input-invalid');
    }

    return classes;
}
```

Thanks!

HugoDF:
Hmm no it would have a similar API

`:class="objClass({ 'valid': valid, 'invalid': !valid })"`

Implementation would then be Alpine agnostic.

philwolstenholme:
@OwenMelbz one option might be to use the [Classnames package](https://www.npmjs.com/package/classnames) to compose the list of class strings. I've put together a small demo here: https://jsfiddle.net/philw_/ydvm8r1h/

Btw I found this issue while looking at the Roadmap, I've never used Alpine before but have been experimenting with adapting the dropdown from the README and have really enjoyed it!

PR #238, opened 2020-03-05
Documentation: explicitly set x-data functions on window when using a bundler.

### LINKED ISSUE #152: Importing as a separate script does not appear to work
This may be an issue with Parcel specifically, but I tried setting up the basic `x-data` example on Codesandbox. If the script is not inlined in the page, but is imported from a separate file, Alpine will not function.

See:
https://codesandbox.io/s/cranky-wozniak-rgn7s

HugoDF:
Closes #152 

Issue resolution mentioned it might be a good thing to document and I tend to agree.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ Include it in your script.
 import 'alpinejs'
 ```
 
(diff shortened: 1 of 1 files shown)


ryangjchandler:
Agree with this one. Easy one for people to get caught out by for sure.

Issue #190, opened 2020-02-21
Add eslint to ensure code quality


julioocamargoo:
Like editorconfig, I think eslint will help with this. What do you think?

Calinou:
Which style should be picked? Since the current codebase doesn't seem to use semicolons, going for Standard would make more sense. It's not as strict as the Airbnb style though.

thormeier:
Suggestion: A pretty robust setting I personally use for many projects is `vue/recommended` (https://eslint.vuejs.org/rules/) paired with prettier (https://prettier.io/docs/en/options.html):
```
// .eslintrc.json
{
  "root": true,
  "extends": [
    "plugin:vue/recommended"
  ],
  "plugins": [
    "prettier"
  ],
  "rules": {
    "prettier/prettier": "error"
  }
}
```

```
// .prettierrc
{
  "trailingComma": "none",
  "tabWidth": 2,
  "semi": false,
  "singleQuote": true
}
```

Some of the Vue rules are obviously ignored, but the standard JS stuff is pretty sane, imho. What do you think?

HugoDF:
I've had pretty good success with just prettier, fewer config settings, harder to break prettier rules (since it's a formatter)

HugoDF:
I've had pretty good success with just prettier, fewer config settings, harder to break prettier rules (since it's a formatter)

Calinou:
ESLint and Prettier detect a different range of issues, so I think we should be either using just ESLint or ESLint + Prettier, not just Prettier. ESLint catches "code smells" and poor programming practices in addition to some code style issues, whereas Prettier *only* focuses on code style.

HugoDF:
To be fair, with a project/codebase this size and this number of contributors ESLint/Prettier aren't necessary until @calebporzio gives write access to a bunch of other contributor.

Right now all PRs need to be merged by him, so he can decide whether there are any style/format issues or "code smells".

jonaskuske:
@thormeier Instead of setting `plugins`
` and `rules` manually, you also add `plugin:prettier/recommended`, just like with `vue`. Oh and as the third entry to `extends` you'll probably want to add `prettier/vue` to handle Vue-specific ESLint rules :)

HugoDF:
I'm going to close this for now, as mentioned, Caleb is the only one with write access to the repository so he ends up doing the linting himself before/after merging 😄

PR #247, opened 2020-03-07
Chore: "format" npm script with Prettier

HugoDF:
Relates to #190 

Creates a prettier config that causes relatively few changes (mostly wanted changes since there were random `;` and missing/extra spaces)

Feel free to close if this isn't something we need, but I would rather let a tool format my code than have to worry about it myself.

Creates "format" & applies it in separate commits for the source & the tests.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,5 @@
+{
+    "semi": false,
+    "singleQuote": true,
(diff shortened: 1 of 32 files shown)


calebporzio:
Going to hold off on this for now. Thanks!

PR #260, opened 2020-03-09
fix(IE11): Run observable-membrane through Babel for IE11 build

### LINKED ISSUE #259: IE11 - Expected Identifier Error
I'm getting an error 

```
Script1010: Expected Identifier
alpine-ie11.js (5918,9)
```

which points to the last line in the block

```
  /**
   * Copyright (C) 2017 salesforce.com, inc.
   */
  const { isArray: isArray$1 } = Array;
```

I tried adding in the head tag the following based on others receiving this error, but no luck:

```
  <meta http-equiv="X-UA-Compatible" content="IE=edge">
```

Wondering if anyone else is receiving this error as well?

HugoDF:
Closes #259 

- Switch to babel `includes` syntax for IE11 rollup build
- include "src/**" and observable membrane globs for babel compilation for IE11
- Build assets (doesn't seem to have

I'm going to need someone to test on IE11, can be done through CDN: https://cdn.jsdelivr.net/gh/HugoDF/alpine@fix-compile-membrane/dist/alpine-ie11.js

### DIFF
--- a/dist/alpine-ie11.js
+++ b/dist/alpine-ie11.js
@@ -5912,380 +5912,1454 @@
     }.bind(this);
   }
 
+  var nativeGetOwnPropertyNames = objectGetOwnPropertyNames.f;
+
+  var toString$1 = {}.toString;
+
+  var windowNames = typeof window == 'object' && window && Object.getOwnPropertyNames
+    ? Object.getOwnPropertyNames(window) : [];
+
+  var getWindowNames = function (it) {
+    try {
+      return nativeGetOwnPropertyNames(it);
+    } catch (error) {
+      return windowNames.slice();
+    }
+  };
+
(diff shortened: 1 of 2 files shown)


calebporzio:
Thanks @HugoDF, unfortunately, this is not the fix.

The current proxy-polyfill doesn't support some of the traps that `observable-membrane` uses.

I think the best solution ATM is just to use the old Proxy observable we implemented for the IE build, and keep observable-membrane for the main build.

It's kind of hacky, but honestly, it's the most solid thing I could come up with and ship quickly.

Thanks for this, closing.

HugoDF:
Yep this was just to fix the first error, using the hand rolled proxy (even with the bunch of edge cases that don't work 100%) seems fair enough 👍

PR #271, opened 2020-03-13
Don't react if the mutated key is array length

### LINKED ISSUE #250: Pushing to array under x-data will cause repeated rendering
Hi,

First i have to thanks you all for the great work :clap:. I am working on a **Shopping Cart** app that adds a list items to the cart then calculate the **Total** and **Tax**. To make it easier i will use a todo app to demonstrate my issues. Here is a codepen https://codepen.io/tomorrowsystem/pen/ZEGavzW?editors=1001

My issues are: 

1. When `x-text` directive is assigned to a function output the function gets call one on `x-data` update or delete which is correct while on `x-data` addition it gets called multiple times, i doubt it may be a BUG.

2. The `todoCount()` is a computed property. Is there a better way to get it other than assigning a function to `x-text` directive ?

3. After the `x-for` directive and the computed values finish rendering i want to do some DOM manipulation e.g. focus on an input. I can't use `$nextTick` as there is no event. So how can i listen to Render or Refresh or `$nextTick` at component level ?

Thanks.

ahmedkandel:
Fixes #250
Inside observable-membrane the `ReactiveProxyHandler` calls back `valueMutated` twice when adding a new index to an array, first time for the new index and second time unnecessarily for the change in array length which causes repeated component rendering.

### DIFF
--- a/src/component.js
+++ b/src/component.js
@@ -88,6 +88,9 @@ export default class Component {
                 // Don't react to data changes for cases like the `x-created` hook.
                 if (self.pauseReactivity) return
 
+                // Don't react if the mutated key is array length as it is a consequence of another mutated property like pushing to the array
+                if (Array.isArray(target) && key === 'length' ) return
+
                 debounce(() => {
                     self.updateElements(self.$el)
 
--- a/src/component.js
+++ b/src/component.js
@@ -88,8 +88,9 @@ export default class Component {
                 // Don't react to data changes for cases like the `x-created` hook.
                 if (self.pauseReactivity) return
 
-                // Don't react if the mutated key is array length as it is a consequence of another mutated property like pushing to the array
-                if (Array.isArray(target) && key === 'length' ) return
+                // Don't react if the mutated key is array length as it is a consequence of another mutated property
+                // like pushing to the array
+                if (Array.isArray(target) && key === 'length') return
 
                 debounce(() => {
                     self.updateElements(self.$el)
--- a/src/component.js
+++ b/src/component.js
@@ -99,7 +99,7 @@ export default class Component {
                     while (self.nextTickStack.length > 0) {
                         self.nextTickStack.shift()()
                     }
-                }, 0)()
+                }, 0, self)()
             },
         })
 
--- a/src/component.js
+++ b/src/component.js
@@ -99,7 +99,7 @@ export default class Component {
                     while (self.nextTickStack.length > 0) {
                         self.nextTickStack.shift()()
                     }
-                }, 0, self)()
+                }, 0)()
             },
         })
 
--- a/dist/alpine-ie11.js
+++ b/dist/alpine-ie11.js
@@ -4910,18 +4910,17 @@
     }
   }
   function debounce(func, wait) {
-    var timeout;
+    var context = arguments.length > 2 && arguments[2] !== undefined ? arguments[2] : this;
     return function () {
-      var context = this,
-          args = arguments;
+      var args = arguments;
 
       var later = function later() {
-        timeout = null;
+        context.debounce_timeout = null;
         func.apply(context, args);
       };
 
-      clearTimeout(timeout);
-      timeout = setTimeout(later, wait);
+      clearTimeout(context.debounce_timeout);
+      context.debounce_timeout = setTimeout(later, wait);
     };
   }
   function saferEval(expression, dataContext) {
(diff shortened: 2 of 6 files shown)


src/component.js
```diff
@@ -88,6 +88,9 @@ export default class Component {
                 // Don't react to data changes for cases like the `x-created` hook.
                 if (self.pauseReactivity) return
 
+                // Don't react if the mutated key is array length as it is a consequence of another mutated property like pushing to the array
+                if (Array.isArray(target) && key === 'length' ) return
```
HugoDF: Are we sure we want to do the `Array.isArray` check?

What about things like strings?

HugoDF:
Also, it would be cool to have a unit test for this.

src/component.js
ahmedkandel: `Array.isArray` is not that necessary but I think it is better to include it for two reasons:

1. To be self-explanatory about what we mean by length property. 
2. This fix is to reverse observable-membrane unwanted behavior which verifies first if `(key === 'length' && isArray(originalTarget))` So it is better to narrow down the fix to only this scenario.

Regarding strings will work with or without this fix as observable-membrane specifies this unwanted behavior to array length mutations.

Another side note, If the component x-data object contains a length property will also work as `(oldValue !== value)` will be true.

ahmedkandel:
> Also, it would be cool to have a unit test for this.

Do you think it is better to be inside mutations.spec.js or data.spec.js or just create a separate one?

src/component.js
HugoDF: Makes sense to me 👍

SimoTod: 👍 Strings are immutable object in js as well (string.length is readonly so it will never mutate).

Have you tried if other methods such as array.pop and array.shift are still triggering a refresh?

HugoDF:
Should be data.spec.js, 3-4 tests for push, pop, unshift, shift

src/component.js
HugoDF: You're right, I was thinking about maybe Set/Map but that stores the number of items under `.size`

I think that's beyond the scope of this PR though, should be a follow up PR

ahmedkandel: array.pop, shift, push, unshift are fine as the change in index length will be ignored by the fix. While the changes in the indexes by shift and unshift will cause repeated rendering for each single index change but ar lease one less rendering related to index length. The same thing for or reverse, splice, fill, ...

HugoDF: Cool, just missing tests to avoid this fix regressing then

ahmedkandel: Will do right now

HugoDF:
Is it possible to check the number of reactive updates when adding/removing elements from an array? Since that's the goal of this PR

ahmedkandel:
I think those unit tests are to make sure the fix will not break the current reactivity, But why not we also add a test to prove the fix itself.

What do you think this new test should be: specific to the fix or a generic test to count reactive updates in case of other properties mutations?

HugoDF:
Specific to this new code I think

HugoDF:
Nice way to write that test 👍👍👍

I haven't seen Caleb active on this repo for a bit but let's see if he can start getting some of these PRs merged

HugoDF:
Might be worth keeping open til the PRs are merged

HugoDF:
Isn't this complimentary to the other fix?

We might want to port the unit tests to check

ahmedkandel:
Hi @HugoDF,

This fix was a start point to reduce number of unnecessary components rendering here #250 but later on, as I explained in #272 we needed more improvements and then found the `debounce timeout` issue that needs another commit.

Anyway, I was working on the timeout issue so I will reopen the PR and add the commits here may be both fixes are needed. Who knows maybe in the future debounce get changed so this array.length fix remains valid.

Thanks

ahmedkandel:
Will proceed with more unit tests and commit

ahmedkandel:
Fixes #250 and #272

HugoDF:
I wouldn't pull that commit in here, this is a fix in its own right (due to length change, array mutation does cause 2 renders)

ahmedkandel:
Got that, added a new PR #276

ahmedkandel:
Salesforce have just replied [here](https://github.com/salesforce/observable-membrane/issues/40#issuecomment-599461205), I think it is better to close this PR

ahmedkandel:
Fixed by #276

Issue #272, opened 2020-03-13
Unnecessary multiple component refresh for single update


ahmedkandel:
Original I wanted to use a computed property but after reading #79 found that it is easy to call a method in `x-text` and all is done while still `x-watch` will be helpful addition.

Then I faced an issue of repeated rendering because inside the computed property I had to mutate some `x-data` properties e.g. inside `cartTotal()` i had to set `this.cart_tax` and `this.cart_discount`.

So in order to wait for the last mutated property then update once, I found the undocumented boolean property `self.pauseReactivity` and `updateElements($el)` method which will help a lot in controlling reactivity and repeated rendering.

**Q:** Is there a better approach to access those properties and methods other than `this.$el.__x.pauseReactivity` and `this.$el.__x.updateElements($el)`?

Also, I think Alpine needs to avoid repeated rendering, especially with a large set of data. Am I right or just obsessed with performance?

SimoTod:
Calling a function in x-text to check when a component is refreshed feels a weird. 

I know @ryangjchandler is working on a PR to emit events when something happens in your component lifecycle which will probably help you a lot.

About performance, it would be nice if the component refreshed only when required but it's not a big concern at the moment. 

The goal would be to have a lot of small components in your page so the overall size shouldn't be too relevant (this is one of the reasons why Alpine doesn't support passing the scope from a parent to a child component, we don't want people to create a god component using the body tag: that would affect performances a lot when Alpine needs to walk down the DOM).

cekvenich:
@SimoTod  I made example of x-data for you, you can run/view source & console here:
- https://cekvenich.github.io/alp

( But full source is https://github.com/cekvenich/alp ) 

Let me know if this helps an if you need more examples.

SimoTod:
@cekvenich I think you commented on the wrong issue 🤣, this one is not the ticket about custom elements.
Thanks, I'll have a look in the next few days. 👍

cekvenich:
@SimoTod Oooh. Should I delete all the comments and put in right place?

ahmedkandel:
@SimoTod Sure I like Alpine's approach not to create **a god component**  but to have small **sprinkled components**. While my request was a little bit different:

> Calling a function in x-text to check when a component is refreshed feels a bit weird.

I didn't understand what do you mean? Sorry, maybe my comment was not clear. Here is a CodePen to demo what I mean. https://codepen.io/tomorrowsystem/pen/xxGWzrp?editors=1000

[...]

PR #275, opened 2020-03-14
Performance improvements

SimoTod:
As mentioned in #272, Alpine refreshes a component many times for each update, especially when using arrays. Despite not being critical right now, this causes an unnecessary performance drop.
The reason sits in the debounce function which doesn't have a context and always instantiates a new timeout variable so the previous calls are never cancelled.

I've chosen to pass the context as a variable rather then using `.bind(self)` because I want to enforce it if the same function will be used in the future (it would be too easy to forget to bind the right context otherwise).

Big thanks to @ahmedkandel for pointing out the issue and suggesting the unit test.

### DIFF
--- a/test/data.spec.js
+++ b/test/data.spec.js
@@ -105,3 +105,22 @@ test('Proxies are not nested and duplicated when manipulating an array', async (
     document.querySelector('button').click()
     await wait(() => { expect(document.querySelector('span').innerText).toEqual('bar') })
 })
+
+test('component does not refresh multiple times when data changes', async () => {
+    window.renderCount = 0
+
+    document.body.innerHTML = `
+        <div x-data="{ items: [], test() {return ++renderCount}  }">
+            <button @click="items.push('ok')"></span>
+            <span x-text="test()"></span>
+        </div>
+    `
+
+    Alpine.start()
+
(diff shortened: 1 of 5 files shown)


test/data.spec.js
```diff
@@ -105,3 +105,22 @@ test('Proxies are not nested and duplicated when manipulating an array', async (
     document.querySelector('button').click()
     await wait(() => { expect(document.querySelector('span').innerText).toEqual('bar') })
 })
+
+test('component does not refresh multiple times when data changes', async () => {
+    window.renderCount = 0
+
+    document.body.innerHTML = `
+        <div x-data="{ items: [], test() {return ++renderCount}  }">
```
HugoDF: Ah this test again, very nice 💪💪

HugoDF:
Really good catch 🥳🥳

test/data.spec.js
SimoTod: Haha. I just copied Ahmed's test. All credits to him. ;)

ahmedkandel: Good job guys, I will close #271

SimoTod:
Fixed by #276

PR #276, opened 2020-03-15
Debounce timeout never gets cleared

### LINKED ISSUE #272: Unnecessary multiple component refresh for single update

ahmedkandel:
Fixes #272
Inside utils.js the debounce function declares timeout variable for each call which preventing the timeout from being cleared and causes repeated component rendering.

### DIFF
--- a/dist/alpine-ie11.js
+++ b/dist/alpine-ie11.js
@@ -4909,8 +4909,7 @@
       node = node.nextElementSibling;
     }
   }
-  function debounce(func, wait) {
-    var context = arguments.length > 2 && arguments[2] !== undefined ? arguments[2] : this;
+  function debounce(func, wait, context) {
     return function () {
       var args = arguments;
 
--- a/dist/alpine-ie11.js
+++ b/dist/alpine-ie11.js
@@ -4910,18 +4910,17 @@
     }
   }
   function debounce(func, wait) {
-    var timeout;
+    var context = arguments.length > 2 && arguments[2] !== undefined ? arguments[2] : this;
     return function () {
-      var context = this,
-          args = arguments;
+      var args = arguments;
 
       var later = function later() {
-        timeout = null;
+        context.debounce_timeout = null;
         func.apply(context, args);
       };
 
-      clearTimeout(timeout);
-      timeout = setTimeout(later, wait);
+      clearTimeout(context.debounce_timeout);
+      context.debounce_timeout = setTimeout(later, wait);
     };
   }
   function saferEval(expression, dataContext) {
@@ -6389,7 +6388,7 @@
               while (self.nextTickStack.length > 0) {
                 self.nextTickStack.shift()();
               }
-            }.bind(this), 0)();
+            }.bind(this), 0, self)();
           }
         });
         return {
--- a/dist/alpine.js
+++ b/dist/alpine.js
@@ -90,7 +90,7 @@
       node = node.nextElementSibling;
     }
   }
-  function debounce(func, wait, context = this) {
+  function debounce(func, wait, context) {
     return function () {
       var args = arguments;
 
--- a/dist/alpine.js
+++ b/dist/alpine.js
@@ -90,19 +90,17 @@
       node = node.nextElementSibling;
     }
   }
-  function debounce(func, wait) {
-    var timeout;
+  function debounce(func, wait, context = this) {
     return function () {
-      var context = this,
-          args = arguments;
+      var args = arguments;
 
       var later = function later() {
-        timeout = null;
+        context.debounce_timeout = null;
         func.apply(context, args);
       };
 
-      clearTimeout(timeout);
-      timeout = setTimeout(later, wait);
+      clearTimeout(context.debounce_timeout);
+      context.debounce_timeout = setTimeout(later, wait);
     };
   }
   function saferEval(expression, dataContext, additionalHelperVariables = {}) {
@@ -1290,7 +1288,7 @@
             while (self.nextTickStack.length > 0) {
               self.nextTickStack.shift()();
             }
-          }, 0)();
+          }, 0, self)();
         }
 
       });
--- a/src/component.js
+++ b/src/component.js
@@ -95,7 +95,7 @@ export default class Component {
                     while (self.nextTickStack.length > 0) {
                         self.nextTickStack.shift()()
                     }
-                }, 0)()
+                }, 0, self)()
             },
         })
 
--- a/src/utils.js
+++ b/src/utils.js
@@ -44,7 +44,7 @@ export function walk(el, callback) {
     }
 }
 
-export function debounce(func, wait, context = this) {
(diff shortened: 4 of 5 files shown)


HugoDF:
Seems like this and https://github.com/alpinejs/alpine/pull/275 do the same thing except #275 doesn't talk about debounce timeout in the title + has tests

ahmedkandel:
Wow, guys, you are so fast. Test added, Anyway I was working on the same thing #272 and had no time to PR

HugoDF:
Right but what's the difference between this PR and #275 ?

ahmedkandel:
The test is more precise as the mutations included are not limited to `array.push` but also has `x-data` property mutation which is the main purpose of the PR. Also passing the component as optional parameter `context` otherwise use the current context `this` will allow the default behavior of `debounce` function. What do you think?

HugoDF:
@ahmedkandel I'm not sure I understand what you're saying, #275 is a PR by @SimoTod (not your `array.length` double reactive update fix) I think this and that PR are applying the same fix but I might be wrong

ahmedkandel:
In #275 the test uses `@click="items.push('ok')"` only, and the mutations are `items[0]` and `items.length`, So if #271 gets merged the #275 test will pass anyway as it is not precise.

While in this PR the test uses `x-on:click="items.push('baz'); qux = 'corge';"` so the mutations will be in `items[0]` and `items.length` and also `qux` property. So the test is more precise and will pass only if the debounce timeout gets fixed.

Another point was passing the context as an optional parameter.

Maybe my English was not that good to make it more clear, But let me know if it needs more explanation.

HugoDF:
> In #275 the test uses `@click="items.push('ok')"` only, and the mutations are `items[0]` and `items.length`, So if #271 gets merged the #275 test will pass anyway as it is not precise.
> 
> While in this PR the test uses `x-on:click="items.push('baz'); qux = 'corge';"` so the mutations will be in `items[0]` and `items.length` and also `qux` property. So the test is more precise and will pass only if the debounce timeout gets fixed.
> 
> Another point was passing the context as an optional parameter.
> 
> Maybe my English was not that good to make it more clear, But let me know if it needs more explanation.

Nice, thanks for clarifying.

ahmedkandel:
As a final output both PRs "this and #275" will have almost the same result. But I think this is more finetuned and as I have stated before "I was working on it so didn't want to through it away" :relaxed:

SimoTod:
Sorry, I was out so I missed the whole conversation (i found it quite confusing as well, lol).
I wasn't aware you already found the bug because there wasn't anything last night, happy to close the PR so you get credits in the contributor page. 

About the other PR, we reallly don't need it. This debounce fix already covers it without the need of referring a specific property (length) by name which does not look neat or right (property did mutated so it's logically correct). You can keep the PR open but it will just add noise since Caleb is not around and when he's back he will have to review quite a lot of stuff. 

About this PR, be careful because you are assigning `this` as a default for your context but when not provided, it used to be undefined. The new implementation explicitly accesses context.debounce_timeout which will return an error if context is undefined.

ahmedkandel:
Thanks, @SimoTod .

Regarding #271 I think it is better to keep it open as advised by @HugoDF . The #271 fix was to revert unexpected behavior of observable-membrane.

So even if this PR will fix the component refresh issue, Still a good idea to revert the observable-membrane behavior.

SimoTod:
I don't agree with some of the feedback, but probably it's just me, lol. It's Caleb's call anyway. 

The behavior is unexpected for this specific use case (Alpine refresh) but it makes sense for the membrane. 2 properties are mutated so the handler is triggered twice as per specs (other people maybe want to observe the length property, we don't really know.(public lib shouldn't be tailored on the needs of one of their dependencies). 

The fix here covers all cases in a generic and nice way. 
271 only covers length but it won't be a problem if not merged so the PR is redundant and it just adds an unnecessary line to the code base. 
What if we had 3 properties caolusing a bug, were we going to add an if for each case?
I don't think we would. 

If debounce changes in the future, your new unit test will fail so it's enough to guarantee that we won't introduce regressions by mistake. 

Btw, no problem if you want to keep it open. Let's wait for Caleb to be back on the project so he can review and merge the fix. Thanks for working on that.

ahmedkandel:
A very good point,

> public lib shouldn't be tailored on the needs of one of their dependencies

For sure, but if you have a look at observable-membrane [Here ](https://github.com/salesforce/observable-membrane/blob/master/src/reactive-handler.ts#L80) they stated that considering array.length as a mutation to be handled is a fix for a specific use case inherited from an old multipackage [Here](https://github.com/salesforce/observable-membrane/commit/ed0dfa01be261be4648680489dbdf907b24fd0d6), So I think what #271 do is resetting to normal behavior.

BTW I have to thank @SimoTod and @HugoDF for the appreciate help.

ahmedkandel:
> you are assigning `this ` as a default for your context

I wanted to keep the original behavior of debounce function where `var context = this`. I expect that it will be always defined.

SimoTod:
It's never defined because the code uses arrow function which don't have this by default. Can you try please? I'm pretty sure it will error.

About the fix, i saw that line tge first time you mentioned it but I think they said exactly the opposite (if you read the following lines). They were expecting length to work normally but because it has already changed when they do the first check, they have to call it manually. The goal of the membrane is to call valueMutated (with the old and new value) whenever something changes. When you add a new element to an array you are changing 2 properties. Not that it matter that much. Just leave it, Caleb will review and decide. 👍

ahmedkandel:
@SimoTod correct :clap: `this` is never defined by default. Currently, it is not an issue no errors, But it is better to enforce `context` as a mandatory parameter for future use as `this` is useless here. I will commit this change.

Regarding #271 I don't agree with your interpretation of observable-membrane comment I still think it was for a specific use case that is unwanted in Alpine. Anyway I will leave it to Caleb to decide.

calebporzio:
Thanks for this fix. There were some really subtle context issues happening there. I still don't understand why the original implementation wasn't working. I got it from this blog post: https://davidwalsh.name/javascript-debounce-function

PonyJackal:
Does this work now?
I still  have the same issue

PR #285, opened 2020-03-17
Fix x-model .number modifier

### LINKED ISSUE #284: x-model .number modifier prevents typing negative numbers and fractions
Using `number` modifier without `lazy` modifier on `<input type="number">` will not allow users to type in negative values `-123` or fractions `.5` . While using it on `<input type="text">` will give a `NaN`.

The reason for this is that Alpine tries to cast the value of HTML input elements _" which always returns a string"_  to numeric using `parseFloat()`.

In HTML5 spec. [here](https://html.spec.whatwg.org/multipage/input.html#number-state-(type%3Dnumber)), `<input type="number">` will pass empty string to Alpine in `event.target.value` which makes it even worse.

One solution is to use `lazy` modifier to delay the model binding until input changes. But in some cases where live updates are needed still an issue. So I was working on a PR for `number` modifier to return:

- `null` if the input is empty or invalid in case of `<input type="number">`  start typing - or .
- original value "_string_" if casting to numeric fails
- numeric value if casting passes

This will fix the `NaN` issues and allow typing negative numbers and fractions in `<input type="number">.

ahmedkandel:
Closes #284

### DIFF
--- a/dist/alpine-ie11.js
+++ b/dist/alpine-ie11.js
@@ -5903,14 +5903,18 @@
         return modifiers.includes('number') ? Array.from(event.target.selectedOptions).map(function (option) {
           _newArrowCheck(this, _this3);
 
-          return parseFloat(option.value || option.text);
+          // If parseFloat() gives NaN return the original value
+          var number = option.value || option.text ? parseFloat(option.value || option.text) : null;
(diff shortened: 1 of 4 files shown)


calebporzio:
Perfect, thank you!

PR #289, opened 2020-03-18
Remove pointless Walk through the $nextTick

### LINKED ISSUE #287: $nextTick after component update or after render?
`$nextTick` is expected to execute a given expression AFTER Alpine has made its reactive DOM updates the same as Vue-nextTick. But looking at the source code there is some confusion "Maybe I am wrong".

Here in [component.js line 107](https://github.com/alpinejs/alpine/blob/master/src/component.js#L107) the expression should be excuted. But it will not, as `self.nextTickStack` will be empty already when reaching this point because it has been iterated in `initializeElements` or `updateElements`.

@HugoDF @SimoTod what do you think? should lines [107-109] be removed if they are useless?

ahmedkandel:
Closes #287

### DIFF
--- a/dist/alpine-ie11.js
+++ b/dist/alpine-ie11.js
@@ -6403,11 +6403,7 @@
             debounce(function () {
               _newArrowCheck(this, _this3);
 
-              self.updateElements(self.$el); // Walk through the $nextTick stack and clear it as we go.
-
-              while (self.nextTickStack.length > 0) {
-                self.nextTickStack.shift()();
-              }
+              self.updateElements(self.$el);
             }.bind(this), 0)();
           }
         });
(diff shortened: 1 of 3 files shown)


calebporzio:
Please provide a more thorough explanation. (I know you referenced a GH issue, but I'd still like more clarity on this contribution. Also, tests would be a great way to demonstrate its use, thanks!)

ahmedkandel:
Sure, in component.js we "walk through the $nextTick stack and clear it as we go" in three different places. Under `initializeElements()` and `updateElements()` which is expected, But also under `valueMutated() > debounce()` which i think it is useless and make some confusion.

The point is inside `debounce()` we call `updateElements()` which clears out the `nextTickStack` so after `updateElements()` finishes there will not be any remaining expressions in `nextTickStack`.

The current $nextTick test already passes after removing this part which means it still working as expected.

calebporzio:
Thanks! Good catch!

PR #291, opened 2020-03-19
Documentation: added descriptions to directive & magic props overview in README

christopher-siegel:
Since the README is the only Documentation at the moment, I've found it easier to get a quick overview about the directives and magic props when I added a short description.

When you're not used to Vue those directives are not always obvious. 

[See changes in rich diff](https://github.com/alpinejs/alpine/pull/291/commits/61dd7c8c01d27a7f984a95e32a415b0d456e3cff?short_path=04c6e90#diff-04c6e90faac2675aa89e2176d2eec7d8)

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,32 +86,32 @@ You can even use it for non-trivial things:
 
 There are 13 directives available to you:
 
-| Directive
-| --- |
-| [`x-data`](#x-data) |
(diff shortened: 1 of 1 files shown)


calebporzio:
Very nice work on this. Thanks!

PR #294, opened 2020-03-19
Fix $watch scope, Support nested properties, Support watch deep

ahmedkandel:
Replaces #292

### DIFF
--- a/README.md
+++ b/README.md
@@ -515,6 +515,22 @@ You can also use `$dispatch()` to trigger data updates for `x-model` bindings. F
 
 You can "watch" a component property with the `$watch` magic method. In the above example, when the button is clicked and `open` is changed, the provided callback will fire and `console.log` the new value.
 
+Use dot-delimited paths to watch properties of nested objects like `foo.bar.baz`. For example:
+
+```html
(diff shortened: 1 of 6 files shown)


ahmedkandel:
Here is a codepen to cover watch cases:
https://codepen.io/tomorrowsystem/pen/BaNPpXz?editors=1001

ahmedkandel:
It may need some refactoring, While it will not be easy to get rid of recursive iterations.

ahmedkandel:
What do you think guys? Is this updated implementation better?
@calebporzio @HugoDF @SimoTod

calebporzio:
I'm going to close this for now and only support shallow "watching" in Alpine V2. I believe this feature will just work out of the box with the new v3 reactivity engine.

I know this wasn't an easy implementation, so thank you for the contribution 🙏

justingolden21:
How many months or years has this been? I still can't watch an object? How am I supposed to react to changes to the object? DO I need to attach a listener to literally every single item in the UI that is bound to any part of the object? Doesn't seem productive for a JS framework. Makes many, many UIs impractical if not impossible to write.

ryangjchandler:
@justingolden21 You can use `Alpine.effect()` as mentioned in my comment above.

PR #299, opened 2020-03-23
Add simple core events

### LINKED ISSUE #283: Component lifecycle events

ryangjchandler:
This PR closes #283 and introduces **three** new events that are dispatched throughout the Alpine lifecycle.

1. `alpine:loaded` - fired at the document level once Alpine has discovered and initialised all components.

2. `alpine:transition-start` - fired at the transitioning element level once a transition has started but will bubble up to higher levels.

3. `alpine:transition-end` - fired at the transitioning element level after a transition has ended but will bubble up to higher levels.

Happy to make changes to this or introduce new events that people might be interested in. You might note that the original PR had some more events when data changes occur, but at the moment I feel the new `$watch` helper will suffice.

I'm happy to add some tests too, just didn't feel the need to do so at the moment.

### DIFF
--- a/dist/alpine-ie11.js
+++ b/dist/alpine-ie11.js
@@ -5315,6 +5315,10 @@
   function transition(el, stages) {
     var _this12 = this;
 
+    var eventContext = {
+      target: el
+    };
+    dispatch('alpine:transition-start', eventContext);
     stages.start();
     stages.during();
     requestAnimationFrame(function () {
@@ -5342,6 +5346,7 @@
             stages.cleanup();
           }
         }.bind(this), duration);
+        dispatch('alpine:transition-end', eventContext);
       }.bind(this));
     }.bind(this));
   } // I grabbed this from Turbolink's codebase.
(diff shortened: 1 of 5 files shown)


ryangjchandler:
As a further note, I just took the dispatch function for Livewire...

calebporzio:
Closing this for now. Added "Alpine events" to the Roadmap for V3 in the Readme

abinoda:
> You might note that the original PR had some more events when data changes occur, but at the moment I feel the new $watch helper will suffice.

When does $watch fire? Is it simply after component data is mutated, or is it after the component has been refreshed based on mutated data?

PR #315, opened 2020-03-30
Implement value binding on radio inputs

### LINKED ISSUE #308: binding radio input values
Hi there!

Is there a way to bind radio input values? I had figured this would be possible - and it works fine for checkboxes - but I can't get any kind of `x-bind:value="blah"` to work on radios. All the examples I can find online only have hardcoded values.

CodePen example: https://codepen.io/pcohen12/pen/ExjGmEG?editors=1000

pcohen12:
Closes #308 

Distinguishes between `x-bind:value` and `x-model` value bindings so radio value can be set from x-bind. Also forces `x-model` to be processed after `x-bind:value` for radios, so a radio input's value will be set before checking its value against any state values (allowing for a correct `checked` property).

p.s. Thank you for this awesome framework!

### DIFF
--- a/dist/alpine-ie11.js
+++ b/dist/alpine-ie11.js
@@ -3115,15 +3115,38 @@
     ArrayPrototype$1[UNSCOPABLES][key] = true;
   };
 
+  var $findIndex = arrayIteration.findIndex;
+
+
+
+  var FIND_INDEX = 'findIndex';
+  var SKIPS_HOLES = true;
+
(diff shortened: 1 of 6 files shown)


calebporzio:
Thanks! And very nice work on the comment line lengths 😉 👌

ryangjchandler:
@calebporzio Looks like the new test has some conflict <<<<<<... HEAD tags left in still, which I guess is why the suite is failing :)

pcohen12:
@calebporzio Hah, thanks 😛Happy to help!

PR #316, opened 2020-03-30
Fix nested x-for loop scoping

calebporzio:
This is now possible:
```html
<div x-data="{ foos: [{bars: [{bobs: ['one', 'two']}, {bobs: ['three', 'four']}]}, {bars: [{bobs: ['five', 'six']}, {bobs: ['seven', 'eight']}]}] }">
    <template x-for="foo in foos">
        <div>
            <template x-for="bar in foo.bars">
                <div>
                    <template x-for="bob in bar.bobs">
                        <span x-text="bob"></span>
                    </template>
                </div>
            </template>
        </div>
    </template>
</div>
```
(This outputs:)
image

### DIFF
--- a/dist/alpine-ie11.js
+++ b/dist/alpine-ie11.js
@@ -5349,7 +5349,7 @@
     return !isNaN(subject);
   }
 
-  function handleForDirective(component, el, expression, initialUpdate) {
+  function handleForDirective(component, el, expression, initialUpdate, extraVars) {
     var _this = this;
 
     if (el.tagName.toLowerCase() !== 'template') console.warn('Alpine: [x-for] directive should only be added to <template> tags.');
@@ -5369,7 +5369,7 @@
       // empty, effectively hiding it.
       items = [];
     } else {
-      items = component.evaluateReturnExpression(el, bunch);
+      items = component.evaluateReturnExpression(el, bunch, extraVars);
     } // As we walk the array, we'll also walk the DOM (updating/creating as we go).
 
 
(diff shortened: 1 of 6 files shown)


husayt:
How do you access `foo` or `bar` inside the last `x-for`?

husayt:
seems my last question is not supported yet  and waiting for [this to merge](https://github.com/alpinejs/alpine/pull/357)

timohoelzer:
Any updates on nesting.

Try to do something like the following which seams a practical case to me.
Creating a new **x-data** within **x-for** makes foo unreachable.

Am I missing something?

Many thanks in advance.

`
<div 
     x-data="{ 
             products:[ 
               { name:'Product 1', images:[1,2] }, 
               { name:'Product 2', images:[3,4] } 
             ]}">
  
  <template x-for="foo in products">
    
    <div x-data="Product(foo)">
        ...
    </div>
  </template>
</div>

<script>
  function Product(params) {
    return {
      ...params
    }
  }
</script>`

SimoTod:
That use case is not supported. Nested components cannot communicate with the parent components. 
The fact that a component is inside an x-for is not relevant, you would get a similar error with
```
<div x-data="{product: { name:'Product 1', images:[1,2] }]">
    <div x-data="product">
        ...
    </div>
</div>
```

As long as you define a new x-data attribute, that component creates its own standalone scope. You can have a look at [spruce](https://github.com/ryangjchandler/spruce) if you need your components to share a common state

PR #328, opened 2020-04-02
Fix x-if inside x-for

calebporzio:
The following snippet was broken before:
```html
<div x-data="{ foos: [{bar: 'baz'}, {bar: 'bop'}]}">
            <template x-for="foo in foos">
                <div>
                    <template x-if="foo.bar === 'baz'">
                        <div>
                            <span x-text="foo.bar"></span>
                        </div>
                    </template>
                </div>
            </template>
        </div>
```

### DIFF
--- a/dist/alpine-ie11.js
+++ b/dist/alpine-ie11.js
@@ -5708,7 +5708,7 @@
     component.showDirectiveLastElement = el;
   }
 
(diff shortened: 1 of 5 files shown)


WL1981:
Thank you! I've switched back to the most recent version.

PR #341, opened 2020-04-04
nextTick runs before x-for finishes updating

### LINKED ISSUE #317: x-ref won't focus an input
Hi so, I've a modal wich i  open abusing the dispatch events (so i can open from outide the component), and i want to focus an input so I've tried using `x-ref` on the input and then applied focus via ` $nextTick`  but it won't focus the input, contrary to when i do `$el.focus()` it'll focus the modal correctly. I've setup an JSBIN here: https://jsbin.com/wigupiloxa/edit?html,output. Any idea what can i do to focus the input?

### LINKED ISSUE #333: $nextTick doesnt work properly
Here's codepen i created to illustrate the issue:

https://codepen.io/piotrpog/pen/dyoLwJO

I created list of images and initialized slider script **swiper** on it. So far so good. Then I swap these images and need re-initialize slider so it is displayed correctly.

I attached updating slider to button in codepen example. I used `$nextTick`, so slider updates only after now images are inserted. But id does not seems to work. I need to click button two times in order to make it work.

SimoTod:
Fixes #333
When using x-for and updating the list, Alpine updates all the directives but ignores the x-for item and then it flushes the nextTick stack so the callbacks are currently running before x-for finishes its rendering. After that point, each element of the list calls either initializeElements or updateElements (each of them will try again to empty the nextTick stack).

~~To fix it, I've debounced the stack execution so it should run only at the end (either because the main refresh is the only call or it's the last call of a x-for loop).~~
~~I've duplicated the debounce function because it seemed it was conflicting but I might be wrong.~~

To fix it, will run the stack only if the root element of the update function is the Alpine component and we ignore element generated by x-for and other templating directives.

~~Fixes 317 as well since the nextTick debouncing allows the transition to progress further.~~ (Not true anymore since i've removed the debouncing for a better and neater solution, it needs to be fixed in a different PR)

### DIFF
--- a/src/component.js
+++ b/src/component.js
@@ -392,8 +392,17 @@ export default class Component {
     }
 
     executeAndClearNextTickStack() {
-        while (this.nextTickStack.length > 0) {
-            this.nextTickStack.shift()()
+        const self = this
+
+        const later = function () {
+            this.nextTickTimeout = null
+            while (self.nextTickStack.length > 0) {
+                self.nextTickStack.shift()()
+            }
         }
+
+        clearTimeout(this.nextTickTimeout)
+
+        this.nextTickTimeout = setTimeout(later, 10)
     }
 }
--- a/src/component.js
+++ b/src/component.js
@@ -36,6 +36,7 @@ export default class Component {
         this.unobservedData.$el = this.$el
         this.unobservedData.$refs = this.getRefsProxy()
 
+        this.nextTickTimeout = null
         this.nextTickStack = []
         this.unobservedData.$nextTick = (callback) => {
             this.nextTickStack.push(callback)
@@ -151,12 +152,7 @@ export default class Component {
 
         this.executeAndClearRemainingShowDirectiveStack()
 
-        // Walk through the $nextTick stack and clear it as we go.
-        debounce(function() {
-            while (this.nextTickStack.length > 0) {
-                this.nextTickStack.shift()()
-            }
-        }, 0).bind(this)()
+        this.executeAndClearNextTickStack()
     }
 
     initializeElement(el, extraVars) {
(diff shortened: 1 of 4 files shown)


ryangjchandler:
I assume this fix will also apply to any `$nextTick` callbacks inside of that `x-for` loop too?

SimoTod:
The other next tick calls seemed to behave already but yeah, if there are multiple calls in the same "update", only the last one should run. (well, I hope :)) 

Since debounced, it seems to partially fix x-show with transitions, too. I said partially because the next tick stack is still running before the end of the transition (but at least after the element is shown so we can apply the focus on its subelement).

ryangjchandler:
> The other next tick calls seemed to behave already but yeah, if there are multiple calls in the same "update", only the last one should run. (well, I hope :))
> 
> Since debounced, it seems to partially fix x-show with transitions, too. I said partially because the next tick stack is still running before the end of the transition (but at least after the element is shown so we can apply the focus on its subelement).

Yeah just tested that on a Codepen. Nice work mate :)

src/component.js
```diff
@@ -395,4 +390,19 @@ export default class Component {
...
+        const later = function () {
+            this.nextTickTimeout = null
+            while (self.nextTickStack.length > 0) {
+                self.nextTickStack.shift()()
+            }
+        }
+
+        clearTimeout(this.nextTickTimeout)
+
+        this.nextTickTimeout = setTimeout(later, 10)
```
calebporzio, line 406: I don't like adding an arbitrary buffer of 10ms. I'd much rather an implementation that is truly the "nextTick" meaning (which is usually achieved by a setTimeout of 0)

Does that make sense?

I'm grateful you found the bug and contributed a test, just hoping for a better implementation.

SimoTod: yeah, there was some sort of race condition between the 2 debounce but i'm not a big fan of the duplicated debounce. I'll see if i can fix it in a different way. Thanks for reviewing

SimoTod: Got a simpler solution which does not involve debouncing and adding a timeout.

calebporzio:
Very nice, I like this much better. Thanks @SimoTod

PR #350, opened 2020-04-07
Duplicate listeners when resolving x-if directives.

### LINKED ISSUE #348: @click fires twice within an x-if template
@click handlers on a button are firing twice each time the button is clicked if the button is inside an x-if template.

I wrote up a simple example to highlight this behavior:

https://codepen.io/atomgiant/pen/ExVYxxy

Thanks!

SimoTod:
Closes #348

When alpine runs the initial update calling initializeElements, it walks down the component children, resolving data and attaching elements.
When it finds a x-if, if the guard resolve to true, it attaches a new element after the template tag and calls another initializeElements on the new element. The second calls already binds data and attaches the required listeners. 
After that the first call continues and gets the next element which is now the one created by x-for but, since Alpine doesn't know, it duplicates any possible event listeners.

To fix it, similar to what we do for x-for, we should skip anything generated by x-if in initializeElements.
To allow the function to run correctly when called from the x-of handler, I've moved the key setter after the initializeElements call.

### DIFF
--- a/dist/alpine-ie11.js
+++ b/dist/alpine-ie11.js
@@ -23,10 +23,10 @@
   	return module = { exports: {} }, fn(module, module.exports), module.exports;
   }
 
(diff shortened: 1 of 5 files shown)


calebporzio:
Thanks @SimoTod!!

overtrue:
@calebporzio Hi, When will this fix be released? I have the same problem.

PR #354, opened 2020-04-08
Add support for looping over objects with `x-for`

### LINKED ISSUE #340: x-for "items.forEach is not a function" error when looping through an object
On **Alpine 2.2.5**, with this sample data:

```js
function app() {
    return {
        colours: {
            red: '#FF0000',
            green: '#00FF00',
            blue: '#0000FF',
        }
    }
}
```
and this sample markup:

```html
<div x-data="app()">
    <template x-for="(colour, index) in colours" :key="index">
        <div>
           <span x-text="index"></span>
           <span x-text="colour"></span>
        </div>
    </template>
</div>
```

I get this error.
```log
alpine.js:420 Uncaught (in promise) TypeError: items.forEach is not a function
    at handleForDirective (alpine.js:420)
    at alpine.js:1504
    at alpine.js:1377
    at alpine.js:1369
    at walk (alpine.js:85)
    at walk (alpine.js:89)
    at walk (alpine.js:89)
```

If I drop the object notation for a simple array, the for loop works. Like this:

```js
function app() {
    return {
        colours: [
            '#FF0000',
            '#00FF00',
            '#0000FF'
        ]
    }
}
```

Any ideas. Am I doing something wrong, overlooking something or does Alpine not like nested objects? Thanks!

ryangjchandler:
This pull request closes #340 and any others relating to using `x-for` with an object. This is supported with Vue, so it makes sense to support it here too.

I've kept the usage details similar to Vue, but the implementation details need some explaining (there's not many changes actually).

1. I first had to add support for looping over the object. This was actually simple enough using `Object.entries()` since you can pass in an array and it will work exactly the same as with an object. This did change the function signature slightly, as it will now receive an array instead of individual `index` and `i` variables.

2. Destructure the `entry` array to receive the `index` and `i` variables. In the case that the property in actually an array, then the `index` variable will hold the correct index for that item. 

3. This bit seems a bit strange at first, and I'm not sure if I should change the variable `group` to `index` since that's actually what it is at the start of the function but I didn't want to change too many variables and break something else. Small explanation for the if statement:

When looping over an array, you can use the following expression: 

```html
<template x-for="(value, key, collection) in items">
    ...
</template>
```

But when looping over an object, to keep the syntax inline with Vue's implementation you can use the following syntax:

```html
<template x-for="(value, key, index) in items">
    ...
</template>
```

Without assigning the original `items` variable to `group`, the `collection` variable inside of the loop would actually be the index of the array instead. We also want to make sure that the `index` is an integer, the same as it would be for `Array.forEach()`.

From there, it's exactly the same. There is probably some discussion needed on this too, so I'm not expecting a quick merge as there are currently work arounds like using `x-for="(value, key) in Object.entries(items)"` manually.


SimoTod:
It looks nice and neat 👍 . Well done.

I'm also confused about the current implementation returning the collection in the third variable: it's different from what Vue does. One for Caleb, I guess.

It would be good to update the documentation in the same PR.

ryangjchandler:
Yeah, good point on the documentation. I'll wait to see if Caleb wants any changes first and then update the docs.

It is a bit confusing, but I guess there's a reason behind it.

Thanks though :)

ryangjchandler:
After looking at the build output for IE11, I'm not sure we need the added `Object.entries()` polyfill since Babel should do the work for us. Wait for Caleb to confirm.

ryangjchandler:
I've also done some benchmarks and found that `Object.entries()` is quite a bit slower than `Object.keys()` and then `obj[key]` to get the value. I assume that's why Vue uses the latter. Might be worth looking into when we come around to implementing this in core.

calebporzio:
Going to hold off on this addition for now. Thanks Ryan.

PR #357, opened 2020-04-08
inherit extraVars in x-for scope

### LINKED ISSUE #356: Nested x-for: Cannot access outer loop variable from inside inner loop
```html
<h1>Alpine loops 1</h1>
<script src="https://cdn.jsdelivr.net/gh/alpinejs/alpine@v2.2.3/dist/alpine.js" defer></script>
<h2>Push Add First</h2>
<div x-data="{items: [  { name: 'a', sub_items: [ { name: 'c' }, { name: 'd' }] }, { name: 'b', sub_items: []}]}">
  <button x-on:click="items.unshift({ name: 'first', sub_items: []})">Insert first</button>
  <template x-for="(item, i) in items" :key="item.name">
    <div>
      <span x-text="item.name"></span>
      <div>
        <template x-for="(item2, j)  in item.sub_items">
          <div>
            <span x-text="item2.name"></span>
            <!-- error below trying to access "item" defined above -->
            (parent: <span x-text="item.name"></span>)
          </div>
        </template>
      </div>
    </div>
  </template>
</div>
```

This is the code created by `saferEval`:

```javascript
(function anonymous($data,item2,$dispatch
) {
var result; with($data) { result = item.name }; return result
})
```

There's no `item` argument there. Can the nested `x-for` accumulate the created variables?

Thanks!

rhengles:
Fixes #356

This has worked for me, but I don't know if it's the proper solution.

Includes tests.

Thanks

### DIFF
--- a/src/directives/for.js
+++ b/src/directives/for.js
@@ -10,7 +10,7 @@ export function handleForDirective(component, templateEl, expression, initialUpd
     // As we walk the array, we'll also walk the DOM (updating/creating as we go).
     let currentEl = templateEl
     items.forEach((item, index) => {
-        let iterationScopeVariables = getIterationScopeVariables(iteratorNames, item, index, items)
+        let iterationScopeVariables = getIterationScopeVariables(iteratorNames, item, index, items, extraVars())
         let currentKey = generateKeyForIteration(component, templateEl, index, iterationScopeVariables)
         let nextEl = currentEl.nextElementSibling
 
@@ -65,8 +65,9 @@ function parseForExpression(expression) {
     return res
 }
 
-function getIterationScopeVariables(iteratorNames, item, index, items) {
-    let scopeVariables = { [iteratorNames.item]: item }
+function getIterationScopeVariables(iteratorNames, item, index, items, scopeVariables) {
+    if (! scopeVariables) scopeVariables = {}
+    scopeVariables[iteratorNames.item] = item
     if (iteratorNames.index) scopeVariables[iteratorNames.index] = index
     if (iteratorNames.collection) scopeVariables[iteratorNames.collection] = items
 
--- a/src/directives/for.js
+++ b/src/directives/for.js
@@ -66,6 +66,7 @@ function parseForExpression(expression) {
 }
 
 function getIterationScopeVariables(iteratorNames, item, index, items, extraVars) {
+    // We must create a new object, so each iteration has a new scope
     let scopeVariables = extraVars ? {...extraVars} : {}
     scopeVariables[iteratorNames.item] = item
     if (iteratorNames.index) scopeVariables[iteratorNames.index] = index
--- a/src/directives/for.js
+++ b/src/directives/for.js
@@ -65,8 +65,8 @@ function parseForExpression(expression) {
     return res
 }
 
-function getIterationScopeVariables(iteratorNames, item, index, items, scopeVariables) {
-    if (! scopeVariables) scopeVariables = {}
+function getIterationScopeVariables(iteratorNames, item, index, items, extraVars) {
+    let scopeVariables = extraVars ? {...extraVars} : {}
     scopeVariables[iteratorNames.item] = item
     if (iteratorNames.index) scopeVariables[iteratorNames.index] = index
     if (iteratorNames.collection) scopeVariables[iteratorNames.collection] = items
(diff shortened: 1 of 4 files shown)


ryangjchandler:
Seems like the proper solution. Thanks for submitting it!

WL1981:
Hope it will be merged soon

rhengles:
@calebporzio Rebased on #362 👍

HugoDF:
@rhengles any idea why the tests are now failing?

rhengles:
@HugoDF It's something to do with the integration server, for some reason it couldn't run any of the tests.

From what I can see, it's not something related to the implementation. I'll try a force push and see if the tests run again...

HugoDF:
@rhengles cool!

WL1981:
There is a issue with the parent vars when I want to pass them to a x-on:click="Func(parentVar, subVar)" attribute / function. 

To demonstrate the issue i've made a [jsfiddle](https://jsfiddle.net/WFL1981/xnpgk19y/) (using latest commit  [f439c35](https://github.com/alpinejs/alpine/commit/f439c35b455d307833d99de66f63ed519f8bd423))

When I click on a checkbox the alert always shows the last subitem.name in the subitems object.
```javascript
<div x-data="GetData()">

  <template x-for="item in Object.values(items)">
    <div>

        <div x-text="item.title"></div>
                <input type="checkbox" x-on:click="toggle(item.title. 'test' )" />

        <template x-for="subitem in Object.values(item.data.subitems)">
          <div>

            <label>
            
              <input type="checkbox" x-on:click="toggle(item.title,  subitem.name )" />
              <span x-text="subitem.name"></span>
            </label>

          </div>
        </template>


    </div>
  </template>
</div>

```
I hope I have described the problem clearly enough. If not then I would like to hear it so that it can be solved as soon as possible.

src/directives/for.js
```diff
@@ -65,8 +65,9 @@ function parseForExpression(expression) {
     return res
 }
 
-function getIterationScopeVariables(iteratorNames, item, index, items) {
-    let scopeVariables = { [iteratorNames.item]: item }
+function getIterationScopeVariables(iteratorNames, item, index, items, scopeVariables) {
+    if (! scopeVariables) scopeVariables = {}
```
SimoTod, line 69:
```suggestion
function getIterationScopeVariables(iteratorNames, item, index, items, extraVars) {
    let scopeVariables = extraVars ? {...extraVars} : {}
```
It should fix the issue highlighted by @WL1981
Since the variable is an object, javascript passes the reference so on each iteration you are overriding the previous value. It doesn't happen at level 1 since the variable is null so you instantiate a brand new object each time.
I hope it makes sense (You need to recompile dist after applying the change).

WL1981: Great job! This change will fix [this](https://github.com/alpinejs/alpine/pull/357#issuecomment-614510539) issue. Also, it will pass all (122) tests.

Thank you @SimoTod

HugoDF: I've been thinking of maybe better names for `extraVars` would `scopeInheritedVariables` be accurate (instead of `extraVars`)?

ryangjchandler: > I've been thinking of maybe better names for `extraVars` would `scopeInheritedVariables` be accurate (instead of `extraVars`)?

By scope inherited you mean inherited from an outside scope I assume. That's not always the case though with every handler, for example the $dispatch function is just an extraVar, not necessarily a variable / piece of data from a higher level scope. I agree that it does need to be renamed though so clearly state *what* it is (even though `extraVars` does kind of do that).

rhengles: Thanks @WL1981 @SimoTod @HugoDF @ryangjchandler , good catch! Will add this fix.

SimoTod: Yep, I'm not precious about the suggestion. I just wanted to highlight that object are passed by reference in js so the code needs to create a fresh copy everytime. I didn't want to use the same name to avoid confusion but you can just keep 1 variable and reassign it at line 1. Up to the author.

rhengles:
@WL1981 I will also add this test to ensure this doesn't happen again.

src/directives/for.js
HugoDF: Agree with all this, ignore the renaming if it's going to add knowledge overhead

calebporzio:
Thanks @rhengles !

calebporzio:
Thanks for this!

Issue #342, opened 2020-04-04
Maybe highlight the word "classes" in transition readme


bnomei:
https://github.com/alpinejs/alpine#x-transition

otherwise it might be confusing for beginners that do not recognize the tailwind.css classes and think that alpine might magically apply props.

HugoDF:
@bnomei would you be up for submitting that change?

PR #358, opened 2020-04-09
explained that transition example uses tailwindcss classes

bnomei:
TODO: The japanese README should be updated as well.

closes: https://github.com/alpinejs/alpine/issues/342

### DIFF
--- a/README.md
+++ b/README.md
@@ -449,7 +449,7 @@ You can nest `x-for` loops, but you MUST wrap each loop in an element. For examp
 </template>
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -449,7 +449,7 @@ You can nest `x-for` loops, but you MUST wrap each loop in an element. For examp
 </template>
 ```
 
-Alpine offers 6 different transition directives for applying classes to various stages of an element's transition between "hidden" and "shown" states. These directives work both with `x-show` AND `x-if`.
+Alpine offers 6 different transition directives for applying classes to various stages of an element's transition between "hidden" and "shown" states. The example above uses classes from [tailwindcss](https://tailwindcss.com). These directives work both with `x-show` AND `x-if`.
```
ryangjchandler, line 452: Tailwind CSS*

PR #362, opened 2020-04-10
Refactor `x-for`

calebporzio:


WL1981:
There is an issue when using the extra var into an x-on or x-model.
Also when adding an index to the x-for (item, index) .... The returning index within x-on or x-model attribute always returns the last one in loop.

```javascript
<div x-data="{ items: ['foo', 'bar'], count: 1 }">
    <button @click="count++">increment</button>
    <template x-for="item in items" x-if="count < 2">

        // This one is working
        <span x-text="item"></span>

        // Returns: ReferenceError: item is not defined
        <input type="checkbox" x-on:click="toggle(item )" />

        // Maybe a new issue but also returning ReferenceError
        <input type="checkbox" x-model="test" :value="item" />

    </template>
</div>
```

SimoTod:
Template needs a unique root node to work in Alpine @WL1981

ryangjchandler:
> Template needs a unique root node to work in Alpine @WL1981 

Should probably make this a bit clearer in the docs. I'm not sure it's mentioned at all for `x-for`, only for `x-if`. I'll throw in a PR.

WL1981:
Can u give a small example?

ryangjchandler:
Sure, given the following code:

```html
<template x-for="(item, index) in items" :key="index">
    <span x-text="item.title"></span>
    <span x-text="item.anotherProperty"></span>
</template>
```

This will show a warning in the console, like:

"Alpine: <template> tag with [x-for] encountered with multiple element roots. Make sure <template> only has a single child node."

You might need to change your verbosity levels in the console to see it.

You'd instead need your loop to look this this:

```html
<template x-for="(item, index) in items" :key="index">
    <div>
        <span x-text="item.title"></span>
        <span x-text="item.anotherProperty"></span>
    </div>
</template>
```

Everything should be wrapped in a single root element, in this case the `<div>`.

SimoTod:
Just an heads up: `<input type="checkbox" x-model="test" :value="item" />` is potentially a source of headache.
You are looping on multiple items and binding the same variable to independent checkbox so they are likely to conflict.

PR #364, opened 2020-04-11
Prevent error when turbolinks reloads the page

### LINKED ISSUE #319: data-turbolinks-permanent: Uncaught ReferenceError on first navigation
This is tested with v2.2.3 -- but this isn't a new problem.

This is basically #309 -- but I think I have narrowed it down a little to make it easier to grasp.

See this commit: https://github.com/bep/hugo-alpine-test/commit/4860d5fbde7a25133e9df94f5d9ae3c3e8f02543#diff-32861fdd10fb962832bc30955219001bR31

Here I "load" the data in the `init()` function.

There is a running version at https://agitated-kilby-da786e.netlify.com/

If you:

* Navigate to https://agitated-kilby-da786e.netlify.com/ -- the dropdown is populated OK.
* Click on one of the links down on the page (powered by Turbolinks 5.2.0) you get the error below. The dropdown is still populated OK.
* If you continue to navigate there is no more error.

```bash
VM141:3 Uncaught ReferenceError: item is not defined
    at eval (eval at saferEval (alpine.js:109), <anonymous>:3:52)
    at saferEval (alpine.js:109)
    at Component.evaluateReturnExpression (alpine.js:1515)
    at handleAttributeBindingDirective (alpine.js:533)
    at alpine.js:1473
    at alpine.js:1377
    at alpine.js:1369
eval @ VM141:3
saferEval @ alpine.js:109
evaluateReturnExpression @ alpine.js:1515
handleAttributeBindingDirective @ alpine.js:533
(anonymous) @ alpine.js:1473
resolveBoundAttributes @ alpine.js:1459
initializeElement @ alpine.js:1396
(anonymous) @ alpine.js:1377
(anonymous) @ alpine.js:1369
walk @ alpine.js:85
walk @ alpine.js:89
walk @ alpine.js:89
walk @ alpine.js:89
walkAndSkipNestedComponents @ alpine.js:1357
initializeElements @ alpine.js:1374
Component @ alpine.js:1304
initializeComponent @ alpine.js:1660
(anonymous) @ alpine.js:1650
(anonymous) @ alpine.js:1630
discoverUninitializedComponents @ alpine.js:1629
(anonymous) @ alpine.js:1649
(anonymous) @ alpine.js:1643
childList (async)
r @ turbolinks.js:5
o.relocateCurrentBodyPermanentElements @ turbolinks.js:5
o.replaceBody @ turbolinks.js:5
(anonymous) @ turbolinks.js:5
e.renderView @ turbolinks.js:5
o.render @ turbolinks.js:5
e.render @ turbolinks.js:5
t.renderSnapshot @ turbolinks.js:5
t.render @ turbolinks.js:5
r.render @ turbolinks.js:6
(anonymous) @ turbolinks.js:5
(anonymous) @ turbolinks.js:5
requestAnimationFrame (async)
r.render @ turbolinks.js:5
r.loadResponse @ turbolinks.js:5
r.visitRequestCompleted @ turbolinks.js:5
r.requestCompletedWithResponse @ turbolinks.js:5
(anonymous) @ turbolinks.js:5
r.endRequest @ turbolinks.js:5
r.requestLoaded @ turbolinks.js:5
(anonymous) @ turbolinks.js:5
load (async)
r.createXHR @ turbolinks.js:5
r @ turbolinks.js:5
r.issueRequest @ turbolinks.js:5
r.visitStarted @ turbolinks.js:5
r.start @ turbolinks.js:5
r.startVisit @ turbolinks.js:6
r.startVisitToLocationWithAction @ turbolinks.js:5
r.visitProposedToLocationWithAction @ turbolinks.js:5
r.visit @ turbolinks.js:5
r.clickBubbled @ turbolinks.js:6
(anonymous) @ turbolinks.js:5
```

SimoTod:
Fixes #319

As described in the issue, alpine acts funny when turbolinks reloads the page if a component has the `data-turbolinks-permanent` attribute. 
I believe that the reason is that turbolinks load the new page and then replace the permanent element with the one from the previous page as described in their documentation.
This seems to trigger the mutation observer which tries to reevaluate all the directives.
If the component has a x-for directive, the injected html already has the generated x-for elements in it which refers variables that do not exists in the DOM.

The quickest solution is to pause the observer once turbolinks does its own stuff. Not sure if you can think of a better solution.

As I quickly mentioned in that issue, I believe that those 2 listeners don't belong to the core but they should be extracted into a separate bridge library.

### DIFF
--- a/dist/alpine-ie11.js
+++ b/dist/alpine-ie11.js
@@ -6579,6 +6579,13 @@
                   _newArrowCheck(this, _this);
 
                   this.initializeComponent(el);
+                }.bind(this)); // To avoid issue with the mutation observer going mad when turbolinks
+                // reload the page, we disconnect it
+
+                document.addEventListener("turbolinks:click", function () {
+                  _newArrowCheck(this, _this);
+
+                  this.observer.disconnect();
                 }.bind(this)); // It's easier and more performant to just support Turbolinks than listen
                 // to MutationObserver mutations at the document level.
 
@@ -6591,7 +6598,15 @@
                     _newArrowCheck(this, _this2);
 
                     this.initializeComponent(el);
-                  }.bind(this));
+                  }.bind(this)); // Once done, we reconnect the mutation observer
+
+                  var targetNode = document.querySelector('body');
+                  var observerOptions = {
+                    childList: true,
+                    attributes: true,
+                    subtree: true
+                  };
+                  this.observer.observe(targetNode, observerOptions);
                 }.bind(this));
                 this.listenForNewUninitializedComponentsAtRunTime(function (el) {
                   _newArrowCheck(this, _this);
@@ -6599,7 +6614,7 @@
                   this.initializeComponent(el);
                 }.bind(this));
 
-              case 6:
+              case 7:
               case "end":
                 return _context.stop();
             }
@@ -6647,7 +6662,7 @@
         attributes: true,
         subtree: true
       };
-      var observer = new MutationObserver(function (mutations) {
+      this.observer = new MutationObserver(function (mutations) {
         var _this6 = this;
 
         _newArrowCheck(this, _this5);
@@ -6673,7 +6688,7 @@
           }
         }
       }.bind(this));
-      observer.observe(targetNode, observerOptions);
+      this.observer.observe(targetNode, observerOptions);
     },
     initializeComponent: function initializeComponent(el) {
       if (!el.__x) {
--- a/dist/alpine-ie11.js
+++ b/dist/alpine-ie11.js
@@ -7193,8 +7193,8 @@
           attributes: true,
           subtree: true
         });
-      }.bind(this)); // We don't want tutbolinks to cache elements created by Alpine directives
-      // so we clean up the stage before letting turbolinks do its duty
+      }.bind(this)); // We don't want turbolinks to cache elements created by Alpine directives
+      // so we clean up the stage before letting turbolinks does its duty
 
       document.addEventListener("turbolinks:before-cache", function () {
         var _this9 = this;
@@ -7207,7 +7207,7 @@
 
           var nextEl = el.nextElementSibling;
 
-          while (nextEl && nextEl.__x_for_key) {
+          while (nextEl && typeof nextEl.__x_for_key !== 'undefined') {
             var currEl = nextEl;
             nextEl = nextEl.nextElementSibling;
             currEl.remove();
@@ -7219,7 +7219,7 @@
 
           var nextEl = el.nextElementSibling;
 
-          if (nextEl && nextEl.__x_inserted_me) {
+          if (nextEl && typeof nextEl.__x_inserted_me !== 'undefined') {
             nextEl.remove();
           }
         }.bind(this));
--- a/dist/alpine-ie11.js
+++ b/dist/alpine-ie11.js
@@ -7078,42 +7078,11 @@
                   _newArrowCheck(this, _this);
 
                   this.initializeComponent(el);
-                }.bind(this)); // To avoid issue with the mutation observer going mad when turbolinks
-                // reload the page, we disconnect it
-
-                document.addEventListener("turbolinks:click", function () {
-                  _newArrowCheck(this, _this);
-
-                  this.observer.disconnect();
-                }.bind(this)); // It's easier and more performant to just support Turbolinks than listen
-                // to MutationObserver mutations at the document level.
-
-                document.addEventListener("turbolinks:load", function () {
-                  var _this2 = this;
-
-                  _newArrowCheck(this, _this);
-
-                  this.discoverUninitializedComponents(function (el) {
-                    _newArrowCheck(this, _this2);
-
-                    this.initializeComponent(el);
-                  }.bind(this)); // Once done, we reconnect the mutation observer
-
-                  var targetNode = document.querySelector('body');
-                  var observerOptions = {
-                    childList: true,
-                    attributes: true,
-                    subtree: true
-                  };
-                  this.observer.observe(targetNode, observerOptions);
                 }.bind(this));
-                this.listenForNewUninitializedComponentsAtRunTime(function (el) {
-                  _newArrowCheck(this, _this);
+                this.listenForNewUninitializedComponentsAtRunTime();
+                this.initTurbolinksListeners();
 
-                  this.initializeComponent(el);
-                }.bind(this));
-
-              case 7:
+              case 6:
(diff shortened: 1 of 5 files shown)


ryangjchandler:
Agreed RE the bridge approach. I was actually looking at this earlier to see how difficult it would be and the answer is not at all. The only reason I can assume Caleb included it in core is because it's another package to maintain.

bep:
Personally I don't mind adding another bridge library or even manually do observer.Pause/Start etc., I will say that some of the beauty of this library is that it's small and can be just CDN-included. And I think you would be surprised how many people who would raise their hands saying they would want to use this in combo with Turbolinks, adding some navigational dynamic to their content rich site.

I think you're underestimating having Turbolinks as a first class citizen (just working, out of the box).

If you're worried about the extra event listeners, maybe make it optional?

ryangjchandler:
Yeah, I'm not partial to either option. There are pros and cons to each as you said. I'll leave it down to Caleb though.

calebporzio:
I'll weigh in more later, but here's the gist: I do like the idea of Alpine supporting Turbolinks out of the box.

SimoTod:
Since we want to support Turbolinks in the core I've updated this PR to cover #372 as well.
Unfortunately there's no way to test those issue with Jest but I've updated the example page so we can check them manually.

willcosgrove:
@SimoTod this fixed my issue!  I had just gotten a theory about what was wrong when I saw you had replied pointing me to this issue 😄 Thanks for the quick help! 🙏

bep:
@SimoTod the first part of this PR works great, but I'm not sure it's a great idea to add #372 to this mix as well -- mostly because it does not work. 

I took this for a spin now, and had a head-scratching moment looking at elements being removed after a Turbolinks-navigation:

My construct looked something like this:

```html
 <template x-if="!data.open">
 <div class="">Open</div>
</template>
<template x-if="data.open">
<div class="">Closed</div>
</template>
```

Removing those `template` does not get them back.

SimoTod:
@bep If you can checkout this PR and try the turbolinks-manual-test/index.html with the current release you can replicate the original issue.

SimoTod:
Closing for now since the solution is not fully working (and likely not optimal). It needs to be looked at by someone with more knowledge about Turbolinks so they can test all the nuisances (unfortunately. it's not possible to write automated tests for it).

SimoTod:
@bep Yeah, i think you are right and it's still not the right solution. 
I'll leave it with @calebporzio since my knowledge about Turbolinks is quite limited.

bep:
@SimoTod OK. I still think the fix to the initial problem was solid, though. As to my remark above, this "manual test" test case reproduces that for me:

```html
<h2>This component should be preserved and not be logging any errors when navigating to the Second page</h2>
        <div x-data="{ foo: ['bar', 'baz'] }" id="bar" data-turbolinks-permanent>
            <template x-for="item in foo">
                <span x-text="item"></span>
            </template>
        </div>
```

With this PR, there is now items shown on the second page.

SimoTod:
Yeah, it used to work but i believe it clashes with the other bugfix.
Probably turbolinks uses the same caching system so those bugs need to be looked at again. 
I'm going to have time to experiment again next weekend if nobody fixes it before (although I would be more confident if someone with more turbolinks knowledge picked it up). 
In the meantime, I've closed the PR to avoid it could get merged into master by mistake.

willcosgrove:
I have some ideas but I don’t have time to type them up right now 😒 it would probably be helpful to look and see how Stimulus handles it, since it plays very well with Turbolinks.

Stated simply and in hand wavy fashion: I think Alpine needs a way to rehydrate from DOM that’s been previously initialized.  I’m sure that’s easier said than done.

bep:
Adjusting the patch to stay away from the "permanent" elements makes the manual tests pass for me:

```html
document.addEventListener('turbolinks:before-cache', () => {
				const els = document.querySelectorAll(':not([data-turbolinks-permanent])');
				els.forEach((el) => {
					const xFors = el.querySelectorAll('template[x-for]');
					xFors.forEach((el) => {
						let nextEl = el.nextElementSibling;

						while (nextEl && typeof nextEl.__x_for_key !== 'undefined') {
							const currEl = nextEl;
							nextEl = nextEl.nextElementSibling;
							currEl.remove();
						}
					});
					const xIfs = el.querySelectorAll('template[x-if]');
					xIfs.forEach((el) => {
						let nextEl = el.nextElementSibling;

						if (nextEl && typeof nextEl.__x_inserted_me !== 'undefined') {
							nextEl.remove();
						}
					});
				});
			});
		}
```

bep:
@SimoTod I had a look at it, and come up with a variant of your patch that both passes the "manual tests" and the issues I had in my site.

In short:

1. Move the observer.disconnect() to before-cache (to make sure it's also triggered on back button).

2. Skip permanent elements when cleaning in before-render (see below). I will see if I can create a PR from this, as I really need a fix for this sooner rather than later.

```js
document.addEventListener('turbolinks:before-render', (e) => {
				walk(document.body, (el) => {
					if (el.hasAttribute('data-turbolinks-permanent')) {
						return false;
					}
					if (el.hasAttribute('x-data')) {
						const xFors = el.querySelectorAll('template[x-for]');
						xFors.forEach((el) => {
							let nextEl = el.nextElementSibling;

							while (nextEl && typeof nextEl.__x_for_key !== 'undefined') {
								const currEl = nextEl;
								nextEl = nextEl.nextElementSibling;
								currEl.remove();
							}
						});
						const xIfs = el.querySelectorAll('template[x-if]');
						xIfs.forEach((el) => {
							let nextEl = el.nextElementSibling;

							if (nextEl && typeof nextEl.__x_inserted_me !== 'undefined') {
								nextEl.remove();
							}
						});
					}
					return true;
				});
			});
```

SimoTod:
@bep 
No worries, go ahead with your PR. 👍
It seems a sensible approach but there's just a thing I'm not sure about. You use the walk function to clean up the DOM (ok) but then you use querySelectorAll which i believe could overrule your initial check if you have something like
```
<div x-data="...">
  <div data-turbolinks-permanent>
    <template x-for="...">
     ... 
    </template>
  </div>
</div>
```

I assume you can just keep walking the dom and you don't need to check for x-data. Something like
```
walk
  if permanent, skip
  if x-for, clean up nodes
  if x-if, clean up node
```

If I said a nonsense, ignore the comment. I haven't tried the code yet. 🤣

bep:
@SimoTod yes, you are correct -- I think my code is logically correct, but it makes more sense to just walk, walk...

PR #367, opened 2020-04-12
Fixes #366 cursor position jumping to the end

### LINKED ISSUE #366: x-model jumps cursor to end in Safari
To reproduce:

- Open [this flems](https://flems.io/#0=N4IgzgpgNhDGAuEAmIBcIB0ALeBbKIANCAGYCWMYaA2qAHYCGuEamO+RIsA9nYn6wA8YWACcyAB3gACMKNgBeADogc8CWFQB6LbCR0MAKzBJoZAG6iMdCPC0BzLFoZQJZG8eev3EAALmAJgwggFYtJDIwOxc3GwxcdyMwFQA+QS0RcSkUpTpcwSgGACNoFIAVAE8JCGksCFEIVHTCkqgcukEI82kADwBaJAZ4BmUQYGkkVGkAcmnpAF9U3OlpQXcJAFcZeCqIUcQe+BVevtxuUyhRpCW6FeEJBlv+g-gr1PSwB7p29K72zkgMAQZF4VHQIVQAQAbCB5oR6EwWOgkpweHwIAJ0LD4SBGMxWBhYGAqMQ0fx4KxYQBdeZAA) in desktop or mobile safari
- Type some text into the text box
- Move your cursor to the middle of your text
- Type or delete a character

The cursor jumps to the end of the input.

muzafferdede:
Thanks to @SimoTod  for his suggestion to apply the fix in the right file.   Fixes #366

### DIFF
--- a/src/directives/bind.js
+++ b/src/directives/bind.js
@@ -42,8 +42,10 @@ export function handleAttributeBindingDirective(component, el, attrName, express
             el.value = value
 
             //Set the origin cursor position after updating the value
-            el.selectionStart = cursorPosition
-            el.selectionEnd = cursorPosition
+            if (el === document.activeElement) {
+                el.selectionStart = cursorPosition
+                el.selectionEnd = cursorPosition
+            }
         }
     } else if (attrName === 'class') {
         if (Array.isArray(value)) {
--- a/src/directives/bind.js
+++ b/src/directives/bind.js
@@ -39,7 +39,9 @@ export function handleAttributeBindingDirective(component, el, attrName, express
             // Cursor position should be restored back to origin due to a safari bug
             const cursorPosition = el.selectionStart 
             el.value = value
-            el === document.activeElement ? el.setSelectionRange(cursorPosition, cursorPosition) : null
+            if(el === document.activeElement) { 
+                el.setSelectionRange(cursorPosition, cursorPosition)
+            }
         }
     } else if (attrName === 'class') {
         if (Array.isArray(value)) {
--- a/src/directives/bind.js
+++ b/src/directives/bind.js
@@ -36,16 +36,10 @@ export function handleAttributeBindingDirective(component, el, attrName, express
         } else if (el.tagName === 'SELECT') {
             updateSelect(el, value)
         } else {
-            //Get the origin cursor position before updating the value
-            const cursorPosition = el.selectionStart
-
+            // Cursor position should be restored back to origin due to a safari bug
+            const cursorPosition = el.selectionStart 
             el.value = value
-
-            //Set the origin cursor position after updating the value
-            if (el === document.activeElement) {
-                el.selectionStart = cursorPosition
-                el.selectionEnd = cursorPosition
-            }
+            el === document.activeElement ? el.setSelectionRange(cursorPosition, cursorPosition) : null
         }
     } else if (attrName === 'class') {
         if (Array.isArray(value)) {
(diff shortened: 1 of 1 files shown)


ryangjchandler:
@muzafferdede looks like the tests are failing, could you take a look when you get a second please :)

muzafferdede:
```js
test('x-model casts value to number if number modifier is present', async () => {
    document.body.innerHTML = `
        <div x-data="{ foo: null }">
            <input type="number" x-model.number="foo"></input>
        </div>
    `

    Alpine.start()

    fireEvent.input(document.querySelector('input'), { target: { value: '123' }})

    await wait(() => { expect(document.querySelector('[x-data]').__x.$data.foo).toEqual(123) })
})
```
Is it because "text" and "number" inputs uses same test?

SimoTod:
I think you need to set those properties back only if they are defined. Changes can come from the linked property being updated by something else.
In that case, your element doesn't have a cursor so when you read selectionStart and selectionEnd they return something like null or undefined.
If you try to set them back, you are going to get the error because a cursor position has to be a number.

HugoDF:
> I think you need to set those properties back only if they are defined. Changes can come from the linked property being updated by something else.

Maybe "only if they are different"?

SimoTod:
> Maybe "only if they are different"?

I would say defined, they are often going to be different in safari due to the apple bug but it doesn't mean that the value received is valid, I think.

It seems to happen only in the test framework due to a partial implementation of jsdom any way.

However, in safari, it results in the input acquiring the focus when it didn't have it so it needs some additional fixes. (See https://codepen.io/SimoTod/pen/oNjXbmy)

SimoTod:
@muzafferdede You might want to check if your current element matches `document.activeElement` before setting those values back.

It means that you won't change those properties if the input field doesn't have the focus.

That should prevent the other bug in the codepen and it should also fix the tests.

muzafferdede:
Yes, @SimoTod i did check if current element is active element, then apply the cursor positions. Tests are green now.

ryangjchandler:
Can we not just use `el.setSelectionRange(start, end)` instead of the individual setters?

SimoTod:
I think it's the same, right?

ryangjchandler:
> I think it's the same, right?

Yeah, just thinking one less line. No worries, just me being pedantic

muzafferdede:
I was thinking that as well but if you notice, cursorPosition is selectionStart and  applied for both selectionStart  and selectionEnd. I see some weird behave in safari if i store both and apply both. Since i was thinking this is a safari bug so maybe because of that it's behave weird. So i use start position only. We could still do something like this for sake of better code. 
```js
const cursorPosition = {
   start: el.selectionStart,
   end: el.selectionEnd,
}

el.value = value

el === document.activeElement ?: restoreCursor(el,cursorPosition)
...
function restoreCursor(el,cursorPosition) {
   el.selectionStart = cursorPosition.start
   el.selectionEnd = cursorPosition.end
}
```

ryangjchandler:
I think what you done is fine. I was only suggesting consolidating it into one single line instead of the 2 setters.

src/directives/bind.js
```diff
@@ -36,7 +36,10 @@ export function handleAttributeBindingDirective(component, el, attrName, express
         } else if (el.tagName === 'SELECT') {
             updateSelect(el, value)
         } else {
+            // Cursor position should be restored back to origin due to a safari bug
+            const cursorPosition = el.selectionStart 
             el.value = value
+            el === document.activeElement ? el.setSelectionRange(cursorPosition, cursorPosition) : null
```
ryangjchandler, line 42: I'm not sure a ternary should be used here since the else value is null. I'd instead use a regular if statement.

muzafferdede: Yeah, that's the right path.

calebporzio:
This looks great. Thanks!

PR #373, opened 2020-04-14
fix spelling error in comments

AdetunjiTejumade:


AdetunjiTejumade:
@HugoDF 
@ryangjchandler

src/component.js
```diff
@@ -55,7 +55,7 @@ export default class Component {
         // If x-init is present AND we aren't cloning (skip x-init on clone)
         if (initExpression && ! seedDataForCloning) {
             // We want to allow data manipulation, but not trigger DOM updates just yet.
-            // We haven't even initialized the elements with their Alpine bindings. I mean c'mon.
+            // We haven't even initialized the elements with their Alpine bindings. I mean come on.
```
SimoTod, line 58: This is more a deliberate slang than a spelling error. There should be a `from` spelled `form` somewhere else in this file.

PR #375, opened 2020-04-14
Bug/x for regression

### LINKED ISSUE #371: Regression in x-for
This error came after 9e91c4704638ee5c3472eb8a19d54fcd0b5caa59

I don't have a codepen for this, but the component that fails has 3 nested x-for loops.

```bash
alpine.js:418 Uncaught TypeError: Cannot convert undefined or null to object
    at alpine.js:418
    at Proxy.forEach (<anonymous>)
    at handleForDirective (alpine.js:404)
    at alpine.js:1495
    at Component.resolveBoundAttributes (alpine.js:1450)
    at Component.updateElement (alpine.js:1426)
    at alpine.js:1394
    at alpine.js:1358
    at walk (alpine.js:85)
```

SimoTod:
Fixes #371

The x-for refactoring introduced a regression. When updating a keyed list, if new items are not inserted at the end, the lookahead function doesn't return anything and the code doesn't handle it correctly.

### DIFF
--- a/dist/alpine-ie11.js
+++ b/dist/alpine-ie11.js
@@ -5816,7 +5816,7 @@
           return nextEl.__x_for;
         }.bind(this));
       } else {
-        nextEl = lookAheadForMatchingKeyedElementAndMoveItIfFound(nextEl, currentKey); // If I haven't found a matching key, just insert the element at the current position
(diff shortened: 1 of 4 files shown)


calebporzio:
Thanks @SimoTod!!!

PR #377, opened 2020-04-14
Fix bound classes being removed unnecessarily

### LINKED ISSUE #374: x-bind:class is removing classes
See https://codepen.io/bepsays/pen/xxwwzVP

Ref. the documentation

>For classes, you pass in an object who's keys are class names, and values are boolean expressions to determine if those class names are applied or not.

So, given this:

```html
<div x-data="{ open: true, level: 32 }">
  <style>
    .red {
      background-color: red;
    }
    .blue {
      background-color: blue;
    }
    
    .text-yellow {
      color: yellow;
    }
    
    .text-blue {
      color: green;
    }
    
    
  </style>
  <button x-bind:class="{ 
                          'red text-yellow' : open,
                          'red text-blue' : !open && level === 12,
                        }">Button</button>
 
</div>
```

I would expect a red button with yellow text. But the `red` class is removed by the second statement.

I now notice that if I switch the conditions above, I get the colors I want, but that is for this very simple case.

Is there a reason why it's implemented like this? Is this how Vue does it?

ryangjchandler:
Closes #374. This fix isn't quite what was being suggested in the issue. The actual problem was that the order of operations (adding or removing the classes) was unknown, so the removal of a class might come after the addition of it and vice versa.

I've chosen to keep the footprint of the method the same near enough and instead sort the keys by the value before looping over them. This way, the classes that need to be removed (value evaluates to false) will become the iteratee first, and then new classes will be added afterwards.

If anyone thinks the method suggested in the PR is worth doing instead, let me know and I'll close this for someone else to implement or update the PR accordingly. Both have the same outcome either way and tests pass still.

### DIFF
--- a/dist/alpine-ie11.js
+++ b/dist/alpine-ie11.js
@@ -5547,17 +5547,18 @@
         el.value = value;
       }
     } else if (attrName === 'class') {
-      var originalClasses = el.__x_original_classes || [];
-
       if (Array.isArray(value)) {
+        var originalClasses = el.__x_original_classes || [];
         el.setAttribute('class', arrayUnique(originalClasses.concat(value)).join(' '));
       } else if (_typeof(value) === 'object') {
-        var keysSortedByBoolean = Object.keys(value).sort(function (a, b) {
+        // Sorting the keys / class names by their boolean value will ensure that
+        // anything that evaluates to `false` and needs to remove classes is run first.
+        var keysSortedByBooleanValue = Object.keys(value).sort(function (a, b) {
           _newArrowCheck(this, _this);
 
           return value[a] - value[b];
         }.bind(this));
-        keysSortedByBoolean.forEach(function (classNames) {
+        keysSortedByBooleanValue.forEach(function (classNames) {
           var _this2 = this;
 
           _newArrowCheck(this, _this);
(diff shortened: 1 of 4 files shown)


ryangjchandler:
The test suite seems to be a bit unstable at the moment. `x-on` tests are failing, despite none of the code for that part changing and all tests passing locally.

ryangjchandler:
> The test suite seems to be a bit unstable at the moment. `x-on` tests are failing, despite none of the code for that part changing and all tests passing locally.

Ignore this. The tests seem to pass on the last commit, there's obviously some sort of inconsistency with the time it takes for the test suite to insert the DOM, Alpine to start and assertions to be made between each run.

SimoTod:
It occurs to me locally at times. The transition test fails, then I run all tests again and it passes. 🤷‍♂️

ryangjchandler:
> It occurs to me locally at times. The transition test fails, then I run all tests again and it passes. 🤷‍♂️

Yeah, I just ran the test suite 10 times and it failed twice, passed the rest. Oh well 🤦🏻‍♂️

calebporzio:
Thanks Ryan!

Issue #378, opened 2020-04-14
How to prevent inner elements from triggering mouse events?


mikebronner:
Consider the following: https://jsfiddle.net/cshoypu8/2/

What ends up happening is that the HTML elements within the photoRight element call the drag function. How can this be avoided? And why are they responding in the first place -- isn't the event handler only added to the photoRight element?

muzafferdede:
you can use ` pointer-events: none` css way to avoid any pointer events. If you are using `tailwindcss` you can apply `pointer-events-none` class to those elements.

mikebronner:
That's not a full solution in this case. I need to listen to different events in on the inner elements unrelated to this functionality. I'm asking more for clarity as to why things aren't behaving the way I would expect, rather than finding a work-around. Thanks though @muzafferdede :)

ryangjchandler:
This is the nature of JavaScript events. See this codepen: https://codepen.io/ryangjchandler/pen/ZEbbZmG

Clicking the button inside of the `<div>` will cause the handler to be triggered (check the console). 

When the button is clicked, an event is fired from the button but it bubbles up through the DOM and the handler attached to the div will react. 

To stop the button from propagating upwards, you can add `@click.stop` to it. This calls `event.stopPropagation()` which will prevent bubbling. 

Let me know if you have any more problems!

muzafferdede:
since he needs to use mousedown event, `x-on:mousedown.stop='methodname()` is the right way?

ryangjchandler:
> since he needs to use mousedown event, `x-on:mousedown.stop='methodname()` is the right way?

Exactly that. If one of the nested elements needs it's own `mousedown` event listener, but `mousedown.stop` should work fine. 

I'm on mobile so formatting might be a bit off, sorry!

mikebronner:
I had tried that. It seems the problem is the opposite in that the `x-on:mousemove` event is being registered on that element, but also on its child-nodes. This is what is confusing. I understand bubbling up, but this almost seems like bubbling down.

ryangjchandler:
> I had tried that. It seems the problem is the opposite in that the `x-on:mousemove` event is being registered on that element, but also on its child-nodes. This is what is confusing. I understand bubbling up, but this almost seems like bubbling down.

Interesting. I've got a PR ready to submit tomorrow that adds a new `.self` modifier for event listeners. It will only run the handler if the event target matches the element it was registered for (same as Vue's implementation). I think that might solve your problem. 

[...]

PR #379, opened 2020-04-14
Added new `.self` modifier for event listeners

ryangjchandler:
This PR introduces a new `.self` modifier with the same behaviour as Vue's implementation. This should also make it easier to prevent event's bubbled from nested elements being handled by parent elements, as in the case of #378.

When used on an element, i.e. `<div @click.self="handle()">`, the `handle()` method will only be called (or expression evaluated) if the element triggering the event matches the element that registered it. This will remove the need to add several `.stop` modifiers to nested elements when you don't want any event bubbling for that particular parent / component.

I'm not sure if it's worth noting in the docs / README that chaining modifiers in different orders, for example `@click.self.prevent` and `@click.prevent.self`, won't have any effect on the order of execution / evaluation since they are not being handled sequentially. The `.prevent` and `.stop` modifiers are always handled first, then the `.self` modifier.

P.s. there's a new test for this feature.

### DIFF
--- a/README.md
+++ b/README.md
@@ -289,6 +289,11 @@ Adding `.prevent` to an event listener will call `preventDefault` on the trigger
 
 Adding `.stop` to an event listener will call `stopPropagation` on the triggered event. In the above example, this means the "click" event won't bubble from the button to the outer `<div>`. Or in other words, when a user clicks the button, `foo` won't be set to `'bar'`.
 
+**`.self` modifier**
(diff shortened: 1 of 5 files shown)


ryangjchandler:
I realise now that the test I added doesn't test whether the handler runs when clicking the '.self' element... bozo moment 🤦‍♂️

calebporzio:
Thanks @ryan

bszyman:
Seconding that this is a really great addition. Thank you! :)

PR #382, opened 2020-04-15
Updated README to include note about single root element

ryangjchandler:
This was missing from the docs, just a simple addition. Not sure it closes any issues but is worth noting.


calebporzio:
Thanks!

PR #384, opened 2020-04-15
Refactor html/text directives into separate files

muzafferdede:
Not sure if `text.js` was intentionally left empty. I was thinking a filter idea so just want to move it to its own file for sake of readability. 

 >update
updated this PR to include `html.js` file as well. It will be better for future updates and maintaince. Now every directive has its own file.

### DIFF
--- a/dist/alpine-ie11.js
+++ b/dist/alpine-ie11.js
@@ -4206,43 +4206,6 @@
     return regexpExec.call(R, S);
   };
 
-  // @@match logic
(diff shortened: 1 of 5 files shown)


ryangjchandler:
Good shout :)

ryangjchandler:
@muzafferdede Would you be able to build the files and commit those too please?

muzafferdede:
is this failing because of that "common test failing" issue? my local is green.

SimoTod:
Probably if you force-push the branch again, it should pass.

calebporzio:
Thanks!

PR #385, opened 2020-04-16
fix spellings and grammar error in component.js 

AdetunjiTejumade:


src/component.js
```diff
@@ -69,7 +69,7 @@ export default class Component {
         this.listenForNewElementsToInitialize()
 
         if (typeof initReturnedCallback === 'function') {
-            // Run the callback returned form the "x-init" hook to allow the user to do stuff after
+            // Run the callback returned from the "x-init" hook to allow the user to do stuff after
```
ryangjchandler, line 72: This is the only error I was looking for actually, I don't think we need the other changes.

src/component.js
```diff
@@ -363,12 +363,12 @@ export default class Component {
         var refObj = {}
 
         /* IE11-ONLY:START */
-            // Add any properties up-front that might be necessary for the Proxy polyfill.
+            // Add any property up-front that might be necessary for the Proxy polyfill.
```
ryangjchandler, line 366: The plural (properties) is correct here.

AdetunjiTejumade:
i will redo thanks

ryangjchandler:
> i will redo thanks

No need to re-do and open another PR. Just reset your branch based on master, make the changes and force push :)

PR #395, opened 2020-04-19
x-ignore directive

### LINKED ISSUE #393: Add an no-alpine attribute
I mentioned it in #237 -- but I think it deserves its own issue.

A typical Alpine use case may look like this:

```html
<body>

<nav x-data="{ }">
</nav>

<div class="content" no-alpine>
... lots of potentially untrusted content.
</div>
</body>
```

I have jumped to the conclusion above, but the `no-alpine` would tell Alpine to skip walking this DOM node.

It should give 

* some level of security protection if you receive content from potentially untrusted sources
* some speedup not having to walk megabytes of blog HTML

And it should be easy to implement, I think.

muzafferdede:
>Updated description!

Fixes #393 New directive to ~~secure Alpine for untrusted DOM nodes~~ stop Alpine walk in the defined node. It still does not cover any security of the untrusted content. Just a way to stop Alpine initialize components within the node. 

Usage:

```html
   <div x-ignore="ignore" x-data="{ignore: true}">
        <div x-data="{foo:'foo'}">
            <span x-text="foo"></span>
        </div>
    </div>
```
it will ignore any `x-data` components within the applied node. There is a `x-ignore` test and all green. Docs updated as well.
> if `ignore` set to `false`, it will initialize  components within

### DIFF
--- a/README.md
+++ b/README.md
@@ -479,7 +479,7 @@ These behave exactly like VueJs's transition directives, except they have differ
 ### `x-ignore`
 **Example:**
 ```html
-<div x-ignore>
+<div x-ignore="true">
     <div x-data="{foo:'UNTRUSTED CONTENT'}">
         <span x-text="foo">"foo" never gets bind by alpine</span>
     </div>
--- a/README.md
+++ b/README.md
@@ -481,7 +481,7 @@ These behave exactly like VueJs's transition directives, except they have differ
 ```html
 <div x-ignore="true">
     <div x-data="{foo:'bar'}">
-        <span x-text="foo">"foo" will not set to 'bar' by Alpine</span>
+        <span x-text="foo">foo won't be set to 'bar' by Alpine</span>
     </div>
 </div>
 ```
--- a/README.md
+++ b/README.md
@@ -481,11 +481,11 @@ These behave exactly like VueJs's transition directives, except they have differ
 ```html
 <div x-ignore="true">
     <div x-data="{foo:'bar'}">
-        <span x-text="foo">"foo" never gets bind by alpine</span>
+        <span x-text="foo">"foo" will not set to 'bar' by Alpine</span>
     </div>
 </div>
 ```
-`x-ignore` can be used to stop DOM walking and ignoring any components within the element.
+`x-ignore` can be used to mark an element's children as "to be ignored" by Alpine. This means that Alpine will not walk the DOM tree any further and will not initialise any components within the `x-ignore`-ed element.
 
 ---
 
--- a/README.md
+++ b/README.md
@@ -480,7 +480,7 @@ These behave exactly like VueJs's transition directives, except they have differ
 **Example:**
 ```html
 <div x-ignore="true">
-    <div x-data="{foo:'UNTRUSTED CONTENT'}">
+    <div x-data="{foo:'bar'}">
         <span x-text="foo">"foo" never gets bind by alpine</span>
     </div>
 </div>
--- a/README.md
+++ b/README.md
@@ -104,6 +104,7 @@ There are 13 directives available to you:
 | [`x-if`](#x-if) | Remove an element completely from the DOM. Needs to be used on a `<template>` tag. |
 | [`x-for`](#x-for) | Create new DOM nodes for each item in an array. Needs to be used on a `<template>` tag. |
 | [`x-transition`](#x-transition) | Directives for applying classes to various stages of an element's transition |
+| [`x-ignore`](#x-ignore) | Ignores components within the the element. |
 | [`x-cloak`](#x-cloak) | This attribute is removed when Alpine initializes. Useful for hiding pre-initialized DOM. |
 
 And 6 magic properties:
@@ -472,6 +473,19 @@ These behave exactly like VueJs's transition directives, except they have differ
 | `:leave` | Applied during the entire leaving phase. |
 | `:leave-start` | Added immediately when a leaving transition is triggered, removed after one frame. |
 | `:leave-end` | Added one frame after a leaving transition is triggered (at the same time `leave-start` is removed), removed when the transition/animation finishes.
+
+---
+
+### `x-ignore`
+**Example:**
+```html
+<div x-ignore>
+    <div x-data="{foo:'UNTRUSTED CONTENT'}">
+        <span x-text="foo">"foo" never gets bind by alpine</span>
+    </div>
+</div>
+```
+`x-ignore` can be used to stop DOM walking and ignoring any components within the element.
 
 ---
 
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ There are 13 directives available to you:
 | [`x-if`](#x-if) | Remove an element completely from the DOM. Needs to be used on a `<template>` tag. |
 | [`x-for`](#x-for) | Create new DOM nodes for each item in an array. Needs to be used on a `<template>` tag. |
 | [`x-transition`](#x-transition) | Directives for applying classes to various stages of an element's transition |
-| [`x-ignore`](#x-ignore) | Ignores components within the the element. |
+| [`x-ignore`](#x-ignore) | Ignores components within the element. |
 | [`x-cloak`](#x-cloak) | This attribute is removed when Alpine initializes. Useful for hiding pre-initialized DOM. |
 
 And 6 magic properties:
--- a/test/ignore.spec.js
+++ b/test/ignore.spec.js
@@ -7,7 +7,7 @@ global.MutationObserver = class {
 
 test('x-ignore ignores initialization of any new components within the node', async () => {
     document.body.innerHTML = `
-    <div x-ignore>
+    <div x-ignore="true">
         <div x-data="{foo:'foo'}">
             <span x-text="foo"></span>
         </div>
--- a/test/ignore.spec.js
+++ b/test/ignore.spec.js
@@ -19,7 +19,7 @@ test('x-ignore ignores initialization of any new components within the node', as
     await wait(() => { expect(document.querySelector('span').innerText).toEqual(undefined) })
 })
 
-test('x-ignore if set to false it will initialize any new components within the node', async () => {
+test('if x-ignore set to false it will initialize any new components within the node', async () => {
     document.body.innerHTML = `
     <div x-ignore="false">
         <div x-data="{foo:'foo'}">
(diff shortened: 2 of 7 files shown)


ryangjchandler:
I don't think `x-secure` is the correct name for the attribute, it suggests that it is 100% security related. I think a smart name would be something like `x-noparse` or `x-noop`, since that's really what it does. Would love to here others thoughts, especially @bep since he opened the original issue.

muzafferdede:
> I don't think `x-secure` is the correct name for the attribute

@ryangjchandler Thanks for the feedback. :) Well i was thinking this issue too. `x-secure` sounds like taking all the security responsibilities of the node. But yeah, definitely deserves an argument on the name.

In case we need to change name, an alternative from my side would be `x-isolate`.

SimoTod:
It's probably worth to hold the fire a bit on this and wait for Caleb since it's not a bug but a feature request. We don't know if he is willing to add it to Alpine (specially V2) so you might end up with spending hours on a feature that won't be accepted in the core.

My 2 pence: It's more about parsing then security because security is trickier then you would think.
You don't trust the content of that section because you don't have full control so it would be better to parse it and sanitize it server-side rather then adding a directive which claim to secure your section but can be easily removed by untrusted content which would lead lazy developers to skip the backend part. See https://codepen.io/SimoTod/pen/KKdMbVX for an example.

For that reason I would avoid any reference about security in the name and also I agree with Ryan: x-secure and x-isolate suggest me that it's some sort of sandbox environment where Alpine is still running but in isolation.

I'm not suggesting you to change the name now for the same reason I mentioned in the first paragraph. Caleb can give input on this point as well if he decides to include it in the core.

muzafferdede:
@SimoTod Thanks for your feedback. 

Yeah, probably because i had the question in my mind as `security issue` when i read first and based on that doing some implementation, i decide to calling it `x-secure`.  :) 

I think we all agreed that name should not be called anything but what exactly what it does. Which we can decide once we confirm adding into core. I just don't want us to spend unnecessary discussion on the `name` and losing the real point which is **its use case**.

Just because of i spend hours on a PR doesn't means it has to be accepted of course. I enjoyed it already discovering the Alpine more and more, tho probably i just spent 30 mins, hehe :) So it already paid it cost. 

BTW changing the name would be a quick fix since core function is already there. So other than the `name`, any other thoughts on use case or implementation?

SimoTod:
It does what it says.

HugoDF:
Thoughts about `x-ignore`?

muzafferdede:
> Thoughts about `x-ignore`?

Sounds good to me.

ryangjchandler:
> Thoughts about `x-ignore`?

`x-ignore` works well too.

HugoDF:
This is a compact solution, good stopgap I think potentially the V3 default might be to _not_ automatically interpret injected components

test/ignore.spec.js
```diff
@@ -0,0 +1,20 @@
+import Alpine from 'alpinejs'
+import { wait } from '@testing-library/dom'
+
+global.MutationObserver = class {
+    observe() {}
+}
+
+test('x-ignore ignores initialization of any new components within the node', async () => {
+    document.body.innerHTML = `
+    <div x-ignore>
```
HugoDF, line 10: What happens if x-ignore is on the same element as x-data?

muzafferdede:
> This is a compact solution, good stopgap.

Agreed, actually the solution is not for future core. it's just for the time being in case of need. At least fix for the #393 without breaking changes.

test/ignore.spec.js
muzafferdede: To watch root ignore element by alpine,  it will be initialized as component. Because we might use it as :
```html
        <div x-bind:x-ignore="ignore" x-data="{ignore: true}">
            <div x-data="{foo:'foo'}">
                <span x-text="foo"></span>
            </div>
        </div>
``` 

so if ignore condition is true then it can be ignored. This is the logic for now. I hope it makes sense.

muzafferdede: Actually i think i need to adjust this part. Let me see.

muzafferdede: should be handled by ab545c15b0cf855ea93b29c9232d10d707a492a3

bep:
I like `x-ignore`. Also, I think it would be good to tone down the "security aspect" of this in the documentation -- it's easy for people to read "oh, so this will make everything 100% secure" when they see something like this.

muzafferdede:
@bep  Thanks for the feedback.

>  Also, I think it would be good to tone down the "security aspect" of this in the documentation 

Where exactly is mentioning about security aspect? Not sure if you check the latest changes on the PR. I think i revised all descriptions and removed all security related comments. Just want to make sure  i didn't miss somewhere.

bep:
@muzafferdede it looks great now (my security remark was more related to the old name "x-secure").

bep:
@SimoTod 
 your code snippet would not work if you also define a CSP. A "compromise CSP" for Alpine would look like this:

```
Content-Security-Policy = "script-src 'self' 'unsafe-eval'"
```

You could still argue that JavaScript "src includes" could do what you mention, but you certainly have better control and can test/verify that those don't do any harm.

But web security is ... hard.

SimoTod:
Agreed, that was just a very quick, and probably unrealistic, example of something that would break if a developer just relies on that feature. 

As you said, security is not easy and developers need to spend a bit of time to iron out their sites. That was the reason why I wanted to remove the word secure (which I believe everyone agreed with) so we don't give the false impression that Alpine does all the work for you.

I would still advise data to be sanitised on the backend when possible, even with this feature as a safety net (for example if you use node, https://github.com/cure53/DOMPurify/blob/master/README.md is a good option to remove all script tags and it can be configured to remove any x-data attributes in our case).

bep:
>I would still advise data to be sanitised on the backend when possible, 

Sure, but this isn't always possible (e.g. doing search with `fetch`). I still think this PR has great value considering how simple it is.

test/ignore.spec.js
```diff
@@ -0,0 +1,20 @@
+import Alpine from 'alpinejs'
+import { wait } from '@testing-library/dom'
+
+global.MutationObserver = class {
+    observe() {}
+}
+
+test('x-ignore ignores initialization of any new components within the node', async () => {
+    document.body.innerHTML = `
+    <div x-ignore="true">
```
HugoDF, line 10: Could we extend this/add new tests to include cases for

- `<div x-ignore>` (I'm assuming it will work the same as `x-ignore="true"` unless I'm missing something
- `<div x-ignore="false">` -> the component should actually initialise

Like I mentioned in an earlier comment, what happens when we've got `<div x-data="{}" x-ignore>`? I reckon in that case, the component shouldn't initialise, but whatever we decide there should be a test to validate the behaviour.

README.md
```diff
@@ -475,6 +476,19 @@ These behave exactly like VueJs's transition directives, except they have differ
 
 ---
 
+### `x-ignore`
+**Example:**
+```html
+<div x-ignore="true">
+    <div x-data="{foo:'bar'}">
+        <span x-text="foo">"foo" never gets bind by alpine</span>
```
HugoDF, line 484: `bind` should probably be `bound`

Also `alpine` -> `Alpine` (this one is a total nitpick, feel free to ignore)

README.md
```diff
@@ -475,6 +476,19 @@ These behave exactly like VueJs's transition directives, except they have differ
...
+    </div>
+</div>
+```
+`x-ignore` can be used to stop DOM walking and ignoring any components within the element.
```
HugoDF, line 488: `x-ignore` can be used to mark an element's children as "to be ignored" by Alpine. This means that Alpine will not walk the DOM tree any further and will not initialise any components within the `x-ignore`-ed element.

Depending on whether we implement the `<div x-data x-ignore>` behaviour we might have to amend this: `mark an element's children` to `mark an element`

test/ignore.spec.js
muzafferdede: - Yes, when i was implementing, i consider this and allow `x-ignore` without value as defaults to `x-ignore="true"`
- Yes `x-ignore="false"` will allow initialize the components within the applied node. 
- I have already included tests for each use cases. Let me know if all good.

README.md
muzafferdede: - Done

README.md
muzafferdede: -  Done. Thanks for the better description.

README.md
```diff
@@ -475,6 +476,19 @@ These behave exactly like VueJs's transition directives, except they have differ
 
 ---
 
+### `x-ignore`
+**Example:**
+```html
+<div x-ignore="true">
+    <div x-data="{foo:'bar'}">
+        <span x-text="foo">"foo" will not set to 'bar' by Alpine</span>
```
HugoDF, line 484: 😂 I'm starting to feel bad about the reviews

```suggestion
        <span x-text="foo">"foo" will not be set to 'bar' by Alpine</span>
```

test/ignore.spec.js
```diff
@@ -0,0 +1,48 @@
...
+import { wait } from '@testing-library/dom'
+
+global.MutationObserver = class {
+    observe() {}
+}
+
+test('x-ignore ignores initialization of any new components within the node', async () => {
+    document.body.innerHTML = `
+    <div x-ignore="ignore" data="{ignore:true}">
+        <div x-data="{foo:'foo'}">
```
HugoDF, line 11: Will it still work if we do the following?

```suggestion
    <div x-ignore="ignore" x-data="{foo:'foo'}">
        <div>
```

README.md
muzafferdede: Yeah, there might be a lot of these. :)

test/ignore.spec.js
muzafferdede: Yes, i actually tried bunch of ways but i think real world usages are more important. I have included similar one to this in the last update.

calebporzio:
My thoughts:

1. The issue mentions `x-ignore` saving resources, and this is only true for `x-ignore` used within an Alpine component if it's used outside of one, that DOM will never get walked anyways
1. The closest convention in Vue is `v-pre`
1. I'd be open to an `x-pre` tag that skips walking a subsection of a components DOM in the future but I don't see an immediate need

I'm going to close this for now.

If there is a significant amount of people experiencing pain from the lack of a feature like this we should reconsider.

Thanks for all the thoughts and work on this.

PR #403, opened 2020-04-21
Fixes #401. Preserve text selection on applicable inputs

### LINKED ISSUE #401: Regression in handleAttributeBindingDirective: el.setSelectionRange is applied naively
Please bear with me as this is the first issue I've ever opened...

Unfortunately the fix from PR #367 has introduced a bug when binding a value to a subset of input types.

According to the [MDN page on setSelectionRange()](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/setSelectionRange):
>Note that accordingly to the WHATWG forms spec selectionStart, selectionEnd properties and **setSelectionRange method apply only to inputs of types text, search, URL, tel and password**. Chrome, starting from version 33, throws an exception while accessing those properties and method on the rest of input types. For example, on input of type number: "Failed to read the 'selectionStart' property from 'HTMLInputElement': The input element's type ('number') does not support selection".

I was playing with some code on Firefox, updated to v2.3.0, and then started getting their version of the above Chrome error:
>  InvalidStateError: An attempt was made to use an object that is not, or is no longer, usable

which is just ridiculously cryptic and took me a while to realize the problem!

https://github.com/alpinejs/alpine/blob/49de05c3278556cac5477a4ca668e9daf280d703/src/directives/bind.js#L45-L51

At the very least L49 should check that the type is not any of those types before attempting to set the selection.

Something like 
```javascript
(el === document.activeElement && ['text', 'search', 'url', 'password'].includes(el.type)) 
```
should fix it, but I don't know if the developers would prefer to hoist this type check to the parent if/else block and only declare `const cursorPosition` for those types, and not also for everything else. Or at least move that const declaration inside the L49 `if` to optimize memory a bit.

Thoughts?

03juan:
Fixes #401 

This fix preserves any already-selected text range on the input value bound with x-model.

According to [the HTML spec](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#dom-textarea/input-selectionstart), only input elements whose text can be selected will return non-null selectionStart, selectionEnd, and selectionDirection values, so non-applicable input elements such as 'button' are skipped.

The test catches the `@select` event, fired on the element when its selection changes, to simulate an external change to the model's value.

### DIFF
--- a/test/bind.spec.js
+++ b/test/bind.spec.js
@@ -372,6 +372,33 @@ test('classes are removed before being added', async () => {
     await wait(() => {
         expect(document.querySelector('span').classList.contains('block')).toBeFalsy()
         expect(document.querySelector('span').classList.contains('hidden')).toBeTruthy()
-        expect(document.querySelector('span').classList.contains('text-red')).toBeTruthy
+        expect(document.querySelector('span').classList.contains('text-red')).toBeTruthy()
+    })
+});
+
+test('cursor position is preserved on selectable text input', async () => {
+    document.body.innerHTML = `
+        <div x-data="{ foo: 'bar' }">
+            <input type="text" 
+                   x-model="foo"
+                   x-ref="textInput" 
+                   @select="foo = 'baz'"
+            >
+        </div>
+    `
+
+    Alpine.start()
+
+    document.querySelector('input').setSelectionRange(0, 3, 'backward')
+    expect(document.querySelector('input').value).toEqual('bar')
+    expect(document.querySelector('input').selectionStart).toEqual(0)
+    expect(document.querySelector('input').selectionEnd).toEqual(3)
+    expect(document.querySelector('input').selectionDirection).toEqual('backward')
+
+    await wait(() => {
+        expect(document.querySelector('input').value).toEqual('baz')
+        expect(document.querySelector('input').selectionStart).toEqual(0)
+        expect(document.querySelector('input').selectionEnd).toEqual(3)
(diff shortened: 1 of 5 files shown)


03juan:
> 
> 
> **This Passes:**
> 

Thanks friend. I seems to have fixed it by first focus()'ing the element and then the positions are preserved. 👍

HugoDF:
Does this also fix https://github.com/alpinejs/alpine/issues/404?

03juan:
It should now fix #404

test/bind.spec.js
```diff
@@ -372,6 +372,50 @@ test('classes are removed before being added', async () => {
...
+
+    await wait(() => {
+        expect(document.querySelector('input').value).toEqual('baz')
+        expect(document.querySelector('input').selectionStart).toEqual(0)
+        expect(document.querySelector('input').selectionEnd).toEqual(3)
+        expect(document.querySelector('input').selectionDirection).toEqual('backward')
+    })
+})
+
+// input elements that are not 'text', 'search', 'url', 'password' types
```
HugoDF, line 405: Maybe we should leave a link to the GitHub issuss?

ryangjchandler: Yeah definitely.

ryangjchandler: Yeah definitely.

HugoDF: Issue numbers are fine

calebporzio:
Thanks for the fix!

PR #418, opened 2020-04-24
Fix x-on in x-for when updating

### LINKED ISSUE #413: Events stop firing when I "change a page"
This is not easy to describe so I've created a small jsfiddle with the required code to reproduce the problem:

https://jsfiddle.net/yn4h37pf/

What happens is that I want to implement pagination in a list of items and have each of these items respond to a click event. When you click on the numbers you'll see the event fire in the console log. However if you press "Next page" to load the next page you'll see that although it displays the next 3 numbers, when you click on their buttons nothing will happen. Even if you press previous page to return to the 1st page you still won't get any output in the console.

Am I doing something wrong or maybe this is something not supported at all or this is a bug ?

TIA !

SimoTod:
Fixes #413 

When we update a x-for, if there is a new item we are wrongly calling update instead of insert. More details in the issue.

### DIFF
--- a/dist/alpine-ie11.js
+++ b/dist/alpine-ie11.js
@@ -5800,9 +5800,8 @@
       _newArrowCheck(this, _this);
 
       var iterationScopeVariables = getIterationScopeVariables(iteratorNames, item, index, items, extraVars());
-      var currentKey = generateKeyForIteration(component, templateEl, index, iterationScopeVariables); // Look ahead for the right element to update
-
-      var nextEl = lookAheadForMatchingKeyedElementAndMoveItIfFound(currentEl.nextElementSibling, currentKey); // If we haven't found a matching key, just insert the element at the current position
+      var currentKey = generateKeyForIteration(component, templateEl, index, iterationScopeVariables);
+      var nextEl = lookAheadForMatchingKeyedElementAndMoveItIfFound(currentEl.nextElementSibling, currentKey); // If we haven't found a matching key, insert the element at the current position.
 
       if (!nextEl) {
         nextEl = addElementInLoopAfterCurrentEl(templateEl, currentEl); // And transition it in if it's not the first page load.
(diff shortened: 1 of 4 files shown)


dist/alpine-ie11.js
```diff
@@ -5911,7 +5906,9 @@
   }
 
   function lookAheadForMatchingKeyedElementAndMoveItIfFound(nextEl, currentKey) {
-    // If the the key's DO match, no need to look ahead.
+    // We don't know if nextEl is avtually an element so we check that it's not a falsy value first
```
muzafferdede, line 5909: // We don't know if nextEl is ~~avtually~~ actually an element so we check that it's not a falsy value first

dist/alpine.js
```diff
@@ -486,7 +481,9 @@
   }
 
   function lookAheadForMatchingKeyedElementAndMoveItIfFound(nextEl, currentKey) {
-    // If the the key's DO match, no need to look ahead.
+    // We don't know if nextEl is avtually an element so we check that it's not a falsy value first
```
muzafferdede, line 484: // We don't know if nextEl is ~~avtually~~ actually an element so we check that it's not a falsy value first

src/directives/for.js
```diff
@@ -114,6 +109,9 @@ function addElementInLoopAfterCurrentEl(templateEl, currentEl) {
 }
 
 function lookAheadForMatchingKeyedElementAndMoveItIfFound(nextEl, currentKey) {
+    // We don't know if nextEl is avtually an element so we check that it's not a falsy value first
```
muzafferdede, line 112: // We don't know if nextEl is ~~avtually~~ actually an element so we check that it's not a falsy value first

SimoTod: right, I can't spell words correctly, lol. Thanks.

muzafferdede:
@SimoTod Nice one! I am afraid say "Oh Yeah!". Because this `x-for` has been buggy since forever. I strongly believe it will be  re-factored again.  Probably what is happing with V3 already.

test/for.spec.js
```diff
@@ -448,3 +448,62 @@ test('nested x-for event listeners', async () => {
         expect(document._alerts[2]).toEqual('foo: bob = 2')
     })
 })
+
+
+test('listener are bound correctyl on update', async () => {
```
HugoDF, line 453: typos
```suggestion
test('listeners are bound correctly on update', async () => {
```

SimoTod: Thanks

calebporzio:
Thanks @SimoTod for the fix, that was a tricky one - great work!

Issue #407, opened 2020-04-22
Model binding tests for all input types


muzafferdede:
As lesson learned from #401 i think we should include all input types in tests for model bindings. 
```html
<input type="button">
<input type="checkbox">
<input type="color">
<input type="date">
<input type="datetime-local">
<input type="email">
<input type="file">
<input type="hidden">
<input type="image">
<input type="month">
<input type="number">
<input type="password">
<input type="radio">
<input type="range">
<input type="reset">
<input type="search">
<input type="submit">
<input type="tel">
<input type="text">
<input type="time">
<input type="url">
<input type="week">
```
It will improve future feature tests. Not sure what tests would do the job other than simple `foo` - `bar` bindings. Suggestions on binding scenarios?

ryangjchandler:
As I mentioned in #401, I actually tihnk the general test coverage for input elements and others is pretty good. There are a few inconsistencies here and there, but these make more sense to tackle as they come up or become a problem, for example Safari and the selection range stuff.

muzafferdede:
> @ryangjchandler  
Whether or not it's sneaky, all edge cases should be tested really. At least we know now going forwards.

> @ryangjchandler  
We (me, @HugoDF, @SimoTod) do try and comment on PRs to ensure they're fully tested, this one just slipped past a little since there are inconsistencies between input types / elements. Definitely a good thing to remember going forward, research the solution too :)

After reading your commends there,  I was thinking it might help to prevent `just slipped past a little` :) You are free the close the issue if you don't see the necessary of it.

HugoDF:
> It will improve future feature tests. Not sure what tests would do the job other than simple `foo` - `bar` bindings. Suggestions on binding scenarios?

There's probably a point of diminishing returns we need to be weary of. At the end of the day currently, all tests are run inside of JSDom which isn't a 1-1 with browsers.

If we really want to be on the safe side, E2E tests using something like Cypress (except it doesn't have IE11 or Safari support) would be the most useful to catch browser quirks before releasing code.

At the end of the day, this project has limited resource and I don't know if anyone is keen on spending that time writing tests.

> After reading your comments there, I was thinking it might help to prevent just slipped past a little :) You are free the close the issue if you don't see the necessary of it.

Testing won't be 100% regardless of what approach we take that's the nature of it, projects much bigger than this one have inadvertently introduced breaking changes in a patch version in the past.

Unfortunately, for the foreseeable future, time to remediation will be bound to Caleb's time/attention.

[...]

Issue #404, opened 2020-04-22
2.3.0 regression with x-bind / x-model and hidden inputs


gilbert:
- [2.2.5 x-model](https://flems.io/#0=N4IgzgpgNhDGAuEAmIBcIB0ALeBbKIANCAGYCWMYaA2qAHYCGuEamO+RIsA9nYn6wA8YWACcyAB3gACMKNgBeADogc8CWFQB6LbCR0MAKzBJoZAG6iMdCPC0BzLFoZQJZG8eev3EAALmAJgwggFYtJDIwOxc3GwxcdyMwFQA+QS0RcSkUpTpcwQjzaQAPAFokBngGZRBgaSRUaQByYqbpAF9U3OlpQXcJAFcZeABPCQgarDIkUzyQaUZmGtHxlWlp5bGINfMXAYmVNbKAI3cG3ah9mqQuuh7hCQY7ssRi+GvU9LBHuhy6dMKf04kBgCDIvCo6AADKgAGxQ0ow2EAZhA7UI9CYLHQSU4PD4EAE6DRGJAi2xmFgYCoxHx-HgrDRAF12kA) – Works
- [2.3.0 x-model](https://flems.io/#0=N4IgzgpgNhDGAuEAmIBcIB0ALeBbKIANCAGYCWMYaA2qAHYCGuEamO+RIsA9nYn6wA8YWACcyAB3gACMKNgBeADogc8CWFQB6LbCR0MAKzBJoZAG6iMdCPC0BzLFoZQJZG8eev3EAALmAJgwAZgwABi0kMjA7FzcbDFx3IzAVAD5BLRFxKTSlOnzBKPNpAA8AWiQGeAZlEGBpJFRpAHJSlukAX3T86WlBdwkAVxl4AE8JCDqsMiRTApBpRmY68cmVaVnViYgN8xchqZUNioAjdyb9qEO6pB66PuEJBgeKxFL4W-TMsGe6PLomWKAM4kBgCDIvCo6DCqAAbMEQJ1CPQmCx0ClODw+BABOgkSiQMt0ZhYGAqMRsfx4KwkQBdTpAA) – Throws [TypeError: Type error]

Same respective behaviors when you replace `x-model` with `x-bind:value`

sjelfull:
I _think_ I'm seeing the same thing. After upgrading, i'm seeing problems in parts of my app that uses hidden inputs.

SimoTod:
It should be the same regression discussed here #401

[...]

PR #424, opened 2020-04-26
Using Vue way to fix cursor bug on Safari

### LINKED ISSUE #423: Error thrown when attempting to bind value to input type="hidden" on Safari 13.0.5
Hello. Thanks a lot for Alpine! It feels great to use, striking the right balance between Stimulus and Vue.

I recently ran into an error when binding `value` (with `:value`) to a hidden input. A TypeError is thrown in Safari (tried it with 13.0.5). No error is thrown in Chrome or Firefox.

Changing the type of the input from `hidden` to `text` seems to stop the error.

Codepen to reproduce: https://codepen.io/agorf/pen/zYvwNGb

The error I get in the console on Codepen:

```
[Error] Unhandled Promise Rejection: TypeError: Type error
	handleAttributeBindingDirective (alpine.js:561)
	(anonymous function) (alpine.js:1512)
	forEach
	resolveBoundAttributes (alpine.js:1498)
	initializeElement (alpine.js:1417)
	(anonymous function) (alpine.js:1401)
	walk (alpine.js:77)
	walk (alpine.js:81)
	walkAndSkipNestedComponents (alpine.js:1379)
	initializeElements (alpine.js:1396)
	Component (alpine.js:1326)
	initializeComponent (alpine.js:1695)
	(anonymous function) (alpine.js:1643)
	(anonymous function) (alpine.js:1659)
	forEach
	discoverComponents (alpine.js:1658)
	(anonymous function) (alpine.js:1642)
	asyncFunctionResume
	(anonymous function)
	promiseReactionJob
```

I'd be glad to provide any other information.

muzafferdede:
Fixes #423
Updates #366 #367 #401 #404

I found that  Vue uses [an easy & better way](https://github.com/vuejs/vue/pull/3077/commits/13d01d078200fad765fcf09895b548377b7133a6).  it fixes the `cursor issue` by only **updating** the `el.value` if `old value` updated. (*Probably **value binding** happens multiple times, so `Safari` resets the `cursor positions` for each time*. )

My only concern if  `el.value`  check  would ever **throw an error** in other type of inputs. Would be nice if you guys can comment on this part.

I have tested this and it works. Not sure what type of tests would be suitable or do we need test at all for cursor positions? IMO we do not need **old cursor position tests**. But we can still consider to have simple test for all input types. (#407).  

I hope you guys can try it and double check if any issue with this approach.

cc @SimoTod @HugoDF @ryangjchandler @calebporzio

### DIFF
--- a/test/bind.spec.js
+++ b/test/bind.spec.js
@@ -376,36 +376,7 @@ test('classes are removed before being added', async () => {
     })
 });
 
-test('cursor position is preserved on selectable text input', async () => {
-    document.body.innerHTML = `
-        <div x-data="{ foo: 'bar' }">
-            <input type="text" x-model="foo" @select="foo = 'baz'">
-        </div>
-    `
-
-    Alpine.start()
-
-    document.querySelector('input').focus()
-
-    expect(document.querySelector('input').value).toEqual('bar')
-    expect(document.querySelector('input').selectionStart).toEqual(0)
-    expect(document.querySelector('input').selectionEnd).toEqual(0)
-    expect(document.querySelector('input').selectionDirection).toEqual('none')
-
-    document.querySelector('input').setSelectionRange(0, 3, 'backward')
-
-    await wait(() => {
-        expect(document.querySelector('input').value).toEqual('baz')
-        expect(document.querySelector('input').selectionStart).toEqual(0)
-        expect(document.querySelector('input').selectionEnd).toEqual(3)
-        expect(document.querySelector('input').selectionDirection).toEqual('backward')
-    })
-})
-
-// input elements that are not 'text', 'search', 'url', 'password' types
-// will throw an exception when calling their setSelectionRange() method
-// see issues #401 #404 #405
-test('setSelectionRange is not called for inapplicable input types', async () => {
+test('value bindings to hidden inputs', async () => {
     document.body.innerHTML = `
         <div x-data="{ foo: 'bar' }">
             <input type="hidden" x-model="foo">
--- a/test/bind.spec.js
+++ b/test/bind.spec.js
@@ -1,5 +1,5 @@
 import Alpine from 'alpinejs'
-import { fireEvent, wait } from '@testing-library/dom'
+import { wait } from '@testing-library/dom'
 
 global.MutationObserver = class {
     observe() {}
@@ -375,22 +375,6 @@ test('classes are removed before being added', async () => {
         expect(document.querySelector('span').classList.contains('text-red')).toBeTruthy()
     })
 });
-
-test('value bindings to hidden inputs', async () => {
-    document.body.innerHTML = `
-        <div x-data="{ foo: 'bar' }">
-            <input type="hidden" x-model="foo">
-        </div>
-    `
-
-    Alpine.start()
-
-    fireEvent.input(document.querySelector('input'), { target: { value: 'baz' } })
-
-    await wait(() => {
-        expect(document.querySelector('input').value).toEqual('baz')
-    })
-})
 
 test('extra whitespace in class binding object syntax is ignored', async () => {
     document.body.innerHTML = `
--- a/dist/alpine-ie11.js
+++ b/dist/alpine-ie11.js
@@ -5994,14 +5994,8 @@
       } else if (el.tagName === 'SELECT') {
         updateSelect(el, value);
       } else {
-        // Cursor position should be restored back to origin due to a safari bug
-        var selectionStart = el.selectionStart;
-        var selectionEnd = el.selectionEnd;
-        var selectionDirection = el.selectionDirection;
-        el.value = value;
-
-        if (el === document.activeElement && selectionStart !== null) {
-          el.setSelectionRange(selectionStart, selectionEnd, selectionDirection);
+        if (el.value !== value) {
+          el.value = value;
         }
       }
     } else if (attrName === 'class') {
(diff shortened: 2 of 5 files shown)


test/bind.spec.js
```diff
@@ -376,36 +376,7 @@ test('classes are removed before being added', async () => {
     })
 });
 
-test('cursor position is preserved on selectable text input', async () => {
```
HugoDF, line 379: So does this behaviour still work? Just the unit test fails?

ryangjchandler:
I guess it does make sense to only update the value if it has changed...

test/bind.spec.js
muzafferdede: I believe that test had a bug. (also the implementation didn't make sense to me). I tried my way of test and it did past but since there was no any way of setting cursor position by code, i removed that test. Open to another way of testing cursor position or discussing the previous way to me get it clear.

muzafferdede: which test fails? 
```js
cursor position is preserved on selectable text input

    expect(received).toEqual(expected) // deep equality

    Expected: 3
    Received: 0

      398 |         expect(document.querySelector('input').value).toEqual('baz')
      399 |         expect(document.querySelector('input').selectionStart).toEqual(0)
    > 400 |         expect(document.querySelector('input').selectionEnd).toEqual(3)
          |                                                              ^
      401 |         expect(document.querySelector('input').selectionDirection).toEqual('backward')
      402 |     })
      403 | })
```
If you look at the test, it's really confusing. I don't know, maybe i am wrong.

muzafferdede:
> I guess it does make sense to only update the value if it has changed...

In that case, whole bind/model directive should check the old value before updating?

test/bind.spec.js
SimoTod: The test is not immediate but it does what it says. 
It has a listener on the select function, it simulates the user selection, the listener updates the model value and the model triggers the input update changing from bar to baz. At the end the test checks that, despite the value being changed. the selection is still the same.

It's not a real case imo because the model value would only change if you take some actions (like clicking buttons or similar) so the input would lose the focus. Also, since the content is changed, your old selection is likely to be wrong anyway.

This implementation seems neater but since there have been a lot of back and forward on this topic, can we make sure that it works correctly for each input fields in all the major browsers please (chrome, firefox, safari and edge)?
Manual tests are okay since we can't really have real automated tests for browsers. Thanks

SimoTod: see https://codepen.io/SimoTod/pen/LYpyjzz, if you select the test it changes but the selection disappear. To me, it's acceptable but the guys who wrote the test thought it was a bug so I guess it's subjected to personal opinions

SimoTod:
FYI Value binding happens once in Safari. 
It does have the bug where, if you set the value via code, it jumps to the last position.
Because you type the value directly on the input, the input updates the model and the model updates the input again with the same value which is a bit silly so your code prevent the second step avoiding safari from triggering the bug.
(Just to clarify, the fix is good but it's also good to understand the problem otherwise we just keep patching a broken thing).

test/bind.spec.js
muzafferdede: > see https://codepen.io/SimoTod/pen/LYpyjzz, if you select the test it changes but the selection disappear. To me, it's acceptable but the guys who wrote the test thought it was a bug so I guess it's subjected to personal opinions

I just tried your sample with Vue js, it behaves the same.

muzafferdede:
Hello everyone,
- I use Vuejs to implement a scenario which what we fixed in this PR and indeed this PR makes Alpine and Vuejs behave same in such a scenario now. 
- I have created a test codepen for this particular issue. if anyone fancy to check other browsers or play around : https://codepen.io/muzafferdede/pen/GRpmMdJ 
- I have test Safari/Chrome/Firefox/Breve browsers in macOS and i can confirm i haven't seen the issue again.
- I did few test Chrome/Edge `(mine might be old version since i don't use it often)` on Windows 10 and i can confirm i haven't seen the issue again.
#
> @SimoTod  Because you type the value directly on the input, the input updates the model and the model updates the input again with the same value which is a bit silly so your code prevent the second step avoiding safari from triggering the bug.

 That was what i want to say by describing it as 'binding multiple times'. Thanks for the better description. 
#
If you think Alpine needs all input tests to be implemented #407, or you think it needs further discussion on it, we can discuss on #407 and have a conclusion. The last thing we decided on that issue was wait for someone to pick that PR. 
#
P.S 
I forget the mention that issue #404 was a new issue IMO. Somehow i even didn't see that one at all. I might be wrong but It's funny, no one (including me) did check if the last PR did fix both issues or not. Everyone assumed tests passed so we are all good. I am still surprised that how come i didn't see it. 

I believe this PR will be the last pr on this issue. Hopefully, optimistically, fingers crossed :)

HugoDF:
> I have created a test codepen for this particular issue. if anyone fancy to check other browsers or play around : https://codepen.io/muzafferdede/pen/GRpmMdJ

Is there an element that this could be tested on the in index.html file that's in this repository? That's what will be used for manual sanity checks.

HugoDF:
Re #407 if you're keen you can go ahead and implement more tests 🙂 in a new PR

muzafferdede:
> > I have created a test codepen for this particular issue. if anyone fancy to check other browsers or play around : https://codepen.io/muzafferdede/pen/GRpmMdJ
> 
> Is there an element that this could be tested on in the index.html file that's in this repository? That's what will be used for manual sanity checks.

No. I am not sure what kind of setup would do the thing you mention. But there is not any hidden input example in the index.html to test. Maybe there are checkboxs and radioss but not sure if they were meant to be used for model binding. 
# 
>Re #407 if you're keen you can go ahead and implement more tests 🙂 in a new PR

Not at the moment.

HugoDF:
Thanks @muzafferdede I'll create a separate PR with what I meant for the index.html and might have a crack at the unit test thing later next week 🙂

Thanks for your contribution, let's see if we can get a "third time lucky" type of situation

PR #430, opened 2020-04-28
Added .once modifier to x-html and x-text

hkan:
I'm not sure if this modifier has a common-enough use case to be included in the library but I needed it and added it, so feel free to shoot it down if it is deemed unnecessary. Though it only adds 271 bytes to the raw output, and probably a minuscule amount to the gzipped output so it shouldn't be an issue performance-wise.

The way it works is that `x-html.once` and `x-text.once` basically does not respond to any updates in the component after the initialization.

### DIFF
--- a/README.md
+++ b/README.md
@@ -368,6 +368,11 @@ If you wish to customize this, you can specifiy a custom wait time like so:
 
 `x-text` works similarly to `x-bind`, except instead of updating the value of an attribute, it will update the `innerText` of an element.
 
+**`.once` modifier**
+**Example:** `<button x-text.once="foo"></button>`
+
+Adding the `.once` modifier to the `x-text` will ensure that the element content will only be set once at the initialization of component. This is useful when you need to show the data in its initial state.
+
 ---
 
 ### `x-html`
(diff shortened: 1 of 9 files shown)


SimoTod:
What's your use case exactly, if the property doesn't need to be reactive, can it just be printed in your html as usual? Or can you just use a different property which you won't update. It feels a bit weird that we use Alpine to render something that is then supposed to be static. Thanks

hkan:
I'm doing all this to replicate the `x-model` functionality on non-input elements.

I have an element with `contenteditable` attribute that I use like a `textarea`. If Alpine updates it on every `input` event, the caret goes to the start of the element, which effectively makes it unusable. With `.once` modifier I only set the content at the init and then just listen to the changes to update data.

> can it just be printed in your html as usual?

I guess I can just print it directly via PHP. At the time it felt wrong, not sure why.

**To be honest I opted out of this `x-text.once` myself.** I've modified my local copy of Alpine to retain the caret position after updating the content via [Selection APIs](https://developer.mozilla.org/en-US/docs/Web/API/Selection). The code is not in the best condition right now, but it works for me. Check the link below if you're curious. My changes are between 7th and 32nd lines except for line 19. I can make another PR if it makes sense to include in Alpine.

https://gist.github.com/hkan/56a7949e99dab0eec1e26f39c02c424c

SimoTod:
Yeah, I would say that x-text.once is not the right solution here. When v3
is out, you can probably add a custom directive for it. The only way it
could fit in the core is if we extend x-model to normal html element but it
would need a lot of talking.
For now, your best shot is to use your own fork since the feature seems
quite tailored on your specific use case.

hkan:
Alright, sounds fair.

SimoTod:
P.s. I'm not the author so you can keep it open if you want to wait for Caleb to give you some feedback.

hkan:
Ah, I didn't realise that. But after thinking about it, I do think `x-text.once` is unnecessary. Though I'd appreciate @calebporzio's and your feedback on the gist I linked above.

calebporzio:
Yeah, if you just need the listening part of x model, use: x-on:input="foo = $event.target.value"

PR #437, opened 2020-04-29
Fix whitespace bug in x-bind:class class list

### LINKED ISSUE #436: DOMException for trailing whitespace in x-bind:class
Hi,
I'm not sure if this is an issue of Alpine itself or just something JS specific.

**Alpine version:** latest 2.x.x (CDN)

The following error appears when a whitespace is added after the last class declaration:

`Uncaught (in promise) DOMException: Failed to execute 'add' on 'DOMTokenList': The token provided must not be empty.
    at https://cdn.jsdelivr.net/gh/alpinejs/alpine@v2.x.x/dist/alpine.min.js:7:8076
    at https://cdn.jsdelivr.net/gh/alpinejs/alpine@v2.x.x/dist/alpine.min.js:7:8053
    at x (https://cdn.jsdelivr.net/gh/alpinejs/alpine@v2.x.x/dist/alpine.min.js:7:8023)
    at https://cdn.jsdelivr.net/gh/alpinejs/alpine@v2.x.x/dist/alpine.min.js:7:18631
    at ie.resolveBoundAttributes (https://cdn.jsdelivr.net/gh/alpinejs/alpine@v2.x.x/dist/alpine.min.js:7:18456)
    at ie.initializeElement (https://cdn.jsdelivr.net/gh/alpinejs/alpine@v2.x.x/dist/alpine.min.js:7:17197)
    at https://cdn.jsdelivr.net/gh/alpinejs/alpine@v2.x.x/dist/alpine.min.js:7:16908`

## Code to reproduce

html (error):
```
<div x-data="{color: 'green'}">
    <h1 x-bind:class="{'green-text ': color === 'green'}">Does my color change</h1>
</div>
```

html (working)
```
<div x-data="{color: 'green'}">
    <h1 x-bind:class="{'green-text': color === 'green'}">Does my color change</h1>
</div>
```

styles:
```
.green-text {
  color: green;
}
```

Notice the whitespace in `'green-text '`. Removing the whitespace fixes it. 

I ran into this today and it wasn't immediately obvious to me why it broke. Especially using tailwind makes it hard to spot this, when the class declarations get longer.

felixsc:
Fixes #436 

When the class list declaration in a x-bind:class is ended with a whitespace or contains multiple consecutive whitespaces an error is thrown.

This PR filters the class list after it is split and removes the empty classes.

### DIFF
--- a/src/directives/bind.js
+++ b/src/directives/bind.js
@@ -65,14 +65,14 @@ export function handleAttributeBindingDirective(component, el, attrName, express
 
             keysSortedByBooleanValue.forEach(classNames => {
                 if (value[classNames]) {
-                    classNames.split(' ').forEach(className => el.classList.add(className))
+                    classNames.split(' ').filter(Boolean).forEach(className => el.classList.add(className))
                 } else {
-                    classNames.split(' ').forEach(className => el.classList.remove(className))
+                    classNames.split(' ').filter(Boolean).forEach(className => el.classList.remove(className))
                 }
             })
         } else {
             const originalClasses = el.__x_original_classes || []
-            const newClasses = value.split(' ')
+            const newClasses = value.split(' ').filter(Boolean)
             el.setAttribute('class', arrayUnique(originalClasses.concat(newClasses)).join(' '))
         }
     } else if (isBooleanAttr(attrName)) {
--- a/test/bind.spec.js
+++ b/test/bind.spec.js
@@ -420,3 +420,14 @@ test('setSelectionRange is not called for inapplicable input types', async () =>
         expect(document.querySelector('input').value).toEqual('baz')
     })
 })
+
+test('unnecessary whitespaces in class list are ignored', async () => {
+    document.body.innerHTML = `
+    <div x-data="{ foo: 'bar' }">
+        <span x-bind:class="{'   foo:class ': foo === 'bar' }"></span>
+    </div>
+    `
+    Alpine.start()
+
+    expect(document.querySelector('span').classList.contains('foo:class')).toBeTruthy()
+})
--- a/dist/alpine-ie11.js
+++ b/dist/alpine-ie11.js
@@ -1143,7 +1143,7 @@
   (module.exports = function (key, value) {
     return sharedStore[key] || (sharedStore[key] = value !== undefined ? value : {});
   })('versions', []).push({
-    version: '3.6.5',
+    version: '3.6.4',
     mode:  'global',
     copyright: '© 2020 Denis Pushkarev (zloirock.ru)'
   });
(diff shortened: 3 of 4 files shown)


src/directives/bind.js
```diff
@@ -65,14 +65,14 @@ export function handleAttributeBindingDirective(component, el, attrName, express
 
             keysSortedByBooleanValue.forEach(classNames => {
                 if (value[classNames]) {
-                    classNames.split(' ').forEach(className => el.classList.add(className))
+                    classNames.split(' ').filter(Boolean).forEach(className => el.classList.add(className))
                 } else {
-                    classNames.split(' ').forEach(className => el.classList.remove(className))
+                    classNames.split(' ').filter(Boolean).forEach(className => el.classList.remove(className))
```
HugoDF, line 70: Would it make sense to have a `getClassNames` util/helper that does the split + filter?

muzafferdede:
@HugoDF  Yeah, There are some changes mixed up with the PR. Build files has some ghost changes.

felixsc:
Yeah, it felt a bit odd to me as well.

I didn't do anything besides installing the dependencies with
`yarn`
and building it after writing the changes with
`npm run build`
Any idea what might have caused it? Should I've done it differently?

src/directives/bind.js
felixsc: Yes, makes sense. It's repeated three times in this PR

SimoTod:
Weirdly, it seems like `yarn install` didn't use the package-lock and pulled a new version of core-js (it should have used this one https://github.com/felixsc/alpine/blob/master/package-lock.json#L2699).

Would you mind deleting the node-folder, using `npm install` and committing the dist folder again please? Thanks

The dependency should probably be updated but not in the scope of this PR. Thanks for your fix, it looks good.

SimoTod:
Right, Yarn doesn't use package-lock unless you run `yarn import` first.

`As of Yarn 1.7.0, you can import your package-lock.json state, generated by npm to Yarn, by using yarn import.`

Mistery solved.

felixsc:
Aha, didn't know that. TIL
Thanks for sharing

test/bind.spec.js
```diff
@@ -420,3 +420,14 @@ test('setSelectionRange is not called for inapplicable input types', async () =>
         expect(document.querySelector('input').value).toEqual('baz')
     })
 })
+
+test('unnecessary whitespaces in class list are ignored', async () => {
```
HugoDF, line 424: Do we need to have tests for when it's using array syntax or string syntax?

calebporzio:
Good fix! Thanks @felixsc!

Issue #445, opened 2020-05-01
Initiate components in separate try/catch blocks


carlmjohnson:
Given this code:

```
<html>
  <script src="https://cdn.jsdelivr.net/gh/alpinejs/alpine@v2.x.x/dist/alpine.min.js" defer></script>
  <body>
    <div x-data="{x: 1}">
      <span x-text="x">loads</span>
    </div>
    <div x-data="foo()">
    </div>
    <div x-data="{x: 1}">
      <span x-text="x">does not load</span>
    </div>
  </body>
</html>
```

The first span gets replace by 1 and the last span does not because the middle component fails to load.

In the real world, this is a problem because you use a feature with only partial support like IntersectionObserver or fetch, it is possible for your components to fail to load in some browsers but not others, and you don't notice this during development.

I think one way to fix with would be to just run every initiation in its own try/catch block.

SimoTod:
👍 If you have a fix in mind, feel free to drop a PR, the project is open source and any contribution is more than welcome. 

P.s. If you're using an unsupported feature without a polyfill, one or more components will still be broken so you still need to test your page on multiple browsers. 
Even with the try catch you won't see the failure during development (and it will be harder to spot since only a minimal part of your website will be broken).

carlmjohnson:
Yeah, the idea is that I don't mind if my scroll thingie is broken as long the most popular fetcher isn't broken too. I'll see if there's anything I can contribute or it's all over my head JS magic… :-)

PR #447, opened 2020-05-02
Initialize components in try/catch blocks

carlmjohnson:
Re #445 .

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -76,7 +76,13 @@ const Alpine = {
 
     initializeComponent: function (el) {
         if (! el.__x) {
-            el.__x = new Component(el)
+            try {
+                el.__x = new Component(el)
+            } catch (err) {
+                window.setTimeout(() => {
+                    throw err;
+                }, 0);
+            }
         }
     },
 
--- a/dist/alpine-ie11.js
+++ b/dist/alpine-ie11.js
@@ -7175,8 +7175,18 @@
       observer.observe(targetNode, observerOptions);
     },
     initializeComponent: function initializeComponent(el) {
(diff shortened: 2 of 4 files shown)


src/index.js
```diff
@@ -76,7 +76,13 @@ const Alpine = {
 
     initializeComponent: function (el) {
         if (! el.__x) {
-            el.__x = new Component(el)
+            try {
+                el.__x = new Component(el)
+            } catch (err) {
```
HugoDF, line 81: Should we be logging the error? Or are we saying "let the browser log the error but not crash the whole JS thread"

mblarsen: I guess throwing it would be at par with what is happening now, but without "crashing" other components.

They are allowed to init before the error is thrown.

carlmjohnson: Yes. If a crashed happens, it still crashes up to the global window.onerror but at least the other components will still initialize.

An alternative design might be to stash the crashes in some Alpine specific location, but since the crashes shouldn’t happen anyway, I like this solution more.

HugoDF: Sounds good to me.

bep: I like this.

carlmjohnson:
Here’s an alternative idea: there could be a Alpine.onerror and its default is to reraise the error, as is done in this case.

carlmjohnson:
I think this is good to merge as is, and if Alpine gets a system of config, the onerror handler can be added to that later. Any disagreement?

SimoTod:
I don't think we need onerror at this stage. Your PR looks good.

PR #449, opened 2020-05-02
MutationObserver reset x-data when nested components are moved around

SimoTod:
Fixes #442 

~~I don't know how to write a good test for it.~~

UPDATE: Also, added a basic tests

PR #455, opened 2020-05-04
Fix Turbolinks error situations

### LINKED ISSUE #319: data-turbolinks-permanent: Uncaught ReferenceError on first navigation

### LINKED ISSUE #372: What is the Best Way to Ensure Idempotent Transformations?
My project is using Alpine, Rails, and Turbolinks in Rails.

I'm using **x-for** to create img nodes for each image source in an array.
```
<template x-for="(imgSrc, index) in items" :key="index">
    <img :src="imgSrc" />
</template>
```

When I navigate to the page, the above code inserts image nodes for a carousel as expected. However, when navigating away, Turbolinks saves a copy of the transformed page to its cache. I then press the Back button and Turbolinks restores the page (and all previously inserted nodes) from the cache and **the above code inserts a second set of image nodes again.**

What is the best way to ensure idempotent transformations so that Alpine does not re-render nodes when they are already on the page from caching?

I have thought about using a rendered variable and then **x-if** on the template checking for the rendered variable, like below, but have not fully fleshed out how this might work.
```
<div x-data="{ rendered: false }" >
  <template x-if="!rendered" x-for="(imgSrc, index) in items" :key="index">
    <img :src="imgSrc" />
  </template>
</div>
```

What is the best way to ensure idempotent transformations so that Alpine does not re-render nodes when they are already on the page from caching?

bep:
This commit disconnects the mutation observer on `turbolinks:before-cache` and then clean up elements created by Alpine directives, unless marked as `data-turbolinks-permanent`.

The mutiation observer is reconnected at `turbolinks:load`.

This fixes two error situations when Alpine is combined with Turbolinks:

* Duplicate elements on navigation with the back or forward button.
* Errors/data races on Turbolinks navigation with Alpine components marked with `data-turbolinks-permanent`.

Fixes #319
Fixes #372

This PR is continuing the foundation from @SimoTod 's work.

No unit tests (that looks hard), but the manual test cases for this passes for me.

I can add that I have tested this OK on a "real app" that uses Turbolinks in combo with AlpineJS with this patch.

### DIFF
--- a/dist/alpine-ie11.js
+++ b/dist/alpine-ie11.js
@@ -7066,25 +7066,9 @@
                   _newArrowCheck(this, _this);
 
                   this.initializeComponent(el);
-                }.bind(this)); // It's easier and more performant to just support Turbolinks than listen
-                // to MutationObserver mutations at the document level.
-
-                document.addEventListener("turbolinks:load", function () {
-                  var _this2 = this;
-
-                  _newArrowCheck(this, _this);
-
-                  this.discoverUninitializedComponents(function (el) {
-                    _newArrowCheck(this, _this2);
-
-                    this.initializeComponent(el);
-                  }.bind(this));
                 }.bind(this));
-                this.listenForNewUninitializedComponentsAtRunTime(function (el) {
-                  _newArrowCheck(this, _this);
-
-                  this.initializeComponent(el);
-                }.bind(this));
+                this.listenForNewUninitializedComponentsAtRunTime();
+                this.initTurbolinksListeners();
 
               case 6:
               case "end":
@@ -7101,50 +7085,49 @@
       return start;
     }(),
     discoverComponents: function discoverComponents(callback) {
-      var _this3 = this;
+      var _this2 = this;
 
       var rootEls = document.querySelectorAll('[x-data]');
       rootEls.forEach(function (rootEl) {
-        _newArrowCheck(this, _this3);
+        _newArrowCheck(this, _this2);
 
         callback(rootEl);
       }.bind(this));
     },
     discoverUninitializedComponents: function discoverUninitializedComponents(callback) {
-      var _this4 = this;
+      var _this3 = this;
 
       var el = arguments.length > 1 && arguments[1] !== undefined ? arguments[1] : null;
       var rootEls = (el || document).querySelectorAll('[x-data]');
       Array.from(rootEls).filter(function (el) {
-        _newArrowCheck(this, _this4);
+        _newArrowCheck(this, _this3);
 
         return el.__x === undefined;
       }.bind(this)).forEach(function (rootEl) {
-        _newArrowCheck(this, _this4);
+        _newArrowCheck(this, _this3);
 
         callback(rootEl);
       }.bind(this));
     },
-    listenForNewUninitializedComponentsAtRunTime: function listenForNewUninitializedComponentsAtRunTime(callback) {
-      var _this5 = this;
+    listenForNewUninitializedComponentsAtRunTime: function listenForNewUninitializedComponentsAtRunTime() {
+      var _this4 = this;
 
-      var targetNode = document.querySelector('body');
       var observerOptions = {
         childList: true,
         attributes: true,
         subtree: true
       };
-      var observer = new MutationObserver(function (mutations) {
-        var _this6 = this;
+      this.observer = new MutationObserver(function (mutations) {
+        var _this5 = this;
 
-        _newArrowCheck(this, _this5);
+        _newArrowCheck(this, _this4);
 
         for (var i = 0; i < mutations.length; i++) {
           if (mutations[i].addedNodes.length > 0) {
             mutations[i].addedNodes.forEach(function (node) {
-              var _this7 = this;
+              var _this6 = this;
 
-              _newArrowCheck(this, _this6);
+              _newArrowCheck(this, _this5);
 
               // Discard non-element nodes (like line-breaks)
               if (node.nodeType !== 1) return; // Discard any changes happening within an existing component.
@@ -7152,7 +7135,7 @@
 
               if (node.parentElement && node.parentElement.closest('[x-data]')) return;
               this.discoverUninitializedComponents(function (el) {
-                _newArrowCheck(this, _this7);
+                _newArrowCheck(this, _this6);
 
                 this.initializeComponent(el);
               }.bind(this), node.parentElement);
@@ -7160,7 +7143,7 @@
           }
         }
       }.bind(this));
-      observer.observe(targetNode, observerOptions);
+      this.observer.observe(document.body, observerOptions);
     },
     initializeComponent: function initializeComponent(el) {
       if (!el.__x) {
@@ -7171,6 +7154,71 @@
       if (!newEl.__x) {
         newEl.__x = new Component(newEl, component.getUnobservedData());
       }
+    },
+    initTurbolinksListeners: function initTurbolinksListeners() {
+      var _this7 = this;
+
+      // It's easier and more performant to just support Turbolinks than listen
+      // to MutationObserver mutations at the document level.
+      document.addEventListener('turbolinks:load', function () {
+        var _this8 = this;
+
+        _newArrowCheck(this, _this7);
+
+        Alpine.discoverUninitializedComponents(function (el) {
+          _newArrowCheck(this, _this8);
+
+          this.initializeComponent(el);
+        }.bind(this));
+        this.observer.observe(document.body, {
+          childList: true,
+          attributes: true,
+          subtree: true
+        });
+      }.bind(this)); // Disconnect the the mutation observer to avoid data races, and then
+      // clean up the elements created by Alpine directives before Turbolinks
+      // stores it to cache, unless it's marked as permanent.
+      // The mutiation observer will be reconnected by the turbolinks:load event.
+
+      document.addEventListener('turbolinks:before-cache', function () {
+        var _this9 = this;
+
+        _newArrowCheck(this, _this7);
+
+        this.observer.disconnect();
+        walk(document.body, function (el) {
+          _newArrowCheck(this, _this9);
+
+          if (el.hasAttribute('data-turbolinks-permanent')) {
+            return false;
+          }
+
+          if (el.hasAttribute('x-for')) {
+            var nextEl = el.nextElementSibling;
+
+            while (nextEl) {
+              var currEl = nextEl;
+              nextEl = nextEl.nextElementSibling;
+
+              if (typeof currEl.__x_for_key !== 'undefined') {
+                currEl.remove();
+              }
+            }
+
+            return true;
+          }
+
+          if (el.hasAttribute('x-if')) {
+            var ifEl = el.nextElementSibling;
+
+            if (ifEl && typeof ifEl.__x_inserted_me !== 'undefined') {
+              ifEl.remove();
+            }
+          }
+
+          return true;
+        }.bind(this));
+      }.bind(this));
     }
   };
 
--- a/dist/alpine-ie11.js
+++ b/dist/alpine-ie11.js
@@ -7045,6 +7045,7 @@
 
   var Alpine = {
     version: "2.3.3",
+    pauseObserver: false,
     start: function () {
       var _start = _asyncToGenerator( /*#__PURE__*/regeneratorRuntime.mark(function _callee() {
         var _this = this;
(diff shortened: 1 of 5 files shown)


ryangjchandler:
Other than some code style differences, this all seems pretty good to me. I'm going to open another PR that adds an .editorconfig file to the repo, should help with the 2 space indents / 4 space indents differences. Nice work @bep & @SimoTod

astrocket:
I faced the same issue today. Hope it gets merged soon 👍

bep:
I have 

* rebased this against master
* fixed the formatting issues (which should make this easier to reason about ...)

The related manual test cases passes. I do, however, see a x-for bug coming in from master that confuses me a little ....

bep:
I see that this 254330253a82ea5a1781354d0c7fab0b4e5b18b5 introduced (or discovered) another related issue in my app... I now see there are 2 `MutationObserver`'s in play, and this PR only covers the first. I will see if I can get that covered in this PR.

SimoTod:
Can we replicate it in the manual test? Have you figured out when it happens?

bep:
>Can we replicate it in the manual test?

Probably ... I suspect the cause is more or less the same: A data race between Alpine and Turbolinks, but since there is one mutation observer per component (making a really busy application), it is hard to pause-em-all ... But yes, a failing test case would be a start.

SimoTod:
Not sure if it's the same issue (it seems to fail even when reverting that changes) but
```html
        <h2>This item should not add duplicate span items nor errors when navigating away and back to this page</h2>
        <div x-data="{ foo: 'bar' }" id="bob" data-turbolinks-permanent>
            <div x-data="{ bob: ['bar', 'baz'] }">
                <template x-for="item in bob">
                    <span x-text="item"></span>
                </template>
            </div>
        </div>
```
breaks as well

bep:
I noticed that in my failing app, that if I remove the root component (only used for navigation state), the error goes away. It is probably the same issue as in your test case above, but I have currently no idea how to fix it.

SimoTod:
Yeah, I had the feeling that it was a tricky one when I closed my PR, to be fair.

Small update: Maybe it's not related to nested component or maybe it's another issue, but if I add
```html
        <h2>This item should not add duplicate span item nor errors when navigating away and back to this page</h2>
        <div x-data="{ bob: ['bar', 'baz'] }" id="test" data-turbolinks-permanent>
            <template x-for="item in bob">
                <span x-text="item"></span>
            </template>
        </div>
```
to the first page but NOT on the second page, when I navigate back I get the error.
It seems like it forgets about the __x attribute (Alpine component data and metadata) somehow.

bep:
Well, my 50 cents before I take the Saturday evening off:

* I suspect that it would have been more effective to have only one (or possibly 2) `MutationObserver` instances. I suspect there are some memory/gc/performance issues with the current setup, esp. if for some reason have a busy component tree (I noticed this in my app when I added a root component and updated to the latest Alpine (with the fix mentioned above).
* If you then, instead of `disconnect` and then `observe` (as in this PR) add some kind of `pause` flag, I suspect that your and mine tests will pass.
* The last part, I guess, can be confirmed by adding some global flag.

SimoTod:
Something to try, but the issue seems to happen before we reconnect the observer when Alpine calls
```html
Alpine.discoverUninitializedComponents(el => {
    this.initializeComponent(el);
});
```
inside the `'turbolinks:load'` callback. Not 100% sure though.

I'll think about that and let you know if I have any ideas. I'm off as well. Have a nice evening.

bep:
Scratch what I said above. Looking at it a little more, I suspect what's happening is that:

* When Turbolinks is done with it's "cache restore", all nodes will appear as new nodes to the MutationObserver.
* The above would also explain why my computer got "glowing hot" when I navigated to a new page.

I did a quick test by adjusting

https://github.com/alpinejs/alpine/blob/master/src/component.js#L354

To

```js
if (mutations[i].addedNodes.length > 0) {
  mutations[i].addedNodes.forEach((node) => {
    if (node.nodeType !== 1 ||  node.__x !== undefined) {
      return;
    }

   if (node.matches('[x-data]')) {
	node.__x = new Component(node);
   }
   this.initializeElements(node);

}
```

Which makes my test case work fine. Note that I doubt the above is a real fix, but I cannot see how the original behaviour could be correct when used with Turbolinks.

bep:
Adding a ` node.__x != 'undefined'` to `listenForNewElementsToInitialize` (`if (node.nodeType !== 1 || node.__x != 'undefined' || node.__x_inserted_me) return;)` fixes my test case. It looks like these nodes comes up as `addedNodes` once Turblink is done restoring from cache. I still suspect Apline is doing too much work in this situation, but I don't see how the above patch could hurt.

SimoTod:
It makes sense. I'll have a look tomorrow but it sounds promising.
Probably we could do with a bit of analysis on the mutation observers at some point. I feel like the generic and the component one overlap sometime.

bep:
@SimoTod I will update this PR in a bit... I think I have something that also covers your test cases above.

SimoTod:
@bep, just had a look. I actually had a similar fix for a different error #449
Yours seems slightly different (maybe better) because it also skips the last line. I guess we don't need to refresh the component in that case.

bep:
I have a local fix for your test cases (will push later today) -- there is one remaining issue (hehe), which seems to be caused by the`turbolinks:load` being triggered "too early". I will look into it.

bep:
@SimoTod thanks, I'm looking into your test cases above now ... a mind twister.

bep:
I'm working on a patch that cover the test cases.

My main problem to wrap my head around is:

* If you clear out the Alpine-state on all x-data unless they live below `data-turbolinks-permanent`, that works great for all target navigations that also contain this component. If not, you get an error in the console when you return. 
* If you clear out the Alpine-state on all x-data before Turbolinks save to cache, this (kind of) works, but creates this very annoying flickering effect.

The prime use case for `data-turbolinks-permanent` is left bar navigation etc., so that flickering effect is kind of not usable, but I'm having some troubles working my head around it, as you don't know anything about the target when doing the "state cleaning".

SimoTod:
Turbolinks also has a before-render hook containing old body and new body maybe we can clean the DOM on that event and skip permanent element only if they exist in the old body.

bep:
Given a little bit more thought on this, I think the very best we can do is a slight variation of the current state of this PR. Some discussions on this over at https://github.com/turbolinks/turbolinks/issues/191#issuecomment-421790372 -- and I think people having issues with `data-turbolinks-permanent` need to handle this themselves, by disabling Turbolinks on some outgoing links, or maybe doing some tricker in `before-cache` on their own.

bep:
@SimoTod I have explored `before-render` and what you mention, and that just does not work (or: it does not solve the main challenge), but I would be happy to be proven wrong...

SimoTod:
I haven't tried, i just saw the event in their documentation but i don't know if what we receive it's suitable for our needs or it's already too late. I'll have a look at that discussion and see if it triggers some other ideas. 👍

SimoTod:
I need more testing but I think I got it. (it's based on your pr + the before-render hook). I just need to check that it's not going to be a performance nightmare with big pages.

bep:
@SimoTod I pushed some additional test cases to this PR-branch. I have gone a few rounds on this now, and I have had a few "oh, now I got it" turned into "not really".

SimoTod:
Ta, I'll have a look tonight.

SimoTod:
@bep So this is my latest attempt on top of your existing PR: https://github.com/alpinejs/alpine/compare/master...SimoTod:bug/turbolink-fixes-2?expand=1
it seems to work with your additional tests as well.

I added a flag instead of destroying and recreating the observer because when I spoke to Caleb last week, it didn't seem keen on that solution (minor detail, it shouldn't change the general implementation).

I mark all the alpine generated element on before-cache. I use an html attribute here because turbolinks uses cloneNode which destroys other non-standard properties. I don't remove anything yet because I don't know if it's going to be used in the next page.

after before-cache turbolinks does its data-permanent replacements

On before-render I check all the previously marked element, if they don't have custom alpine properties, it means that they come from the cache, they are not good so I remove them.
Otherwise it means that they come from the data-permanent swapping so we keep them.

Do you think it could work?

bep:
@SimoTod I didn't think this would be possible, but you proved me wrong. I'm impressed, this is much appreciated. And I agree about the pause flag being a much better solution.

The manual tests passes for me, to, and then I have this "real app" which is a heavy Turbolinks user that also works perfectly (at least in this area).

I'll merge your your commit into this PR and (hopefully) call it a weekend ...

Thanks again.

bep:
@calebporzio a ... friendly bump of this. I have a customer project that hangs on this. Not your problem, I understand, but it's a cool site that should give Alpine some good PR. 

If this PR seems too big, one option would maybe be to just expose a way to pause the "observer mutator" -- which should allow the client to do the needed DOM cleanups in the Turbolinks events. @SimoTod may way in on that option.

Anyway, it would be good with some feedback, worst case I can then cut my losses sooner rather than later.

SimoTod:
I'm happy to decouple the implementation and help maintaining an official plugin under the Alpine namespace if we don't want to add it into the core as well.

calebporzio:
Here are my thoughts:

At first, it was really easy to support TL directly in the core.

This PR makes the core more complex solely to support TL.

There are things in this PR that make me uncomfortable, like introducing an extra "x-generated" concept.

Those are big things that need to be thought through before adding to Alpine.

That said, I think TL support is extremely useful, but I'm thinking at this point we might be better off with a separate plugin, thanks for being willing to maintain that @SimoTod, and thanks for all the effort on this @bep.

Going to close this for now.

Issue #257, opened 2020-03-09
Binding radios with true/false


kez:
I've been attempting to find some `boolean` database columns (via an API and `fetch()`) to 2 grouped radio buttons.  One radio should represent 'true' and the other 'false'.

There seems to be a disconnect between the value on the radio which ends up as a string representation of the boolean versus the bound data which is a proper boolean per JSON.

I've put up a simple demo of the issue here (https://codepen.io/k-dobson/pen/WNvdGeB) - I would have expected all three of these to be set to `checked`, but you can see that only happens when the data model is set as strings from the `x-model` bind.

Not sure if this is a bug, per say, or something I've missed in the docs/examples. Any insight appreciated!

SimoTod:
Hi @kez 

The value of a radio button is always a string when you read it using javascript while the values in your x-data object are booleans.
if you try `'true' == true` in dev console, you can see that it evaluates to false.

At the moment, Alpine doesn't convert it for you. It supports the `number` modifier on x-model directives (although it's not documented) but we don't have a boolean modifiers yet.

If @calebporzio is okay with that, we can create a PR to add it and update the documentation.

In the meantime, you need to make sure that your values are strings if you can.

kez:
Thanks @SimoTod - totally makes sense.  Given the variable nature of the `value` field of radios or checkboxes (e.g. could be set to "on", "true") it makes to force the casting upstream somewhere.  I've got a `toBoolean` helper function to satisfy various `x-show` directives which works fine with the mix of booleans and strings.

Another option could be an extension of the `:attribute` handlers to somehow allow the evaluation of a function to set the `checked` attribute (it's obviously not the same as a `:class='abc xyz' because inputs are just looking for "checked".  That way `x-model` could bind to the value and the `checked` status could be driven by a user function to handle the values directly.

(I've only been using Alpine for a week so forgive me if the above sounds like complete garbage).

I'll cast my booleans in the API response for now and keep an eye out in the future - happy to close this off.

SimoTod:
If you don't need to keep the value bound to a model, you can already have something like
```javascript
        <input :checked="radios.first + '' == 'true'" type="radio" value="true">First <br />
```
of add the relevant function to x-data and have something like 
```javascript
        <input :checked="ifFirstChecked()" type="radio" value="true">First <br />
```
or 
```javascript
        <input :checked="isChecked(radios.first, 'true')" type="radio" value="true">First <br />
```
or any other implementation you think it suits your existing code.

[...]

PR #462, opened 2020-05-05
Fix checkbox bind with value checked by default

### LINKED ISSUE #459: Binding a value on checkboxes = checked by default
Is there a reason why `x-bind:value` on checkboxes check themselves by default?

I **think** this may be a bug. See CodePen below as an example.

https://codepen.io/richardcool/pen/eYpyMWB

jokarz:
Fixes #459 without breaking other test cases. Added testing for radio input as well.

### DIFF
--- a/src/directives/bind.js
+++ b/src/directives/bind.js
@@ -1,4 +1,4 @@
-import { arrayUnique , isBooleanAttr } from '../utils'
+import { arrayUnique, isBooleanAttr } from '../utils'
 
 export function handleAttributeBindingDirective(component, el, attrName, expression, extraVars, attrType) {
     var value = component.evaluateReturnExpression(el, expression, extraVars)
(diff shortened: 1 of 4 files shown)


HugoDF:
@jokarz with those radio tests, would this PR also fix https://github.com/alpinejs/alpine/issues/257?

jokarz:
Not really though. The test cases added simply made sure that radio and checkbox are unchecked by default when other directives, aside from `:value`, are not present. (only checkbox is facing this issue)

minhchu:
Awesome @jokarz 
I'm having the same issue. Rendering an array of checkbox with `:value` and they are all checked.
Hope this PR will be merged soon.

src/directives/bind.js
```diff
@@ -79,7 +80,7 @@ export function handleAttributeBindingDirective(component, el, attrName, express
         }
     } else if (isBooleanAttr(attrName)) {
         // Boolean attributes have to be explicitly added and removed, not just set.
-        if (!! value) {
+        if (!!value) {
```
HugoDF, line 83: Could we get rid of this change?

jokarz: Updated it

PR #465, opened 2020-05-08
Update x-on, $event and $dispatch docs.

atomgiant:
Added an example of how to access $event and $dispatch inside of a
Javascript function.  This also includes adding an example of passing in
a Javascript name to x-on.

### DIFF
--- a/README.md
+++ b/README.md
@@ -544,7 +544,7 @@ These behave exactly like VueJs's transition directives, except they have differ
 
 > Note: The $event property is only available in the DOM.
 
-If you need to access $event inside of a Javascript function you can pass it in directly:
+If you need to access $event inside of a JavaScript function you can pass it in directly:
 
 `<button x-on:click="myFunction($event)"></button>`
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -562,6 +574,12 @@ You can also use `$dispatch()` to trigger data updates for `x-model` bindings. F
 </div>
 ```
 
+> Note: The $dispatch property is only available in the DOM.
+
+If you need to access $dispatch inside of a Javascript function you can pass it in directly:
```
HugoDF, line 579:
```suggestion
If you need to access $dispatch inside of a JavaScript function you can pass it in directly:
```

README.md
```diff
@@ -536,6 +542,12 @@ These behave exactly like VueJs's transition directives, except they have differ
 
 `$event` is a magic property that can be used within an event listener to retrieve the native browser "Event" object.
 
+> Note: The $event property is only available in the DOM.
+
+If you need to access $event inside of a Javascript function you can pass it in directly:
```
HugoDF, line 547:
```suggestion
If you need to access $event inside of a JavaScript function you can pass it in directly:
```

README.md
atomgiant: Thanks @HugoDF 👍

calebporzio:
Thanks!

PR #469, opened 2020-05-10
fix: polyfill SVGElement.prototype.contains for IE

### LINKED ISSUE #468: Add Node.contains() polyfill to IE build
Internet Explorer only supports `Node.prototype.contains` for HTML elements, not for SVG.
So something like this fails (but works if you wrap the SVGs in a span and put the directive on that):

```html
<button x-data="{ isOpen: false }">
  <svg x-show="!isOpen"><!-- open icon --></svg>
  <svg x-show="isOpen"><!-- close icon --></svg>
</button>
```

jonaskuske:
fix #468 

IE actually has two deficiencies regarding `.contains()`: it doesn't support it for SVG elements and it also doesn't support it on `document`.

This PR only polyfills `SVGElement`, not the more generic `Node` (which would require a custom implementation instead of this one-liner), so `document.contains()` still doesn't work – but Alpine doesn't need that and users who rely on it can polyfill `Node.prototype.contains` themselves.  

So polyfilling `SVGElement` should suffice I think – opinions?

### DIFF
--- a/src/polyfills.js
+++ b/src/polyfills.js
@@ -6,3 +6,5 @@ import "element-remove"
 import "classlist-polyfill"
 import "@webcomponents/template"
 import "custom-event-polyfill"
+
+SVGElement.prototype.contains = SVGElement.prototype.contains || HTMLElement.prototype.contains


SimoTod:
I agree. Alpine doesn't use/need document.contains 👍

HugoDF:
Agree as well, we should only polyfill what would break Alpine.js directly, nice one 👍

calebporzio:
Thanks!

PR #479, opened 2020-05-12
Expose $el in x-data

### LINKED ISSUE #471: Expose $el while initializing x-data
Would be really cool if `$el` was already available inside `x-data`, so we can set initial state that depends on the DOM element. My use case is fading in an image, where I need to set the initial state so that it reflects whether the image already finished loading before Alpine was ready or not:

```html
<img
  x-data="{ loaded: false }"
  x-init="loaded = Boolean($el.complete && $el.naturalWidth && $el.naturalHeight)"
  @load="loaded = true"
/>
```

Would be cool if I could skip the `x-init` here and directly initialize `x-data` to the correct value instead of setting a placeholder. So in the best case I could do:

```html
<img x-data="{ loaded: isImageLoaded($el) }" @load="loaded = true" />
```

jonaskuske:
This PR makes `$el` available within the `x-data` expression, for the reasons outlined in https://github.com/alpinejs/alpine/issues/471.
It also adds a notice about the availability of magic properties within `x-data` to the documentation.

close #471

### DIFF
--- a/README.md
+++ b/README.md
@@ -504,6 +504,8 @@ These behave exactly like VueJs's transition directives, except they have differ
 
 ### Magic Properties
 
+> With the exception of `$el`, magic properties are **not available within `x-data`** as the component isn't initialized yet.
+
 ---
 
 ### `$el`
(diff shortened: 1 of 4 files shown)


HugoDF:
@jonaskuske could you add a test for $el usage in x-data?

jonaskuske:
@HugoDF Done, and fixed a bug – turns out initialization is not the only time `x-data` is evaluated. 👍🏻

calebporzio:
Thanks for this. I think `$el` should be a magic variable for every single expression.

PR #498, opened 2020-05-16
Feature - CSS animations

SimoTod:
Right now, Alpine only considers the transition-duration property.
Animation (and animation libraries such ad animate.css) usually define an animation-duration property instead. This means that Alpine runs all the stage steps in less then a millisecond and animations don't have time to complete.

This PR makes easier for developers to use css animations and css animation libraries without workarounds. Note, at the moment is already possible to set a transition-duration property matching the animation length but it's counterintuitive so this PR is more a nice to have.

Vue js for example support it without workarounds: https://vuejs.org/v2/guide/transitions.html#Custom-Transition-Classes

Original discussion: https://github.com/alpinejs/alpine/discussions/489

Mini demo: https://codepen.io/SimoTod/pen/jObeYpQ


calebporzio:
Thanks Simone!

PR #500, opened 2020-05-17
Minor improvement to readme file

SimoTod:
People seems to use the 2.x.x CDN link in production so when we release a new version, if there are bug fixings changing the behaviour or accidental BCs, they'll get bugs straight into production without updating anything.

It should be obvious, but we should suggest they use a specific version number in production.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ Think of it like [Tailwind](https://tailwindcss.com/) for JavaScript.
 
 That's it. It will initialize itself.
 
-For production environments, it's recommend linking to a specific version number to avoid unexpected breakage from newer versions. 
+For production environments, it's recommended to pin a specific version number in the link to avoid unexpected breakage from newer versions. 
 For example, to use version `2.3.5`:
 ```html
 <script src="https://cdn.jsdelivr.net/gh/alpinejs/alpine@v2.3.5/dist/alpine.min.js" defer></script>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -22,6 +22,12 @@ Think of it like [Tailwind](https://tailwindcss.com/) for JavaScript.
 
 That's it. It will initialize itself.
 
+For production environments, it's recommend linking to a specific version number to avoid unexpected breakage from newer versions.
```
HugoDF, line 25: maybe?
```suggestion
For production environments, it's recommended to pin a specific version number in the link to avoid unexpected breakage from newer versions. 
```

README.md
```diff
@@ -22,6 +22,12 @@ Think of it like [Tailwind](https://tailwindcss.com/) for JavaScript.
 
 That's it. It will initialize itself.
 
+For production environments, it's recommended to pin a specific version number in the link to avoid unexpected breakage from newer versions.
```
ryangjchandler, line 25: I'd probably add some bold text here too so it's a bit more obvious.

SimoTod: We can do, although people won't read the readme anyway. 😂
I'll leave it with Caleb if he wants it bold or not or he wants to reword it. To me it just needs to be thee and it's already okay since it's not part of a very long paragraph (same as Vue's documentation). 👍

Issue #228, opened 2020-03-04
Fire events after load, update, reload, etc


ryangjchandler:
I think this would be useful for anyone who wants to hook into Alpine from outside of their components, possibly inside of a scoped JS file.

I have a working branch for this, using the dispatch() function from Livewire. I'm also working on a devtools panel for alpine and this would be super useful so I don't have to mess around with more Proxies etc, cc/ @calebporzio

ryangjchandler:
After some dispute on the events that should be fired, to begin with I'm going to keep the list short:

* alpine:loaded
* alpine:transition-start
* alpine:transition-end

ryangjchandler:
Closing this issue for now, since this will probably be re-evaluated for v3.

TheMangoTrain:
This is exactly what's needed for something like sequential animations, instead of trying to fake it with animation delays.

PR #502, opened 2020-05-19
Improve $nextTick for transitions 

ahmedkandel:
When using transitions, `$nextTick` is called before the DOM is updated as discussed in #491, A workaround could be to `setTimeout` inside `$nextTick` callback but it is not sufficient as the transition uses `requestAnimationFrame` and the browser will prefer to call new task set by `setTimeout` before any repainting. to get around this the user can add some milliseconds to force the browser to repaint before the callback. but it seems inconvenient for users.

This PR will delay the $nextTick after the transition show stage. Only in case of transitions.

Codepen before PR https://codepen.io/tomorrowsystem/pen/yLYQxNp
Codepen after PR https://codepen.io/tomorrowsystem/pen/LYpXJpo

### DIFF
--- a/src/component.js
+++ b/src/component.js
@@ -180,8 +180,12 @@ export default class Component {
             el.__x = new Component(el)
         })
 
-        if (! this.executeAndClearRemainingShowDirectiveStack()) {
-            this.executeAndClearNextTickStack(rootEl)
+        // In presence of x-show directives walk through the $nextTick from inside the directive handler
+        // otherwise proceed with $nextTick
+        if (this.showDirectiveStack.length > 0) {
+            this.executeAndClearRemainingShowDirectiveStack();
+        } else {
+            this.executeAndClearNextTickStack(rootEl);
         }
     }
 
@@ -196,7 +200,6 @@ export default class Component {
     }
 
     executeAndClearRemainingShowDirectiveStack() {
-        if (this.showDirectiveStack.length === 0) return false
         // The goal here is to start all the x-show transitions
         // and build a nested promise chain so that elements
         // only hide when the children are finished hiding.
@@ -215,7 +218,6 @@ export default class Component {
         // We've processed the handler stack. let's clear it.
         this.showDirectiveStack = []
         this.showDirectiveLastElement = undefined
-        return true
     }
 
     updateElement(el, extraVars) {
--- a/src/component.js
+++ b/src/component.js
@@ -180,9 +180,9 @@ export default class Component {
             el.__x = new Component(el)
         })
 
-        this.executeAndClearRemainingShowDirectiveStack()
-
-        this.executeAndClearNextTickStack(rootEl)
+        if (! this.executeAndClearRemainingShowDirectiveStack()) {
+            this.executeAndClearNextTickStack(rootEl)
+        }
     }
 
     executeAndClearNextTickStack(el) {
@@ -196,6 +196,7 @@ export default class Component {
     }
 
     executeAndClearRemainingShowDirectiveStack() {
+        if (this.showDirectiveStack.length === 0) return false
         // The goal here is to start all the x-show transitions
         // and build a nested promise chain so that elements
         // only hide when the children are finished hiding.
@@ -214,6 +215,7 @@ export default class Component {
         // We've processed the handler stack. let's clear it.
         this.showDirectiveStack = []
         this.showDirectiveLastElement = undefined
+        return true
     }
 
     updateElement(el, extraVars) {
--- a/dist/alpine-ie11.js
+++ b/dist/alpine-ie11.js
@@ -6741,9 +6741,12 @@
           _newArrowCheck(this, _this8);
 
           el.__x = new Component(el);
-        }.bind(this));
+        }.bind(this)); // In presence of x-show directives walk through the $nextTick from inside the directive handler
+        // otherwise proceed with $nextTick
 
-        if (!this.executeAndClearRemainingShowDirectiveStack()) {
+        if (this.showDirectiveStack.length > 0) {
+          this.executeAndClearRemainingShowDirectiveStack();
+        } else {
           this.executeAndClearNextTickStack(rootEl);
         }
       }
@@ -6763,10 +6766,9 @@
       value: function executeAndClearRemainingShowDirectiveStack() {
         var _this9 = this;
 
-        if (this.showDirectiveStack.length === 0) return false; // The goal here is to start all the x-show transitions
+        // The goal here is to start all the x-show transitions
         // and build a nested promise chain so that elements
         // only hide when the children are finished hiding.
-
         this.showDirectiveStack.reverse().map(function (thing) {
           var _this10 = this;
 
@@ -6805,7 +6807,6 @@
 
         this.showDirectiveStack = [];
         this.showDirectiveLastElement = undefined;
-        return true;
       }
     }, {
       key: "updateElement",
--- a/dist/alpine-ie11.js
+++ b/dist/alpine-ie11.js
@@ -6107,25 +6107,31 @@
     }
 
     var tick = function tick() {
+      var _this2 = this;
+
       _newArrowCheck(this, _this);
 
-      component.executeAndClearNextTickStack(component.$el);
+      setTimeout(function () {
+        _newArrowCheck(this, _this2);
+
+        return component.executeAndClearNextTickStack(component.$el);
+      }.bind(this));
     }.bind(this);
 
     var handle = function handle(resolve) {
-      var _this2 = this;
+      var _this3 = this;
 
       _newArrowCheck(this, _this);
 
       if (!value) {
         if (el.style.display !== 'none') {
           transitionOut(el, function () {
-            var _this3 = this;
+            var _this4 = this;
 
-            _newArrowCheck(this, _this2);
+            _newArrowCheck(this, _this3);
 
             resolve(function () {
-              _newArrowCheck(this, _this3);
+              _newArrowCheck(this, _this4);
 
               hide();
               tick();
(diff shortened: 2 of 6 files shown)


src/component.js
```diff
@@ -196,6 +196,7 @@ export default class Component {
     }
 
     executeAndClearRemainingShowDirectiveStack() {
+        if (this.showDirectiveStack.length === 0) return false
```
HugoDF, line 199: What's the `true/false` output supposed to represent? Whether there were any `show directives` to run?

In that case I would suggest creating a separate function to check that:
`hasAnyShowDirectivesToRun(){ return this.showDirectiveStack.length > 0 }`

SimoTod:
Thanks @ahmedkandel but I don't think it covers all the cases. This PR assumes that only a show directives is present in your component. As soon as you add another x-show, it won't work anymore since the stack will be emptied by the first directive and before the second one.

SimoTod:
See https://codepen.io/SimoTod/pen/gOaQyMr

src/component.js
ahmedkandel: What about directly checking the `showDirectiveStack` length in `updateElements` something like
```
if (this.showDirectiveStack.length > 0) {
    this.executeAndClearRemainingShowDirectiveStack();
} else {
    this.executeAndClearNextTickStack(rootEl);
}
```

ahmedkandel:
Thanks @SimoTod, Ya also if we have nested x-show with transitions. I suppose it will be better to use a setTimeout in this custom `tick()` functionality to make sure all show directives and transitions are done like.
```
const tick = () => {
    setTimeout(() => component.executeAndClearNextTickStack(component.$el));
}
```
It works but the downside here is it will be useful on the first x-show but will be repeated on the other x-show. Maybe we do some debouncing here, what do you suggest?

SimoTod:
It makes the code a bit more complicated and harder to read/understand/maintain.
I would personally defer executeAndClearNextTickStack so it runs after the next frame rather than setting multiple timeouts. Not sure about using the debounce function because it would conflict with other debounced functions, I think.
Any solution you pick, I would add comments on the code to help Caleb figuring out what you want to achieve. (+ a test to cover it if possible).

ahmedkandel:
Did some trials with deferring `executeAndClearNextTickStack` after the next frame but it has an impact on all transition tests as they rely on frameStack. So in order to reduce the changes, I think it is better to leave the transitions as is. Same thing on debounce functionality.

The setTimeout to 0 without any arbitrary buffer does the job but not elegant enough. I will check out for a better solution and add some testing. Thanks @SimoTod @HugoDF for your support.

ryangjchandler:
I'm not a big fan of these changes, to be honest. I think @SimoTod made a good point about missing edge cases - the fact that changing code is causing existing tests to fail (but fixes the original problem mentioned) is normally a telltale sign that breaking changes are being introduced.

Would we really consider the "problem" being fixed a bug?

ahmedkandel:
I don't agree with you @ryangjchandler this doesn't cause any tests to fail neither introduce any breaking changes. May be my comment regarding transition test was not clear enough. We are trying to save the fact that $nextTick as described should run after the DOM is ready. Please provide suggestions for better implementation.

ryangjchandler:
> I don't agree with you @ryangjchandler this doesn't cause any tests to fail neither introduce any breaking changes. May be my comment regarding transition test was not clear enough. We are trying to save the fact that $nextTick as described should run after the DOM is ready. Please provide suggestions for better implementation.

Based on:

> but it has an impact on all transition tests as they rely on frameStack

It sounds like a breaking change has been introduced, causing existing and passing tests to fail.

I personally haven't looked into this problem too much, nor have I actually experience it since I don't use the transition API and don't have enough time.

ahmedkandel:
Please read the comment again:

> Did some trials with deferring

I was referring to other solutions that I was trying.

SimoTod:
To be fair, I did play a bit with that last night so I have an implementation in mind and I'm a bit biased. 😂 It's not too different nor better than yours (but I'm using requestAnimationFrame) so I don't want to impose my view. I can give you the unit test if you like, though.

ahmedkandel:
Sure @SimoTod please go ahead :smiley:

SimoTod:
This would be the requestAnimationFrame implementation (https://github.com/alpinejs/alpine/compare/master...SimoTod:bug/nexttick-with-transitions).
You can grab the tests from there.
(Note, the transition test updates are just to make it more readable and to use sensible timeouts rather then random ones, you can ignore it, I'll send a separate PR for that).

ahmedkandel:
Good job @SimoTod , in my trials I did the same thing but without ` && this.nextTickStack.length > 0` So the transition tests were failing as it adds another callback to the requestAnimationFrame. While with your implementation it doesn't fail as it skips the callback with this condition.

It is strange even with your implementation if you add a $nextTick in any of the transition tests it fails e.g.
in 'transition in' test change the button click From:
```
<button x-on:click="show = ! show"></button>
```
To
```
<button x-on:click="show = ! show; $nextTick(()=>console.log('tick'))"></button>
```
It will fail, that is why i went with the solution to insert the callback inside the current transition callback.

Please check. Maybe we still missing something.

Thanks for the test.

ahmedkandel:
In the current transition tests, we spy on the animation frames and check them one by one for changes in element classes. When $nextTick introduce a new callback to the animation frames this will unshift frames that cause the test to fail.

I am not sure if it is better to rewrite the tests that use animation frames with a different logic or leave the animation frames reserved for transitions only. But I still not feeling comfortable to skip it `with && this.nextTickStack.length > 0 ` so the test will pass.

@SimoTod  what do you think?

SimoTod:
Hi @ahmedkandel 

The failure would be expected. The transition test suite mocks the requestAnimationFrame. This allows us to run each frame manually and test the classes otherwise it would be too fast. If we introduce a new frame call, the test fails because we need to call the framestack.pop an additional time for that frame.
Those tests does not need changing, in my opinion, they do what they are supposed to do and they are not meant to test nexttick which is covered by a different test suite.

Another consideration,`this.nextTickStack.length > 0` isn't there only for those tests but mostly because it's pointless to set a timeout/requestAnimationFrame if the stack is empty, it's just a micro optimisation, it should work without `if` if we use setTimeout in that function as well.

I believe the 2 implementations are quite similar, (setTimeout 0 and requestAnimationFrame have pretty much the same effect, deferring the call until the main execution pauses the next time) the only difference is the code portion when we call them (and the number of timeout set) so I'm not suggesting you to change the implementation if you prefer to be explicit in your code about the point when you run the stack. Your latest change with timeout fixes the only bug I could find and was the only blocker before Caleb's review for me.

We should add a test, though, since it's quite an important bugfix. I sent you the codediff because you can use the same tests in nexttick.spec.js and index.html.
Thanks for looking into it.

ahmedkandel:
Thanks @SimoTod, your implementation is very elegant, I have also tried to use `setTimeout 0` instead of `requestAnimationFrame` inside the `nextTickStack`, But the result depends a lot on the browser whether to execute the new task before repainting or after in no standard order.

I prefer to be explicit and move the nextTick to a new task with setTimeout from inside the requestAnimationFrame.

I will commit the test you provided. Thank you for your support :thumbsup:

calebporzio:
Thanks for this PR.

Here are my thoughts:

1. $nextTick shouldn't be held up until the transition finishes, it should be the very next "tick" in the browser (according to Alpine). The current behavior STILL is bad because it should wait until the transition STARTS, but currently it doesn't.

2. To perform some action after a transition finishes, we should dispatch events that users can hook into to listen for transition related things. One challenge with this is what to dispatch in the case of nested transitions.

Anyhow, I'm open to more discussions on  this, but I don't think this is the best fix ATM (it wouldn't take much to make it work properly though, just want to get the conversation started.)

HugoDF:
@calebporzio I think @ryangjchandler opened an issue a while ago about what events to expose maybe you could leave some thoughts there? 

https://github.com/alpinejs/alpine/issues/228

ahmedkandel:
@calebporzio totally agree with you that `$nextTick` should fire just after the transition starts and don't wait for the transition end.

This PR does this exactly and doesn't wait for the transition end. Maybe the use of `setTimeout` makes it look a little hacky but it is to grant correct timing in case of nested transitions.

Another implementation that looks more elegant proposed by @SimoTod was to wrap the walk through the $nextTick stack inside `requestAnimationFrame` callback as follows:

```javascript
executeAndClearNextTickStack(el) {
    // Skip spawns from alpine directives
    if (el === this.$el && this.nextTickStack.length > 0) {
        requestAnimationFrame(() => {
            // Walk through the $nextTick stack and clear it as we go.
            while (this.nextTickStack.length > 0) {
                this.nextTickStack.shift()()
            }
        })
    }
}
```

muzafferdede:
Just want to confirm, isn't this already possible by doing like so

```html
...

<div x-data="{whatsup: 'nothin'}" x-text="whatsup" 
    x-on:alpine-transitionrun.window="whatsup = 'transition run'" 
    x-on:alpine-transitionend.window="whatsup = 'transition end'">
</div>
<div x-data="{ show: false }">
  <button x-on:click="show = !show">Toggle</button>
  <div x-show.transition="show"
       x-on:transitionrun="$dispatch('alpine-transitionrun')" 
       x-on:transitionstart="$dispatch('alpine-transitionstart')" 
       x-on:transitionend="$dispatch('alpine-transitionend')">
    Content
  </div>
</div>
```

SimoTod:
@muzafferdede it's possible but you don't have as a userland workaround but you don't have the same level of control: transitionend will fire at the same why for show/hide or if an external js library changes the style of your component, while people would like to use nextTick in the same place where they have their click event changing the show variable, for example. At the moment, it just looks a bit broken

SimoTod:
@ahmedkandel We are probably going with the other solution since changes are localised in a single place. Thanks for helping debugging the issue though.

calebporzio:
Thanks @ahmedkandel for the contribution! Like @SimoTod said, we're going to go with the other solution.

ahmedkandel:
No worries #555 by @SimoTod is better :thumbsup:

PR #505, opened 2020-05-19
Rename `result` variable inside of eval utils

### LINKED ISSUE #503: The result property inside the data object causes an error
Hi guys!
I get an error `TypeError: Cannot read property 'length' of null` in the console when I try to use the `result` property inside the data object. Created an example on [codepen](https://codepen.io/k-ivan/pen/JjYegOQ?editors=1010).

ryangjchandler:
Closes #503.

This PR renames the `result` variable used inside of `saferEval` to `__alpine_result` to prevent conflicts with any `result` variables defined on the component data.

### DIFF
--- a/dist/alpine-ie11.js
+++ b/dist/alpine-ie11.js
@@ -5379,7 +5379,7 @@
   }
   function saferEval(expression, dataContext) {
     var additionalHelperVariables = arguments.length > 2 && arguments[2] !== undefined ? arguments[2] : {};
-    return new Function(['$data'].concat(_toConsumableArray(Object.keys(additionalHelperVariables))), "var result; with($data) { result = ".concat(expression, " }; return result")).apply(void 0, [dataContext].concat(_toConsumableArray(Object.values(additionalHelperVariables))));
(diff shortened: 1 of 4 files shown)


HugoDF:
@ryangjchandler shall we add a test that we can use a variable called `result`?

ryangjchandler:
> @ryangjchandler shall we add a test that we can use a variable called `result`?

Yeah probably makes sense.

SimoTod:
I don't think it's needed, guys. We can't test every single word in the dictionary, after Ryan's change `result` is just a common word like many others. The fact that existing tests pass is the perfect indicator that nothing broke.

ryangjchandler:
> I don't think it's needed, guys. We can't test every single word in the dictionary, after Ryan's change `result` is just a common word like many others. The fact that existing tests pass is the perfect indicator that nothing broke.

Yeah, it's a fair point. If we'd changed the logic but get the variable name the same, it would have made sense to add a test.

HugoDF:
Agree, maybe overzealous, although technically we're fixing a bug where `result` can't be used 🤷🏼‍♀️ I'm happy either way.

calebporzio:
Makes sense, thanks Ryan!

PR #506, opened 2020-05-19
Add support for `data-x` attributes

### LINKED ISSUE #464: Alternative binding  for directives?
Hey,

I wonder if it is possible to specify a custom directive (e.g data-x-data instead of x-data). I am having trouble of implementing Alpinejs in Salesforce lightning framework.

ryangjchandler:
Closes #464 alongside previous issues and discussions regarding Alpine's "non-compliant" directives.

All existing tests are passing. I'm not sure whether we should add tests for each directives `data-x` counterpart, maybe just a single test to ensure they behave the same at the most basic level.

In terms of performance, the overhead will be minimal but this will definitely help developers who using Alpine who need to make sure their HTML attributes are compliant.

### DIFF
--- a/src/component.js
+++ b/src/component.js
@@ -415,6 +415,8 @@ export default class Component {
                 self.walkAndSkipNestedComponents(self.$el, el => {
                     if (el.hasAttribute('x-ref') && el.getAttribute('x-ref') === property) {
                         ref = el
+                    } else if (el.hasAttribute('data-x-ref') && el.getAttribute('data-x-ref') === property) {
+                        ref = el
                     }
                 })
 
--- a/src/component.js
+++ b/src/component.js
@@ -13,9 +13,19 @@ export default class Component {
     constructor(el, seedDataForCloning = null) {
         this.$el = el
 
-        const dataAttr = this.$el.getAttribute('x-data')
+        let dataAttr = this.$el.getAttribute('x-data')
+
+        if (! dataAttr && this.$el.dataset.xData) {
+            dataAttr = this.$el.dataset.xData
+        }
+
         const dataExpression = dataAttr === '' ? '{}' : dataAttr
-        const initExpression = this.$el.getAttribute('x-init')
+
+        let initExpression = this.$el.getAttribute('x-init')
+
+        if (! initExpression && this.$el.dataset.xInit) {
+            initExpression = this.$el.dataset.xInit
+        }
 
         this.unobservedData = seedDataForCloning ? seedDataForCloning : saferEval(dataExpression, {})
 
@@ -126,7 +136,7 @@ export default class Component {
     walkAndSkipNestedComponents(el, callback, initializeComponentCallback = () => {}) {
         walk(el, el => {
             // We've hit a component.
-            if (el.hasAttribute('x-data')) {
+            if (el.hasAttribute('x-data') || el.dataset.xData) {
                 // If it's not the current one.
                 if (! el.isSameNode(this.$el)) {
                     // Initialize it if it's not.
@@ -335,12 +345,13 @@ export default class Component {
         const observer = new MutationObserver((mutations) => {
             for (let i=0; i < mutations.length; i++) {
                 // Filter out mutations triggered from child components.
-                const closestParentComponent = mutations[i].target.closest('[x-data]')
+                const closestParentComponent = mutations[i].target.closest('[x-data], [data-x-data]')
 
                 if (! (closestParentComponent && closestParentComponent.isSameNode(this.$el))) continue
 
-                if (mutations[i].type === 'attributes' && mutations[i].attributeName === 'x-data') {
-                    const rawData = saferEval(mutations[i].target.getAttribute('x-data'), {})
+                if (mutations[i].type === 'attributes' && (mutations[i].attributeName === 'x-data' || mutations[i].attributeName === 'data-x-data')) {
+                    const dataExpression = mutations[i].attributeName === 'x-data' ? mutations[i].target.getAttribute('x-data') : mutations[i].target.dataset.xData
+                    const rawData = saferEval(dataExpression, {})
 
                     Object.keys(rawData).forEach(key => {
                         if (this.$data[key] !== rawData[key]) {
@@ -353,7 +364,7 @@ export default class Component {
                     mutations[i].addedNodes.forEach(node => {
                         if (node.nodeType !== 1 || node.__x_inserted_me) return
 
-                        if (node.matches('[x-data]')) {
+                        if (node.matches('[x-data], [data-x-data]')) {
                             node.__x = new Component(node)
                             return
                         }
--- a/src/component.js
+++ b/src/component.js
@@ -15,16 +15,16 @@ export default class Component {
 
         let dataAttr = this.$el.getAttribute('x-data')
 
-        if (! dataAttr && this.$el.dataset.xData) {
-            dataAttr = this.$el.dataset.xData
+        if (! dataAttr && this.$el.getAttribute('data-x-data')) {
+            dataAttr = this.$el.getAttribute('data-x-data')
         }
 
         const dataExpression = dataAttr === '' ? '{}' : dataAttr
 
         let initExpression = this.$el.getAttribute('x-init')
 
-        if (! initExpression && this.$el.dataset.xInit) {
-            initExpression = this.$el.dataset.xInit
+        if (! initExpression && this.$el.getAttribute('data-x-init')) {
+            initExpression = this.$el.getAttribute('data-x-init')
         }
 
         this.unobservedData = seedDataForCloning ? seedDataForCloning : saferEval(dataExpression, {})
@@ -302,6 +302,7 @@ export default class Component {
 
                 case 'cloak':
                     el.removeAttribute('x-cloak')
+                    el.removeAttribute('data-x-cloak')
                     break;
 
                 default:
@@ -393,6 +394,8 @@ export default class Component {
             this.walkAndSkipNestedComponents(self.$el, el => {
                 if (el.hasAttribute('x-ref')) {
                     refObj[el.getAttribute('x-ref')] = true
+                } else if (el.hasAttribute('data-x-ref')) {
+                    refObj[el.getAttribute('data-x-ref')] = true
                 }
             })
         /* IE11-ONLY:END */
--- a/src/utils.js
+++ b/src/utils.js
@@ -73,7 +73,7 @@ export function saferEvalNoReturn(expression, dataContext, additionalHelperVaria
     )
 }
 
-const xAttrRE = /^x-(on|bind|data|text|html|model|if|for|show|cloak|transition|ref)\b/
+const xAttrRE = /^(data-x|x)-(on|bind|data|text|html|model|if|for|show|cloak|transition|ref)\b/
 
 export function isXAttr(attr) {
     const name = replaceAtAndColonWithStandardSyntax(attr.name)
@@ -91,7 +91,7 @@ export function getXAttrs(el, type) {
             const modifiers = name.match(/\.[^.\]]+(?=[^\]]*$)/g) || []
 
             return {
-                type: typeMatch ? typeMatch[1] : null,
+                type: typeMatch ? typeMatch[2] : null,
                 value: valueMatch ? valueMatch[1] : null,
                 modifiers: modifiers.map(i => i.replace('.', '')),
                 expression: attr.value,
--- a/dist/alpine-ie11.js
+++ b/dist/alpine-ie11.js
@@ -5396,7 +5396,7 @@
 
     return new Function(['dataContext'].concat(_toConsumableArray(Object.keys(additionalHelperVariables))), "with(dataContext) { ".concat(expression, " }")).apply(void 0, [dataContext].concat(_toConsumableArray(Object.values(additionalHelperVariables))));
   }
-  var xAttrRE = /^x-(on|bind|data|text|html|model|if|for|show|cloak|transition|ref)\b/;
+  var xAttrRE = /^(data-x|x)-(on|bind|data|text|html|model|if|for|show|cloak|transition|ref)\b/;
   function isXAttr(attr) {
     var name = replaceAtAndColonWithStandardSyntax(attr.name);
     return xAttrRE.test(name);
@@ -5414,7 +5414,7 @@
       var valueMatch = name.match(/:([a-zA-Z\-:]+)/);
       var modifiers = name.match(/\.[^.\]]+(?=[^\]]*$)/g) || [];
       return {
-        type: typeMatch ? typeMatch[1] : null,
+        type: typeMatch ? typeMatch[2] : null,
         value: valueMatch ? valueMatch[1] : null,
         modifiers: modifiers.map(function (i) {
           _newArrowCheck(this, _this3);
--- a/dist/alpine-ie11.js
+++ b/dist/alpine-ie11.js
@@ -6520,8 +6520,18 @@
 
       this.$el = el;
       var dataAttr = this.$el.getAttribute('x-data');
+
+      if (!dataAttr && this.$el.getAttribute('data-x-data')) {
+        dataAttr = this.$el.getAttribute('data-x-data');
+      }
+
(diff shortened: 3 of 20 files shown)


SimoTod:
It's a good start but I believe it's not enough. There are a bunch of querySelector('[x-data]'), getAttribute('x-data') and getAttribute('x-init') to update as well.
We also need at least 1 test to check that
```html
<div data-x-data="{foo: ''}" data-x-init="foo = 'bar'">
    <span data-x-text="foo"></span>
    <button data-x-on:click="foo = 'bob'"></button>
</div>
```
Bonus if we can add tests for the other bits where we are changing the code.

HugoDF:
Nice one @ryangjchandler 
I think you can take inspiration from the shorthand tests, just a sanity check, literally if you use data-x-data with an x-text will it work

ryangjchandler:
Didn't realise I made it ready to review, it was supposed to be in review still lol.

ryangjchandler:
It's actually ready to review now lol - pretty sure I've got all of the `x-data` spots and added support for `data-x-data`.

Regular `x-` attributes (`x-data`, `x-init`, `x-ref`) attributes will be checked first, then fall back to the `data-x` attributes if they exist.

I also chose to use `getAttribute()` instead of `dataset` for consistency.

src/component.js
```diff
@@ -13,9 +13,19 @@ export default class Component {
     constructor(el, seedDataForCloning = null) {
         this.$el = el
 
-        const dataAttr = this.$el.getAttribute('x-data')
+        let dataAttr = this.$el.getAttribute('x-data')
+
+        if (! dataAttr && this.$el.getAttribute('data-x-data')) {
+            dataAttr = this.$el.getAttribute('data-x-data')
+        }
```
SimoTod, line 20: We can also use a one liner
`const dataAttr = this.$el.getAttribute('x-data') || this.$el.getAttribute('data-x-data')`

src/component.js
```diff
@@ -13,9 +13,19 @@ export default class Component {
...
+
         const dataExpression = dataAttr === '' ? '{}' : dataAttr
-        const initExpression = this.$el.getAttribute('x-init')
+
+        let initExpression = this.$el.getAttribute('x-init')
+
+        if (! initExpression && this.$el.getAttribute('data-x-init')) {
+            initExpression = this.$el.getAttribute('data-x-init')
+        }
```
SimoTod, line 28: Same
`const initExpression = this.$el.getAttribute('x-init') || this.$el.getAttribute('data-x-init')`

src/utils.js
```diff
@@ -73,7 +73,7 @@ export function saferEvalNoReturn(expression, dataContext, additionalHelperVaria
     )
 }
 
-const xAttrRE = /^x-(on|bind|data|text|html|model|if|for|show|cloak|transition|ref)\b/
+const xAttrRE = /^(data-x|x)-(on|bind|data|text|html|model|if|for|show|cloak|transition|ref)\b/
```
SimoTod, line 76: Not a blocker but I thought it was worth sharing:
we can use a non-capturing block in this regex
`const xAttrRE = /^(?:data-x|x)-(on|bind|data|text|html|model|if|for|show|cloak|transition|ref)\b/`
and keep using `typeMatch[1]` later on

ryangjchandler: Yeah good spot. This regex was the first one I tried so I just kept it as is, might be worth changing though.

src/component.js
ryangjchandler: I did use OR syntax originally but I went for the more verbose and clear equivalent. Happy to change it though if you think it's better :)

src/utils.js
HugoDF: Ooo lookbehind right? Will it get transpiled (correctly)?

src/component.js
HugoDF: I agree with changing to `||`s, it'll keep it more compact and the 2 `getAttribute` calls are " the same" (represent the same operation).

src/component.js
```diff
@@ -126,7 +136,7 @@ export default class Component {
     walkAndSkipNestedComponents(el, callback, initializeComponentCallback = () => {}) {
         walk(el, el => {
             // We've hit a component.
-            if (el.hasAttribute('x-data')) {
+            if (el.hasAttribute('x-data') || el.dataset.xData) {
```
HugoDF, line 139: 🤔 wasn't the idea to avoid using `dataset`? `el.hasAttribute('data-x-data')`?

src/component.js
```diff
@@ -335,12 +346,13 @@ export default class Component {
...
             for (let i=0; i < mutations.length; i++) {
                 // Filter out mutations triggered from child components.
-                const closestParentComponent = mutations[i].target.closest('[x-data]')
+                const closestParentComponent = mutations[i].target.closest('[x-data], [data-x-data]')
 
                 if (! (closestParentComponent && closestParentComponent.isSameNode(this.$el))) continue
 
-                if (mutations[i].type === 'attributes' && mutations[i].attributeName === 'x-data') {
-                    const rawData = saferEval(mutations[i].target.getAttribute('x-data'), {})
+                if (mutations[i].type === 'attributes' && (mutations[i].attributeName === 'x-data' || mutations[i].attributeName === 'data-x-data')) {
```
HugoDF, line 353: One way to clean this up would be 

```suggestion
                if (mutations[i].type === 'attributes' && ['x-data', 'data-x-data'].includes(mutations[i].attributeName)) {
```

src/component.js
```diff
@@ -335,12 +346,13 @@ export default class Component {
...
+                    const dataExpression = mutations[i].attributeName === 'x-data' ? mutations[i].target.getAttribute('x-data') : mutations[i].target.dataset.xData
```
HugoDF, line 354:
```suggestion
                    const dataExpression = mutations[i].target.getAttribute('x-data') || mutations[i].target.getAttribute('data-x-data')
```
?

HugoDF: Might be worth extracting a util

```js
function getXData(el) {
  return el.getAttribute('x-data') || el.getAttribute('data-x-data')
}
```

src/component.js
```diff
@@ -382,6 +394,8 @@ export default class Component {
             this.walkAndSkipNestedComponents(self.$el, el => {
                 if (el.hasAttribute('x-ref')) {
                     refObj[el.getAttribute('x-ref')] = true
+                } else if (el.hasAttribute('data-x-ref')) {
+                    refObj[el.getAttribute('data-x-ref')] = true
```
HugoDF, line 398:
```suggestion
                if (el.hasAttribute('x-ref') || el.hasAttribute('data-x-ref')) {
                    refObj[el.getAttribute('x-ref') || el.getAttribute('data-x-ref')] = true
                }
```
?

HugoDF: again might be worth making some utils

```js
function hasXRef(el) {
  return el.hasAttribute('x-ref') || el.hasAttribute('data-x-ref')
}
function getXRef(el) {
  return el.getAttribute('x-ref') || el.getAttribute('data-x-ref')
}
```

src/utils.js
SimoTod: It's slightly different. I think lookbehind is not supported everywhere but non-capturing groups are. It matches the group but it doesn't add a record in the resulting array.

src/component.js
SimoTod: Yeah, if you don't mind changing it, we'll save a few lines. 👍

src/utils.js
HugoDF: Ah yes I see there's a one char difference 😄 Regex are always a fun one

calebporzio:
Is this actually needed? Or just to appease valid-html purists.

I appreciate the work, but these are the things that bloat the code and don't add any tangible value to anyone that I'm aware of.

ryangjchandler:
@calebporzio Yeah, I mean it's just to please developers who want to use Alpine but need to follow some strict guidelines on compliance. 

Now that I think about it, there's more to it than just the attribute names. 

Happy to close this though!

tomhodgins:
> Is this actually needed? Or just to appease valid-html purists.
> I appreciate the work, but these are the things that bloat the code and don't add any tangible value to anyone that I'm aware of.

Valid HTML is also important for:

- tooling
- editors
- web browsers

And if you camp on platform namespaces it actually blocks HTML from introducing any similarly-named things. We can't have a `<modal>` element in HTML because too many people went ahead and used invalid `<modal>` tags so now introducing one would break existing (bad) sites so we get `<dialog>` instead. In JavaScript they just went through this with MooTools dangerously extending `Array.flatten()` so we ended up with `Array.flat()` instead (this was known as SmooshGate). Please don't do this with HTML attributes by tying the hands of what the WHATWG is able to introduce in the future.

If you want user-defined attributes in HTML you have 2 options:

- [data-* attributes](https://html.spec.whatwg.org/multipage/dom.html#embedding-custom-non-visible-data-with-the-data-*-attributes)
- [Custom Elements]()

Either of those let you define the attributes you want to use in a way that's future-proof against anywhere HTML may grow in the future, and it doesn't block or prevent HTML from using anything it might plan or want to use.

For using `data-*` attributes, you're allowed to invent any attribute name on any element so long as it begins with `data-`, and there's the special `dataset` interface in the web APIs for accessing these values in a convenient way.

The HTML spec even has a special note just for library authors:

> JavaScript libraries may use the custom data attributes, as they are considered to be part of the page on which they are used. Authors of libraries that are reused by many authors are encouraged to include their name in the attribute names, to reduce the risk of clashes. Where it makes sense, library authors are also encouraged to make the exact name used in the attribute names customizable, so that libraries whose authors unknowingly picked the same name can be used on the same page, and so that multiple versions of a particular library can be used on the same page even when those versions are not mutually compatible.

To paraphrase that, it means `data-library-thing` is better than `data-thing`, and `data-custom-thing` where the library user gets to choose what `custom` says is the best option.

The other alternative if you want to invent custom attributes is Custom Elements:

> Any namespace-less attribute that is relevant to the element's functioning, as determined by the element's author, may be specified on an autonomous custom element, so long as the attribute name is XML-compatible and contains no ASCII upper alphas. The exception is the is attribute, which must not be specified on an autonomous custom element (and which will have no effect if it is).
> 
> Customized built-in elements follow the normal requirements for attributes, based on the elements they extend. To add custom attribute-based behavior, use data-* attributes.

Using non-standard attributes that get processed away before the HTML hits the DOM is less-bad, but you still have the issue of camping on the platform namespace and not necessarily being able to distinguish between your off-label usage and a potential future feature named the same way. It would be much better to make things future proof and conflict-proof to follow the advice of the HTML spec!

szabeszg:
Hello, I am new to Alpine, thanks for the devs for all their work on it! However, I am surprised to read that being able to produce valid HTML is even a question, on the contrary: it is a must as @tomhodgins explained above. Please do no abuse HTML.

HugoDF:
Thanks all for your input

Having spoken with @calebporzio, I created another PR https://github.com/alpinejs/alpine/pull/516 which creates a special build using `data-alpine-*` attributes. It causes less overhead in the code and I don't think people will mix `data-alpine-*` and `x-*` attributes.

Natetronn:
Has this been implemented? I'd like to see this as well.

HugoDF:
> Has this been implemented? I'd like to see this as well.

This feature was held off for now, see (& feel free to contribute)  https://github.com/alpinejs/alpine/issues/464#issuecomment-632353183

Natetronn:
Thanks @HugoDF 

I don't have a use case like the other post is suggesting rather, trying to validate the HTML. I'd like to eliminate the warning and errors while running a site through the W3C Validater. And this for the reason @tomhodgins has suggested and others too.

For example, part of the commitment I often make to clients is to write valid HTML and then show them the results when their site launches, if need be, per the contract. It can be a selling point and just the "right" thing to do, either way. 

Also, for portfolios, people might like to point out a piece of work that's valid HTML and this to help them land a job. Employers often ask to see 3 or 4 websites a developer has worked on, then they run them said sites through the gauntlet of tests, among other things.

For example, here's a work in progress site that's passing all HTML validation, minus all Alpine JS based attributes:

https://validator.w3.org/nu/?doc=https%3A%2F%2Faggressivecarwraps.com%2F

There are at least five issues with this particular site but, I'm sure there would be more had it been a more advanced use of Alpine or a larger site:

- @keydown.escape
- :class
- @click (@click.away etc.)
- x-data
- x-show

Point being, I'm hoping for full data-* attribute prefixing support.

I can post to that thread as well, if you think I should?

HugoDF:
@Natetronn in that case you should be able to use this PR to generate a custom Alpine build https://github.com/alpinejs/alpine/pull/516

Natetronn:
Thanks @HugoDF! I'll take a look.

FWIW, I was talking with https://htmx.org, as I was wondering if they support data-* and they do. I think HTMX and AlpineJS are two peas in a pod i.e. the both can live side by side very nicely and this for contexts where one doesn't need a full blown framework like Vue, React, Angular etc. It was the lack of data-* that sparked the question in the first place.

https://twitter.com/htmx_org/status/1308789273877864448

Anyway, thanks for AlpineJS, nonetheless! It's really cool.

HugoDF:
> Thanks @HugoDF! I'll take a look.
> 
> FWIW, I was talking with https://htmx.org, as I was wondering if [they support data-* and they do](https://github.com/bigskysoftware/htmx/blob/bd94f7a41fd87b59671aa706d6d12fbb3895eda9/src/htmx.js#L89). I think HTMX and AlpineJS are two peas in a pod i.e. the both can live side by side very nicely and this for contexts where one doesn't need a full blown framework like Vue, React, Angular etc. (jQuery or write the Vanilla js themselves.) It was the lack of data-* that sparked the question in the first place:
> 
> https://twitter.com/htmx_org/status/1308789273877864448
> 
> Anyway, thanks for AlpineJS, nonetheless! It's really cool.

Yeah, like I said, you can run a custom Alpine.js build 🤷‍♀️

szabeszg:
> Yeah, like I said, you can run a custom Alpine.js build

Or we can build our own framework? Would you suggest that too? Is it so difficult to support standardization? Sorry for asking all these questions, but I cannot see the point of basically rejecting this most reasonbale request.

Natetronn:
@szabeszg I can appreciate your passion but, let's avoid biting the hand that feeds us here, yes?

We don't want to build our own framework and no that's not what they're suggesting. We want to use Alpine because, let's face it, it's awesome! Ultimately it's their decision and after a bit of pleading, if they don't want to implement it directly then that's their call to make. We'll just have to adjust accordingly.

HugoDF:
> > Yeah, like I said, you can run a custom Alpine.js build
> 
> Or we can build our own framework? Would you suggest that too? Is it so difficult to support standardization? Sorry for asking all these questions, but I cannot see the point of basically rejecting this most reasonbale request.

😄 I don't lean either way in this debate, we've proposed a couple of solutions but Caleb isn't convinced of the use case.

I'm just pointing out that there's a workaround (albeit quite heavy-handed)

KevinBatdorf:
My vote would be for being able to define a custom prefix to avoid conflicts with anything else using `x-`, but not a critical thing for me today.

No need to have a perfect w3 html validator report _in my opinion_.

lleger:
FWIW, from a real-world standpoint (i.e., not a "valid for valid sake" argument), Alpine simply won't work in some older browsers, e.g. Safari 11.1.2, because it refuses to parse nodes with the `x-*` attrs. That browser still falls within some support policies, e.g. "latest - 2", which is what bit us.

So for that reason it seems like adding support for `data-*` version is 👍 , especially in light of what @tomhodgins wrote, which is that the WHATWG discourages doing this very thing.

HugoDF:
The changes in this PR can be applied to master to create a custom compatibility build for that. 

To be fair as the creator of the PR, I don't see any reason not to have a compat build, it's all done at build time so has very little maintenance overhead apart from some docs 🤷‍♀️ maybe that's why Caleb wasn't keen on integrating this.

Natetronn:
Just ran into this comment and couldn't help be reminded of this thread. A Tailwind CSS based template, which was initially built using Alpine, had to be revert to jQuery instead, and this to meet the validity requirements of Theme Forest:

https://themeforest.net/item/tailstack-tailwind-admin-dashboard/28906006/comments?page=1&filter=all#comment_25759383

PR #513, opened 2020-05-21
replace .forEach with .some on x-bind checkbox

reed-jones:
A small change, `.some` will stop iterating after the first match, but can still keep the "includes fuzzy"

### DIFF
--- a/src/directives/bind.js
+++ b/src/directives/bind.js
@@ -23,14 +23,7 @@ export function handleAttributeBindingDirective(component, el, attrName, express
                 // I'm purposely not using Array.includes here because it's
                 // strict, and because of Numeric/String mis-casting, I
                 // want the "includes" to be "fuzzy".
-                let valueFound = false
-                value.forEach(val => {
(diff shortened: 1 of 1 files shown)


HugoDF:
Good one, this is interesting because since `.some` has to be polyfilled for IE11, it makes the IE11 bundle larger while making the modern bundle smaller 😂

ryangjchandler:
> Good one, this is interesting because since `.some` has to be polyfilled for IE11, it makes the IE11 bundle larger while making the modern bundle smaller 😂

`Array.prototype.some` is supported back to IE9?

HugoDF:
> > Good one, this is interesting because since `.some` has to be polyfilled for IE11, it makes the IE11 bundle larger while making the modern bundle smaller 😂
> 
> `Array.prototype.some` is supported back to IE9? Only needs to be polyfilled for typed arrays.

lol, thanks 😄

calebporzio:
Love this! Very nice!

PR #515, opened 2020-05-22
Add `x-bind`

calebporzio:
Here are a few things I'd like input on:

1) Are there performance implications for running "saferEval" inside "getXAttrs". In other words, is Alpine slower now that there is an eval run every time it checks the attributes of an element. (Maybe caching is one potential answer - but there be dragons)
2) Currently this only works well when the bound directives object has non arrow-function attributes, because `this` binding is funky  with arrow functions. Would love to know if we can force the `this` of arrow functions reliably.
3) Do transitions still work with the getXAttrs -> .expression stuff?? (Because expression is no longer a string (for parsing classes and what not)

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -61,7 +61,7 @@ export function saferEval(expression, dataContext, additionalHelperVariables = {
 
 export function saferEvalNoReturn(expression, dataContext, additionalHelperVariables = {}) {
     if (typeof expression === 'function') {
-        expression.call(dataContext)
+        return expression.call(dataContext)
     }
 
     // For the cases when users pass only a function reference to the caller: `x-on:click="foo"`
@@ -171,7 +171,7 @@ export function transitionIn(el, show, component, forceSkip = false) {
         transitionHelperIn(el, modifiers, show)
     // Otherwise, we can assume x-transition:enter.
     } else if (attrs.filter(attr => ['enter', 'enter-start', 'enter-end'].includes(attr.value)).length > 0) {
-        transitionClassesIn(el, attrs, show)
+        transitionClassesIn(el, component, attrs, show)
     } else {
     // If neither, just show that damn thing.
         show()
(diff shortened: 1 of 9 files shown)


src/utils.js
```diff
@@ -50,12 +50,20 @@ export function debounce(func, wait) {
...
+
     return (new Function(['$data', ...Object.keys(additionalHelperVariables)], `var __alpine_result; with($data) { __alpine_result = ${expression} }; return __alpine_result`))(
         dataContext, ...Object.values(additionalHelperVariables)
     )
 }
 
 export function saferEvalNoReturn(expression, dataContext, additionalHelperVariables = {}) {
+    if (typeof expression === 'function') {
+        expression.call(dataContext)
+    }
```
HugoDF, line 65: Shouldn't we still have an early return here? Do we need any of the rest of the function to run?

HugoDF:
Hey @calebporzio I've had another look (I wasn't 100% switched on during the livestream)

2. The ECMAScript spec allows for `.call(thisContext)` for arrow functions, eg. `(() => { console.log(this) }).call('hello')`  will log out `this` as per where the arrow function was defined, **not** `'hello'`.

From the spec (http://www.ecma-international.org/ecma-262/6.0/#sec-arrow-function-definitions-runtime-semantics-evaluation):
>  An ArrowFunction does not define local bindings for arguments, super, **this**, or new.target. Any reference to arguments, super, this, or new.target within an ArrowFunction must resolve to a binding in a lexically enclosing environment. 

Workarounds:
- explicitly disallow arrow functions for this use-case (when using `this` I'm not sure anyone should be using arrow functions anyways) - this would kind of need a way to detect `this`/arrow function and log a warning/error (seems doable with a regex maybe?).
- only evaluate (`saferEval`) the arrow function at runtime with a specific context (which is what we do everywhere else and why arrow functions work)
  - ie. safeEval the whole attribute ahead of time (what's currently happening), but `.toString` any functions back - I wasn't sure about differing browser implementations during the stream but at the very least it's supported way back in supported browsers and the behaviour from the spec seems to fit our needs see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/toString#Browser_compatibility

HugoDF:
I've made the tests green in this PR FYI @calebporzio https://github.com/alpinejs/alpine/pull/518

PR #516, opened 2020-05-22
Alpine.js HTML compat build with `data-alpine-*` attributes (instead of `x-*`)

### LINKED ISSUE #464: Alternative binding  for directives?

HugoDF:
Closes #464 

- new rollup config that replaces `x-` with `data-alpine`.
- new built JS output

### DIFF
--- a/dist/alpine-data-compat.js
+++ b/dist/alpine-data-compat.js
@@ -1,0 +1,1704 @@
+(function (global, factory) {
+  typeof exports === 'object' && typeof module !== 'undefined' ? module.exports = factory() :
+  typeof define === 'function' && define.amd ? define(factory) :
+  (global = global || self, global.Alpine = factory());
+}(this, (function () { 'use strict';
+
+  function _defineProperty(obj, key, value) {
+    if (key in obj) {
+      Object.defineProperty(obj, key, {
+        value: value,
+        enumerable: true,
+        configurable: true,
+        writable: true
+      });
+    } else {
+      obj[key] = value;
+    }
+
+    return obj;
+  }
+
+  function ownKeys(object, enumerableOnly) {
+    var keys = Object.keys(object);
+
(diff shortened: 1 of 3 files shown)


HugoDF:
Relevant comments from the other PR

https://github.com/alpinejs/alpine/pull/506#issuecomment-632836336
https://github.com/alpinejs/alpine/pull/506#issuecomment-634162527

calebporzio:
Thanks @HugoDF, maybe there was a miscommunication. I thought we settled on NOT addressing this.

Anyhow, thanks for the PR, it will be useful if we do eventually want to support data- attrs, but I think for now this isn't a real tangible issue for anyone that I'm aware of (with the exception of the LWC case, but I believe that wouldn't work anyways.)

Closing for now unless I'm missing something. Thanks!

HugoDF:
👍 happy with that too, this is trivial to revive later in any case

THS-on:
@calebporzio I want to use Alpine.js with [Pandoc](https://pandoc.org/) to add some Interactive elements to a automatically generated Website. Pandoc prefixes all user specified attributes with `data-` which makes it not easily possible to use Alpine.js without some post processing.

I'm currently testing this version and it works really well for my use case. It would be great to have official support for that.

Natetronn:
@calebporzio With all due respect, I think you may be missing something? Here's some of my thoughts on the subject:

https://github.com/alpinejs/alpine/pull/506#issuecomment-698430006 

And consider reading @tomhodgins thoughts a few replies above my own, if you haven't already.

@HugoDF pointed out the PR (thanks for that) but, please add my vote for this to be added officially.

Thanks you!

PR #518, opened 2020-05-23
new `x-bind` to object: fix transitions + tests

HugoDF:
2 issues:
1. missing `component` in call to internal transition function
2. invalid JS in the `object x-bind` test


calebporzio:
Thanks Hugo! Big help!

PR #521, opened 2020-05-24
Russian translation

evgeniyPP:
Hello. I like Alpine. Its simplicity and accessibility, along with all the modern features that it provides.

Alpine is quite popular in the Russian community, so I was slightly surprised that no one had translated its documentation yet.

Given its small size, I have decided I would do it myself. Besides, this will allow me to better understand the work of the library.

In short, here is the translation. I will be glad if you include it in your project. I am pleased with its quality, although I am sure that there is a lot to improve. Perhaps I will continue to improve it, or my compatriots will help me.


evgeniyPP:
A test has failed. I do not even change any code. I have only written some text...

SimoTod:
A few tests are based on javascript timeouts. Some times we get false positive when the server is slower than usual. 
If you push your changes again, it will re-run and probably pass. As you don't have new commits, you can just type `git push - f`

PR #536, opened 2020-05-29
x-show transition overlap fix

### LINKED ISSUE #533: Might be an Alpine Performance bug on `x-show` with transition
Tricky one, 

Screencast:
https://gofile.io/d/0bqt5t

In this example: 

muzafferdede:
Fixes #533 

> Not sure if this is a solid fix. It is working ~perfectly~ in my use case. I would like to test it other than manually via tests but also not sure how/what to test. I would like to know your thoughts. 

> I did a check right before alpine decides starting the transition. If there is an already transition going on the element, do not add new transition and just resolve, if there is no any transition, add show/hide transition depending on the value. 

> P.S
I don't know why but when i pull from master and build without any change, i saw some changes which is out of nowhere. So i am confused if the master is the build version. Since it has some changes not in compiled to `alpine.js`,  so i just included the files i changed without building it. I also notice @calebporzio merges source files only so probably all good. 

Instead of making a new PR for this issue I just force pushed the changes to this existing PR. 

I have to refactored transition to achieve this solution. So there is a `Refactored Transition` PR of this issue which has same solution + refactored transitions to let transitions has its own space so it will be easier to work on it. (it was really hard to work on existing transition code) 

This fix is quite self explanatory. I assigned current transitions to el as `el.__x_remaining_transitions` and make sure they are completed before next one starts. 

By this, quick clicks overlapping transitions and wrong state of display issues will be fixed. There is also a test called `remaining transitions forced to complete if they exists` which tests this bug. 

Let me know if there is any room to improve. Huge thanks @SimoTod and @HugoDF for their comments. .5 x more thanks to @SimoTod for bearing with me :).

### DIFF
--- a/dist/alpine-ie11.js
+++ b/dist/alpine-ie11.js
@@ -5923,7 +5923,7 @@
 
         stages.end(); // Asign current transition to el in case we need to force it
 
-        el.__x_remaning_transitions = function () {
+        el.__x_remaining_transitions = function () {
           _newArrowCheck(this, _this15);
 
           stages.hide(); // Adding an "isConnected" check, in case the callback
@@ -5934,15 +5934,15 @@
           } // Safe to remove transition from el since it is completed
 
 
-          delete el.__x_remaning_transitions;
+          delete el.__x_remaining_transitions;
         }.bind(this);
 
         setTimeout(function () {
           _newArrowCheck(this, _this15);
 
-          // We only want to run remaning transitions in the end if they exists
-          if (el.__x_remaning_transitions) {
-            el.__x_remaning_transitions();
+          // We only want to run remaining transitions in the end if they exists
+          if (el.__x_remaining_transitions) {
+            el.__x_remaining_transitions();
           }
         }.bind(this), duration);
       }.bind(this));
@@ -6250,8 +6250,8 @@
     var initialUpdate = arguments.length > 4 && arguments[4] !== undefined ? arguments[4] : false;
 
     // Resolve any previous pending transitions before starting a new one
-    if (el.__x_remaning_transitions) {
-      el.__x_remaning_transitions();
+    if (el.__x_remaining_transitions) {
+      el.__x_remaining_transitions();
     }
 
     var hide = function hide() {
@@ -6293,7 +6293,7 @@
             _newArrowCheck(this, _this2);
 
             // If previous transitions still there, don't use resolve
-            if (el.__x_remaning_transitions) {
+            if (el.__x_remaining_transitions) {
               hide();
             } else {
               resolve(function () {
--- a/dist/alpine-ie11.js
+++ b/dist/alpine-ie11.js
@@ -5490,7 +5490,7 @@
     var additionalHelperVariables = arguments.length > 2 && arguments[2] !== undefined ? arguments[2] : {};
 
     if (typeof expression === 'function') {
-      expression.call(dataContext);
+      return expression.call(dataContext);
     } // For the cases when users pass only a function reference to the caller: `x-on:click="foo"`
     // Where "foo" is a function. Also, we'll pass the function the event instance when we call it.
 
@@ -5510,7 +5510,7 @@
     var name = replaceAtAndColonWithStandardSyntax(attr.name);
     return xAttrRE.test(name);
   }
-  function getXAttrs(el, type, component) {
+  function getXAttrs(el, component, type) {
     var _this2 = this;
 
     return Array.from(el.attributes).filter(isXAttr).map(parseHtmlAttribute).flatMap(function (i) {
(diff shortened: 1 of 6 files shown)


SimoTod:
It seems a bit weird. Basically we are not responding to the user interaction until the initial transition has finished? Have I understood that correctly? 
Shouldn't it work the other way around with the last transition always early terminates the previous one (like jQuery used to do)? Do we know what Vue does in this case?

HugoDF:
@muzafferdede master doesn't match the latest released version hence why the built output might not be up to date

muzafferdede:
@SimoTod yes, you got it correct. I try to do it like you mentioned earlier, terminate the current transition and move to the last one but I got some promise errors. Also at some point, I want to stop user Interaction because it causes the animation chain to behave weird. I might be wrong but right now everything behave as I expected.

@HugoDF yeah I noticed that. I guess it the issue you opened. Let's see what Caleb says on that.

HugoDF:
Well the issue is a bit different

We just haven't cut a release with the latest changes but the PRs have been merged is all

muzafferdede:
> Well the issue is a bit different
> 
> We just haven't cut a release with the latest changes but the PRs have been merged is all

What i saw is that Caleb merged selective commits out of some PR and that cause build files out of date. And when we pull latest changes, we end up with old build files with new source with changes. And surely once we build, we end up with new build files.

HugoDF:
Yep, no one is supposed to use master built files anyways (even the CI runs build and then test) so no issues really

SimoTod:
> I might be wrong but right now everything behave as I expected.

How about x-transition directives? They won't add an inline transition-duration style so I'm not sure it would work.

Also, it sounds to me like if you have a div with `x-show.transition="open"` and a button with `x-click="open = !open"` and you click twice quickly on the button, you will change the flag two times but just run the first animation so you'll get in an inconsistent state.

Not sure it will behave as expected most of the time.
I can have a play tonight to confirm.

muzafferdede:
@SimoTod `x-show.transition` and `x-transition` they end up using `style` of the element to` bind animations`. So me checking `transition-duration` works on both. And good news is that i found out there is an existing `skip option` in transitions. So i just implement it. I am happy with how it turned out now.

HugoDF:
Sounds like a debounce vs throttle kind of situations and

muzafferdede:
@SimoTod No worries. I think understanding the issue in technical part will help us to create the tests. Once we know what cause the issue and how we can prevent, definitely will have test for it.

For now i am digging the **transitions** and i found out that both `x-show.transition and x-transition` uses this `chain mechanism` to execute `start -> during -> show -> end -> hide -> cleanup` steps. Only difference is `x-show.transition` uses style attribute on the element and x-transition uses css classes`. But they both uses same executing order. So if we fix it here, both will be fixed.

```js
export function transition(el, stages) {
    stages.start()
    stages.during()

    requestAnimationFrame(() => {
        // Note: Safari's transitionDuration property will list out comma separated transition durations
        // for every single transition property. Let's grab the first one and call it a day.
        let duration = Number(getComputedStyle(el).transitionDuration.replace(/,.*/, '').replace('s', '')) * 1000

        if (duration === 0) {
            duration = Number(getComputedStyle(el).animationDuration.replace('s', '')) * 1000
        }

        stages.show()

        requestAnimationFrame(() => {
            stages.end()

            setTimeout(() => {
                stages.hide()

                // Adding an "isConnected" check, in case the callback
                // removed the element from the DOM.
                if (el.isConnected) {
                    stages.cleanup()
                }
            }, duration);
        })
    });
}
``` 

i narrow down to this point where i believe the fix can be applied here. The setTimeout thing cause the issue IMO. No matter what, the setTimeout will run after the given duration regardless if element on animation. 

Any thoughts?

SimoTod:
VueJS has the concept of once functions (function which can run only one time ->
https://github.com/vuejs/vue/blob/76fd45c9fd611fecfa79997706a5d218de206b68/src/shared/util.js#L325) and it applies this concept to transitions.
If your timeout callback is a once function and you find a way to call the same function before starting a new transition, it will be a step in the right direction, I think.

Things to be aware of: 
- You don't want to touch/reset the existing transition if the direction (in/out) is the same: if you are in the middle of a "show" transition and the component refresh because something else has changed, x-show and x-if should not trigger a new "show" transition.
- the logic is split in different files, making it a bit hard to work on, so you might need to stash the once callback into a new Alpine property on the el object (el.__x_something_something = yourOnceFunction) so you can call it from a different file if needed.

SimoTod:
I suggest you close this PR and reopen a new one. There are a lot of messages referred to the previous version which are going to confuse people but up to you. 

Also
```html
<div x-data="{open: false, foo: ''}">
  <div x-show.transition.duration.5000="open">test</div>
  <button @click="open = !open">toggle</open>
  <input x-model="foo">
</div>
```
Click on the toggle button and type something in the input field before 5s. I would expect the first transition to carry on and the additional transitions created by each repainting to be ignored since the direction of the animation does not change. Is it covered?

muzafferdede:
Alright i will close this one.

and for your sample code, there you go
https://codepen.io/muzafferdede/pen/ZEQzqZL

muzafferdede:
Closing this issue due to prevent confusing with related previous issues.

SimoTod:
The codepen does not work correctly. Transition is supposed to last 5 second but the x-model terminates it early

muzafferdede:
@SimoTod Good call, i guess this will do what you expect now.

https://codepen.io/muzafferdede/pen/ZEQzqZL

SimoTod:
Yep, that looks definitely better, thanks

muzafferdede:
Finally huh? :)

PR #538, opened 2020-05-30
Async components

imacrayon:
I modified component initialization to be asynchronous. Now `x-data` can accept a promise that will eventually resolve the component's scope. This means you can have your component code in a separate file that will automatically load on the fly when your component is added to the page.

For example:
```
// async-text.js

export default function () {
  return { text: "I'm in another file!" }
}

```

```
<div x-data="import('./async-text.js').then(component => component.default())">
  <div x-text="text"></div>
</div>
```

### DIFF
--- a/examples/async-dropdown.js
+++ b/examples/async-dropdown.js
@@ -1,0 +1,5 @@
+export default function () {
+    return {
+        open: false,
+    };
+}
--- a/examples/index.html
+++ b/examples/index.html
@@ -202,6 +202,21 @@
                 </tr>
 
                 <tr>
+                    <td>Async Component</td>
+                    <td>
+                        <div x-data="import('/async-dropdown.js').then(component => component.default())">
+                            <button x-on:click="open= true">Open Async Dropdown</button>
+                            <ul
+                                x-bind:class="{ 'hidden': ! open }"
+                                x-on:click.away="open= false"
+                                x-cloak>
+                                Async Dropdown Body
+                            </ul>
+                        </div>
+                    </td>
+                </tr>
+
+                <tr>
                     <td>Append DOM</td>
                     <td>
                         <div id="goHere">Click me.</div>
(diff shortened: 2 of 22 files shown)


imacrayon:
🤦‍♂️ I ran my local tests against the wrong build

SimoTod:
Thanks for your PR

Unfortunately, most of the existing tests are failing, do you know why? 
We need to understand why they were failing: is it slower? Are we adding an additional delay/flickering when the page loads even for component that are supposed to be sync? Are we introducing BC for some users).
If there are no risks, they should be updated before this PR can be reviewed.
Also, we need a test for the additional feature we want to add (It can be as simple as the manual example you created) to avoid future regressions.

As a separate note, it's probably worth to ask @calebporzio about the syntax before implementing it since it will become part of the public API (aka, really hard to change in the future).

For example
`<div x-data="import('./async-text.js').then(component => component.default())">`
vs
`<div x-data="import('./async-text.js')">`
vs
`something else`

Thanks again.

HugoDF:
The reason tests are failing is because we've turned component initialisation from a fully sync operation to an async one. promise.resolve I think yields back to the event loop, what we could do to fix this is to `return Promise.resolve()...` and `await Alpine.start()` in the tests?

imacrayon:
Thanks for the feedback. Hugo, was right in his assumption with the test failures. I've _almost_ got everything back to green.

This change ended up having more side effects that I had originally anticipated and it has slowed the test suite down a bit because of all the async operations. I'd be interested if to know if this is a feature @calebporzio would still consider. I saw `Alpine.component()` on the roadmap, maybe that would be a more appropriate place to do something like this, I wondering if I should hold off until then?

ryangjchandler:
I'm not sure if this is something that Alpine should support first party, since it's not difficult to implement in user-land.

HugoDF:
Considering @ryangjchandler 's userland solution and the fact we're likely to be getting components down the line, I wouldn't invest more time in this. Feel free to dig around or have a stab at other bugs though 🙂

ryangjchandler:
Not that my userland solution doesn't use async / await by default but it can.

imacrayon:
Yep! I think @ryangjchandler's approach gets the job done.

Thanks everyone!

ryangjchandler:
No worries @imacrayon, it's quite opinionated but it works quite well.

Issue #170, opened 2020-02-11
website breaking x-transition bug (update: 1.20.0)


TomS-:
This post was originally about implementing transition lists (https://vuejs.org/v2/guide/transitions.html#List-Move-Transitions), however, the more I use transition, the more I realise that they are slightly buggy and may need work.

I feel like I really good thing to implement is a way of queuing animations so it doesn't happen all at once (https://vuejs.org/v2/guide/transitions.html#Transition-Modes) see "in-out" and "out-in". Maybe you can add x-transition to a template tag for this to work.

For the bugs that I have come across:
1. Animations will not fire when called too often (https://codepen.io/TomS-/pen/poJjpNp) this is happening to me on FireFox, you'll see if you press next often enough the text will disappear. 
2. This maybe an issue with Flickity but (https://codepen.io/TomS-/pen/dyoYmvK) the slider completely break when animation is applied see (https://codepen.io/TomS-/pen/poJjpNp)

TomS-:
Noticed that @calebporzio has done a release addressing some of this issue. So I'm just doing an update to show my findings on v1.11.0 (https://codepen.io/TomS-/pen/eYNJZOx)

Fixed:
1. The text that disappears if you click next fast enough no longer disappears.

Still doesn't work:
1. If you press "People" then "Fruit" before the transition finishes the sliders disappear, both receiving `display: none`.
2. The sliders are still not interactive when an transition is added though they appear less broken. Resizing the window fixes it, so I think there is a sequencing issue when a transition is applied because $nextTick should introduce it's interactivity (https://flickity.metafizzy.co/api.html#resize).

TomS-:
Update: 1.20.0 (https://codepen.io/TomS-/pen/BaNKmaV)

Using `x-show.transition` seems to stop issue 3 (as per above). I believe this is down to transition time, a larger transition time increases the chances of a bug being produced. I'm unable to reproduce issue 1 as the transition happens too fast (you have to catch it between transitions). The reactivity of Flickity still doesn't work, but if you click "People" then "People" again, reactivity will work. I think there is an issue with the timings when transitions of used as this is not an issue when transitions are not used.

calebporzio:
Hey @TomS- It seems like there are a couple of different issues you're referencing here.

Let's just start with interrupting slow transitions breaking things.

I've encountered that issue too. What is the best behavior in a case like that? Let the transition finish? Interrupt it? Block interactions during?

I'd love help researching how Vue and other packages handle this.

[...]

Issue #156, opened 2020-01-31
:class bindings and x-transition are not compatible


danfascia:
Transitions don't work if you also have a `:class` binding on the same element as the `x-transition`. I would imagine that this is something to do with a conflict of multiple events trying to write to the `class` but it's pretty inconvenient.

Has this been noticed before and is there any workaround?

Here is an illustration of the problem, if you trigger an error message by clicking any button then I have created two error messages, one with and one without the `:class` binding to show the issue.

https://codepen.io/danfascia/pen/dyPrwvB?editors=1010

danfascia:
OK, so I've updated and documented in the codepen a solution but it is a workaround.

By separating the `:class` binding and `x-transition(s)` onto different `<div>` elements I can make it work as expected but I still think it is a bit of a bug.

danfascia:
@calebporzio is this a known issue / intended feature or do you think it is worth adding something to the docs about the incompatibility?

SimoTod:
I don't think it's the desired behaviour so it will probably be fixed in one of the next releases. Another workaround for now, since Alpine processes its directives in order, is to put `:class` before `x-show`.

danfascia:
That's interesting, I think I've spotted that since there were inconsistencies in producing this behaviour. Would be good to fix.

calebporzio:
Ok, yeah I think this is a legitamate bug and should be addressed. Would someone mind doing some exploration into the problem to make things easy for me (if not PRing the entire solution/test itself) - Thanks!

SimoTod:
I think the issue here is that we store the original classes when a component is initialised and `:class`, every time is recalculated, wipes out the old classes and restore the original classes + its own result.

This could also be annoying if, in the meantime, a different piece of javascript added some classes to the element.

On the other hand, we can't read the current classes and use them because they will contain results from the previous evaluation of :class that we may want to remove.

ryangjchandler:
Hey everyone that was on this thread - I forked the original CodePen and updated to latest version of Alpine and tested and I couldn't see any problems with the class bindings and transitions (https://codepen.io/ryangjchandler/pen/JjXGZyK).

I'm going to close this issue for now since there wasn't much information on the problem in the first place, but @danfascia if you still think this is a problem, please re-open this issue and let us know (with some new example). Cheers! 🙂

PR #542, opened 2020-06-02
Fixed x-show transition issues by refactoring transitions

### LINKED ISSUE #533: Might be an Alpine Performance bug on `x-show` with transition

muzafferdede:
Fixes #533 (tested) |  #170 (tested, need more test) | #156 (partly tested, need more test) | `you tell me`

Try to click fast on the button to see basic issue

With issue basic sample:
https://codepen.io/muzafferdede/pen/ExPYzwN

With this PR Fix basic sample
https://codepen.io/muzafferdede/pen/NWxKVwV

There is a related simpler PR as well which is applied on current structure. I will update the mention once i created the PR. #543  

As you know this is a long coming issue. I spend few days on it and finally get it working with helpful comments of yours. I have to refactor transition to achieve this solution at first because it was hard to figure out transition within the utilities. So it was best to move it to its own space. 

By refactoring i actually still keep almost entire code as same as before because of the warnings i got from you and knowing @calebporzio  is not fan of changes without benefits. But I figured out that what i achieved is costed much more on Vue, React for example. They have to track entire transition and execute at each stage a function call to prevent overlapping. 

#### Here is what Vue needs to define for transition to work:
```js
const {
    css,
    appear,
    enterClass,
    enterActiveClass,
    appearClass,
    appearActiveClass,
    beforeEnter,
    onEnter,
    enter,
    afterEnter,
    enterCancelled,
    beforeAppear,
    onAppear,
    appear,
    afterAppear,
    appearCancelled
  } = resolveTransition(data, vnode.context)

...
const el: any = vnode.elm
  const vm = vnode.context
  // call enter callback now
  if (el._enterCb) {
    el._enterCb.cancelled = true
    el._enterCb()
  }
  const data = vnode.data.transition
  if (!data) {
    return rm()
  }


```
There is much more i just don't want to make this longer. React did similar approach as well.

####  Here is what i defined to make this work:
```js
el.__x_transition_remaining,
el.__x_transition_last_value
```
By paying only few assignments,  we gain transition power. 

I call this a win specially fixing all those old issues. Hope you call it too and this PR get merged. 

Let me know if there is any room to improve. Huge thanks @SimoTod and @HugoDF for their comments. .5 x more thanks to @SimoTod for bearing with me and making good arguments on the issue :).

Thanks


src/transitions.js
```diff
@@ -0,0 +1,250 @@
...
+
+    let ensureStringExpression = (expression) => {
+        return typeof expression === 'function'
+            ? component.evaluateReturnExpression(el, expression)
+            : expression
+    }
+
+    // Prepare stage group names for given directions
+    let cssClasses = {
+        durring: transition,
```
HugoDF, line 133: Is this a typo?

muzafferdede: Which? one,`cssClasses`? 

No, down on the road it will serve as css class container. The description was for the pre assigned stuff. Basically css classes assigned by their stage group names. So at this point we were preparing the container.

SimoTod: I think Hugo was referring to the word `durring`

muzafferdede: oh it is a typo then.

ryangjchandler:
Do these changes have much of an effect on the bundle size? Obviously, refactoring is nice and it fixes a bug but don't want to drastically add bytes.

muzafferdede:
@ryangjchandler I did check that at some point and It was about to 2kb less then before. 

Simone mentioned that some part is not clear to him so I did add few comments to describe the code. 

But I think code is already readable, so yeah, we could even go further on file size we need too.

SimoTod:
It's still smaller current master is 66.6KB, this build is 66.3KB (unzipped). 
The minified version strips out the comments so I would not remove them, personally.

src/directives/if.js
```diff
@@ -1,4 +1,4 @@
-import { transitionIn, transitionOut } from '../utils'
+import { transitionIn, transitionOut } from '../transitions'
```
SimoTod, line 1: do we need to apply the same fix we did in show? So transition won't overlap?

src/directives/show.js
```diff
@@ -1,55 +1,58 @@
...
         if (! value) {
             if ( el.style.display !== 'none' ) {
-                transitionOut(el, () => {
-                    resolve(() => {
-                        hide()
-                    })
-                }, component)
+                transitionOut(el, component, () => {
+                    // If previous transitions still there, don't use resolve
+                    if (el.__x_transition_remaining) {
```
SimoTod, line 25: Similar to the other PR, I'm sure we need it but I can't understand why. Can you explain please?

SimoTod: also the test passes master, I think we need a better one.

src/directives/if.js
muzafferdede: I haven't went deeper than `x-show`. But i did make sure refactoring did not effect existing directives. Anyways since the remaining transitions are carried with `el`, we can run `el.transition_remaining()` anytime, anywhere, anyhow.

src/directives/show.js
muzafferdede: - Oh It's the where *black hole* begins. In case we have a transition going on but if value did not change since last time, we should not resolve it in that case. This part is related to `immediate` and `hide child elements before parent` related thing. I got lost a lot here. Even while i am explaining to you now :)
- Test has been updated and now master fails these but PRs passes. There was a timing issue where master flicks :) So i needed to find out best time the falsy thing happens.

muzafferdede:
I did apply @SimoTod recommendation "once" method instead of using setTimeout for to make sure we did not overlap to this PR as well. We are 1 variable (`el.transition_timer`) less now.

muzafferdede:
> It's still smaller current master is 66.6KB, this build is 66.3KB (unzipped).
> The minified version strips out the comments so I would not remove them, personally.

Currently master is **66.6KB** and this build is **66KB** (unzipped). By considering all the comments i added, the minified version will have more impact on file size.

calebporzio:
Thank you for the cleanup contribution. However, I'm going to hold off on merging this right now.

PR #543, opened 2020-06-02
Fix x show transition overlap

### LINKED ISSUE #533: Might be an Alpine Performance bug on `x-show` with transition

muzafferdede:
Fixes #533 (tested) |  #170 (tested, need more test) | #156 (partly tested, need more test) | `you tell me`

I have to refactor transition to achieve this solution. So there is a #542 Refactored Transition PR  of this PR which has same solution + refactored transitions to let transitions has its own space so it will be easier to work on it. (it was really hard to work on existing transition code). I have explained much more on #542 since they serve almost same purpose.  

This fix is quite self explanatory. I assigned few variables to check transitions and make sure and control how they are rendered.

By this, quick clicks overlapping transitions and wrong state of display issues will be fixed. There is also a test called `remaining transitions forced to complete if they exists` which tests this bug.

Let me know if there is any room to improve. Huge thanks @SimoTod and @HugoDF for their comments. .5 x more thanks to @SimoTod for bearing with me :).

### DIFF
--- a/src/directives/show.js
+++ b/src/directives/show.js
@@ -2,7 +2,7 @@ import { transitionIn, transitionOut } from '../utils'
 
 export function handleShowDirective(component, el, value, modifiers, initialUpdate = false) {
     // Resolve any previous pending transitions before starting a new one
-    if (el.__x_remaining_transitions) {
+    if (el.__x_remaining_transitions && el.__x_current_transition_value !== value) {
         el.__x_remaining_transitions()
     }
 
@@ -50,9 +50,12 @@ export function handleShowDirective(component, el, value, modifiers, initialUpda
                 }, component)
             }
 
-            // Resolve immediately, only hold up parent `x-show`s for hidin.
+            // Resolve immediately, only hold up parent `x-show`s for hiding.
             resolve(() => {})
         }
+
+        // Asign current value to el to check later on for preventing transition overlaps
+        el.__x_current_transition_value = value
     }
 
     // The working of x-show is a bit complex because we need to
@@ -72,8 +75,10 @@ export function handleShowDirective(component, el, value, modifiers, initialUpda
         component.executeAndClearRemainingShowDirectiveStack()
     }
 
-    // We'll push the handler onto a stack to be handled later.
-    component.showDirectiveStack.push(handle)
+    // If x-show value changed from previous transition we'll push the handler onto a stack to be handled later.
+    if (el.__x_current_transition_value !== value) {
+        component.showDirectiveStack.push(handle)
+    }
 
     component.showDirectiveLastElement = el
 }
--- a/src/directives/show.js
+++ b/src/directives/show.js
@@ -28,36 +28,27 @@ export function handleShowDirective(component, el, value, modifiers, initialUpda
     }
 
     const handle = (resolve) => {
-        if (! value) {
+        if(value) {
+            transitionIn(el,() => {
+                show()
+            }, component)
+            resolve(() => {})
+        } else {
             if ( el.style.display !== 'none' ) {
                 transitionOut(el, () => {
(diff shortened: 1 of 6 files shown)


src/utils.js
```diff
@@ -397,14 +398,29 @@ export function transition(el, stages) {
...
+                if(el.__x_transition_timer){
+                    clearTimeout(el.__x_transition_timer)
+                }
+            }
+
+            el.__x_transition_timer = setTimeout(() => {
+                // We only want to run remaining transitions in the end if they exists
+                if (el.__x_transition_remaining) {
+                    el.__x_transition_remaining()
+                }
```
SimoTod, line 423: I'd really like to use the once function from VUE here. This feels a bit clumsy somehow. It's like we are trying to clear as many variables as possible to make sure the timeout doesn't run.
I'm almost sure we can get the same result with
```
el.__x_transition_remaining = once(() => {
    stages.hide()

    // Adding an "isConnected" check, in case the callback
    // removed the element from the DOM.
    if (el.isConnected) {
        stages.cleanup()
    }

     // Safe to remove transition from el since it is completed
     delete el.__x_transition_remaining
})

setTimeout(el.__x_transition_remaining, duration);
```

src/directives/show.js
```diff
@@ -1,6 +1,11 @@
 import { transitionIn, transitionOut } from '../utils'
 
 export function handleShowDirective(component, el, value, modifiers, initialUpdate = false) {
+    // Resolve any previous pending transitions before starting a new one
```
SimoTod, line 4: is it relevant only for show? transitions are applied to x-if as well.

test/transition.spec.js
```diff
@@ -620,3 +620,91 @@ test('x-transition supports css animation', async () => {
     )
     expect(document.querySelector('span').classList.contains('animation-leave')).toEqual(false)
 })
+
+test('remaining transitions forced to complete if they exists', async () => {
```
SimoTod, line 624: I run the same test against master and it did not fail. It seems a bit complex, is it supposed to test this feature?

src/directives/show.js
```diff
@@ -26,9 +31,14 @@ export function handleShowDirective(component, el, value, modifiers, initialUpda
         if (! value) {
             if ( el.style.display !== 'none' ) {
                 transitionOut(el, () => {
-                    resolve(() => {
+                    // If previous transitions still there, don't use resolve
+                    if ( el.__x_transition_remaining ) {
```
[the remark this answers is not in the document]
muzafferdede: As i said at #542 if value did not change but if there is a remaining transition, we should not use resolve.

src/utils.js
muzafferdede: Good news, i finally did figure out how to implement the way you mentioned. Good call @SimoTod. We are 1 variable less now :)

test/transition.spec.js
muzafferdede: Fixed by improving the test

src/directives/show.js
muzafferdede: As i said in #542 we can run `el.transition_remaning()` any time if we need to clear transitions.

src/utils.js
```diff
@@ -397,14 +398,29 @@ export function transition(el, stages) {
         requestAnimationFrame(() => {
             stages.end()
 
-            setTimeout(() => {
+            // Asign current transition to el in case we need to force it
```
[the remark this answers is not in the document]
muzafferdede: Thanks, should be fine with yours :)

muzafferdede:
Fixed by assigning `el.transition_last_value` at initial.

PR #545, opened 2020-06-03
Allow clients to pause the observer

bep:
This commit adds a flag to pause/start the mutation observer.

This allows clients using Turbolinks and similar to do the needed DOM cleaning without data races.

This is a stripped down version of https://github.com/alpinejs/alpine/pull/455

Updates #319
Updates #372

### DIFF
--- a/dist/alpine-ie11.js
+++ b/dist/alpine-ie11.js
@@ -5490,7 +5490,7 @@
     var additionalHelperVariables = arguments.length > 2 && arguments[2] !== undefined ? arguments[2] : {};
 
     if (typeof expression === 'function') {
-      expression.call(dataContext);
+      return expression.call(dataContext);
     } // For the cases when users pass only a function reference to the caller: `x-on:click="foo"`
     // Where "foo" is a function. Also, we'll pass the function the event instance when we call it.
 
@@ -5510,7 +5510,7 @@
     var name = replaceAtAndColonWithStandardSyntax(attr.name);
     return xAttrRE.test(name);
   }
-  function getXAttrs(el, type, component) {
+  function getXAttrs(el, component, type) {
     var _this2 = this;
 
     return Array.from(el.attributes).filter(isXAttr).map(parseHtmlAttribute).flatMap(function (i) {
@@ -5586,8 +5586,8 @@
     var forceSkip = arguments.length > 3 && arguments[3] !== undefined ? arguments[3] : false;
     // We don't want to transition on the initial page load.
     if (forceSkip) return show();
-    var attrs = getXAttrs(el, 'transition', component);
-    var showAttr = getXAttrs(el, 'show', component)[0]; // If this is triggered by a x-show.transition.
+    var attrs = getXAttrs(el, component, 'transition');
+    var showAttr = getXAttrs(el, component, 'show')[0]; // If this is triggered by a x-show.transition.
 
     if (showAttr && showAttr.modifiers.includes('transition')) {
       var modifiers = showAttr.modifiers; // If x-show.transition.out, we'll skip the "in" transition.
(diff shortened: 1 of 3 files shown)


ryangjchandler:
Hey @bep, makes perfect sense to add since, in reality, Alpine can't cover all bases. Would you mind re-building the dist files and maybe adding a test.

The test might be difficult since the MutationObserver is being "mocked" / "mimicked" in the test.

bep:
I have re-built the dist files -- but I have no idea how to (relatively simply) add a test for this...

ryangjchandler:
> I have re-built the dist files -- but I have no idea how to (relatively simply) add a test for this...

No worries, I'll have a look later tonight to see if I can add a test for you. Thanks again for the contribution, will be useful in many cases.

ryangjchandler:
I'm also wondering whether we want some nice developer experience helpers, maybe `Alpine.pauseObserver()` and `Alpine.unpauseObserver()`, not a big one but it might be nicer than relying on the developer to set the booleans themselves.

Also means that the public API doesn't need to change if we want to add more logic in, in the future. No breaking changes, etc.

bep:
>Alpine.pauseObserver() and Alpine.unpauseObserver()

The above makes for an awful big API for a small thing, but hiding the boolean behind a `setObserverState(running)` or something would probably be a good idea (which would hide the implementation detail).

ryangjchandler:
Yeah, I wasn't set on the actual API I suggested, but I think the implementationn needs to be hidden.

calebporzio:
@bep & @SimoTod - will this allow you to create a Turbolinks Plugin?

bep:
>will this allow you to create a Turbolinks Plugin?

I would say yes. I have this patch successfully running with @SimoTod 's code from the other PR running in my own `turbolinks:*` listeners.

SimoTod:
Hi Caleb, when we started investigating we had a lot of issues with the mutation observer so it was one of the first thing we applied. The reason was that turbolinks was adding cached html generated by x-if and x-for but since they didn't have the __x_* properties, alpine was duplicating and trying to reevaluating stuff which was not supposed to. I suspect it may be not strictly necessary now that we clean the html before cache and restore the permanent data manually but I need to double check.
However, even if not necessary, it may have a performance impact since each turbolink page will be evaluated twice if we don't stop the observer.
I'll start a proof of concept and confirm tomorrow. @bep I'll probably ask you to help testing since your app is going to stress the plugin a lot.

SimoTod:
It seems we still need it to avoid errors in console.

bep:
@SimoTod I have been thinking a little. There are now several issues floating around related to getting _Alpine to take a break_: This issue, `x-ignore` and most recently `x-skip`.

It would maybe be 

* more in line with the rest of Alpine's API if we could make the pausing of the observer a directive/DOM attribute that we could remove once Turbolinks  is done.
* if that is possible, maybe it can be consolidated with one or both of the above...

SimoTod:
I would wait for Caleb to have a look first. X-ignore was already rejected once but you could probably use one of those directives instead of pausing tge observer if they are accepted. 
I don't dislike the idea of an API to pause the observer, though. It's a bit invasive at the minute.

calebporzio:
Thanks for this, will be included in the next release!

SimoTod:
@bep Can you check if https://cdn.jsdelivr.net/gh/SimoTod/alpine-turbolinks-adapter@eac088b4067e1a8d8ea21b5182458b8afb9435ac/dist/alpine-turbolinks-adapter.min.js
fixes your Turbolinks issues please?
It has to be included in your page along Alpine 2.4.0 or above. Order does not matter and you can defer both scripts if needed.

It could probably be moved to the Alpine organisation so don't use it in production yet. Thanks

PR #547, opened 2020-06-03
feat(warn-template-multiple-root-elements)

### LINKED ISSUE #539: `x-for` error message could be more explicit when more than a single root element
## Current Behavior

To be told that _you can't have more than one root element inside a `x-for`_ when I'd accidentally coded with two + root elements. Please not this is invalid HTML...

```
<ul x-data="handler()">
        <span x-text="item"></span>
        <li x-text="item"></li>
    </template>
</ul>
```

The console error is;

```
VM342:3 Uncaught (in promise) ReferenceError: item is not defined
    at eval (eval at a (alpine.js:101), <anonymous>:3:27)
    at a (alpine.js:101)
    at ne.evaluateReturnExpression (alpine.js:1535)
    at alpine.js:1502
    at ne.resolveBoundAttributes (alpine.js:1484)
    at ne.initializeElement (alpine.js:1403)
    at alpine.js:1387
    at alpine.js:1377
    at e (alpine.js:77)
```

## Expected Behavior

To have an error message that says the problem is multiple root elements inside `x-for` rather than `ReferenceError: ${thing} is not defined`

## Failure Information (for bugs)

[Codepen demo](https://codepen.io/stuartjnelson/pen/bGVXQJo)

HugoDF:
Closes #539 

rationalise the warnings for `template` elements:
- if a directive needs a template but isn't using one -> log warning
- if a directive expects a template but it has multiple root elements

### DIFF
--- a/src/directives/for.js
+++ b/src/directives/for.js
@@ -1,7 +1,7 @@
-import { transitionIn, transitionOut, getXAttrs, saferEval } from '../utils'
+import { transitionIn, transitionOut, getXAttrs, warnIfMalformedTemplate } from '../utils'
 
 export function handleForDirective(component, templateEl, expression, initialUpdate, extraVars) {
-    warnIfNotTemplateTag(templateEl)
+    warnIfMalformedTemplate(templateEl, 'x-for')
 
     let iteratorNames = typeof expression === 'function'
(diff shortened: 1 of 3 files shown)


calebporzio:
Love this. Thanks @HugoDF!

PR #554, opened 2020-06-05
Extract convertClassStringToArray util

HugoDF:
We're taking "class names" and converting them to a list of classes in a couple of spots, this extracts that functionality to a util so we don't forget to filter out `''` for example.

Shouldn't affect existing functionality.


calebporzio:
Thanks @HugoDF!

PR #555, opened 2020-06-05
Defer nextTick to the next animation frame

SimoTod:
Minimal PR to address the nextTick issue when used with animations.
It also contains a test which fails with the current master branch.

### DIFF
--- a/src/component.js
+++ b/src/component.js
@@ -187,11 +187,14 @@ export default class Component {
 
     executeAndClearNextTickStack(el) {
         // Skip spawns from alpine directives
-        if (el === this.$el) {
-            // Walk through the $nextTick stack and clear it as we go.
-            while (this.nextTickStack.length > 0) {
-                this.nextTickStack.shift()()
-            }
+        if (el === this.$el && this.nextTickStack.length > 0) {
+            // We run the tick stack after the next frame to allow any
+            // running transitions to pass the initial show stage.
+            requestAnimationFrame(() => {
+                while (this.nextTickStack.length > 0) {
+                    this.nextTickStack.shift()()
+                }
+            })
         }
     }
 
--- a/test/next-tick.spec.js
+++ b/test/next-tick.spec.js
@@ -44,3 +44,35 @@ test('nextTick wait for x-for to finish rendering', async () => {
 
     await wait(() => { expect(document.querySelector('p').innerText).toEqual(3) })
 })
+
+test('$nextTick works with transition', async () => {
+    jest.spyOn(window, 'requestAnimationFrame').mockImplementation((callback) => {
+        setTimeout(callback, 0)
+    });
+
+    document.body.innerHTML = `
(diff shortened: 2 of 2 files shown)


killgt:
I think this may introduced a suble bug. We are opening a new tab on the browser on a click event (to avoid popup blocking), then doing a next tick call, because we need the dom to be updated to submit the form on the new opened window, but the next tick is not called until you focus the tab again.  Maybe the next tick stack should be called as soon as the next window blur event

SimoTod:
@kill I think that use case is a bit of a hack though. The problem is that the next animation frame is required otherwise nextTick won't work with Alpine transitions (e.g. Focusing a field when showing a modal) which is a wider use case. You could maybe change it to a setTimeout 0 in nextTick but, to be fair, I would use the setTimeout in your implementation rather.

SimoTod:
> I think this may introduced a suble bug. We are opening a new tab on the browser on a click event (to avoid popup blocking), then doing a next tick call, because we need the dom to be updated to submit the form on the new opened window, but the next tick is not called until you focus the tab again.  Maybe the next tick stack should be called as soon as the next window blur event

@killgt I think that use case is a bit of a hack, though. The problem is that the next animation frame is required otherwise nextTick won't work with Alpine transitions (e.g. Focusing a field when showing a modal) which is a wider use case. You could maybe change it to a setTimeout 0 in nextTick but, to be fair, I would use the setTimeout in your implementation rather.

killgt:
Yes, it a bit of a hack. If a fix isn't a posibility, the docs should warn about the nextTick (and alpine reactivity) will not continue while the current page is on the background (or the browser decides it dosent want to update the current page)

SimoTod:
I'm not sure i understand completely how you use it, to be fair. If you are interacting with a page, you are supposed to be on that page for nextTick to do its stuff. How do you queue a nextTick call on a page that is not focused?

However, the documentation is in the repo, you are welcome to send PRs I've no saying in what gets merged into the repo as I'm not the maintainer. 

Alternatively, you can also PR a change for the implementation if it doesn't break other features.

killgt:
The problem simple to reproduce, queue a nextTick call, then opening a new TAB, you'll expect that your js will continue executing, but it isn't. 

The use case is a tricky one, as you cannot open a window if it isnt done on the click event, but we need a form on the page to be rendered and submitted to the just opened tab, i won't waste your time triying to detail why it has to be done on this way.

I'm not saying that this PR is breaking anything or wrong, just subtly changing how Alpine is run and maybe sometime in the future it can bite us back.

SimoTod:
So you have a click handler on a link that uses nextTick but the click itself open a new page in a new tab. Gotcha. 
Note that this PR was merged 13 months ago, it's not something that has changed recently.

By the way, I think there's still a v2 branch where you can send PRs for bug fixes if you want to change it. I'm not precious about that implementation. 👍

PR #558, opened 2020-06-06
Feature: x-model checkbox array `.number` modifier

### LINKED ISSUE #296: [Feature Request] Number modifier for checkboxes bound with x-model
From what I understand from the source the number modifier is not supported on the x-model directive for checkboxes. This makes seeding initial data a bit of a hassle as these have to be cast to string to work with the value from the checkboxes.

The behaviour is also inconsistent as the initial state of the checked attribute is set correctly but when you then click the checkbox that was set by the initial state it wont get unchecked.

The example below shows the problem - the second checkbox is checked by default and can not be unchecked by clicking it.

```html
<div
    x-data="{
        selected: [2]
    }"
>
    <input type="checkbox" value="1" x-model="selected" />
    <input type="checkbox" value="2" x-model="selected" />
    <input type="checkbox" value="3" x-model="selected" />
    <input type="checkbox" value="4" x-model="selected" />
</div>
```

HugoDF:
Closes #296 

**Changes**

- Extract `safeParseNumber` from existing `number` modifier code (for `select` and other non-checkbox inputs) - doing this means bundle size is the same/smaller despite adding new functionality
- Apply `safeParseNumber` during checkbox `x-model` computation

**Use case**

Use case is as follows (per the Issue), I've copied it verbatim into a new test.

> The example below shows the problem - the second checkbox is checked by default and can not be unchecked by clicking it.

```html
<div
    x-data="{
        selected: [2]
    }"
>
    <input type="checkbox" value="1" x-model="selected" />
    <input type="checkbox" value="2" x-model="selected" />
    <input type="checkbox" value="3" x-model="selected" />
    <input type="checkbox" value="4" x-model="selected" />
</div>
```

### DIFF
--- a/src/directives/model.js
+++ b/src/directives/model.js
@@ -32,9 +32,10 @@ function generateModelAssignmentFunction(el, modifiers, expression) {
         if (event instanceof CustomEvent && event.detail) {
             return event.detail
         } else if (el.type === 'checkbox') {
-            // If the data we are binding to is an array, toggle it's value inside the array.
+            // If the data we are binding to is an array, toggle its value inside the array.
             if (Array.isArray(currentValue)) {
-                return event.target.checked ? currentValue.concat([event.target.value]) : currentValue.filter(i => i !== event.target.value)
+                const newValue = modifiers.includes('number') ? safeParseNumber(event.target.value) : event.target.value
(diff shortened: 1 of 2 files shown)


calebporzio:
Thanks @HugoDF !

mikaeljorhult:
Great work @HugoDF! This will cleanup my markup a lot!

Issue #573, opened 2020-06-15
Bug in component dispatch?


thiagomajesk:
Hello everyone!

I was doing some work today and came across this weird issue with `$dispatch`. I don't know exactly why but it doesn't seem to be working properly for me. I have an example here:
https://codepen.io/thiagomajesk/pen/MWKjeMZ.

I'm trying to dispatch some events between components and I can't figure out what is happening.

Events are not being captured:

```html
<section x-data="{items: ['foo']}">
  <p @foo-event="alert($event.detail.items)" x-text="items"></p>
  <button @click="$dispatch('bar-event', {items: items})">Transfer</button>
</section>

<section x-data="{items: ['bar']}">
  <p @bar-event="alert($event.detail.items)" x-text="items"></p>
  <button @click="$dispatch('foo-event', {items: items})">Transfer</button>
</section>
```

Also, for some reason, the Alpine's README example only works if I declare the event in the root element (the second example simply doesn't work):

```html
<div x-data="" @custom-event="alert($event.detail.foo)">
  <button @click="$dispatch('custom-event', { foo: 'bar' })">Works</button>
</div>

<div x-data="">
  <div @custom-event="alert($event.detail.foo)"></div>
  <button @click="$dispatch('custom-event', { foo: 'bar' })">Doesn't Work</button>
</div>
```

ryangjchandler:
When listening for the event, you should add the `.window` modifier:

```
@custom-event.window=""
```

thiagomajesk:
Thanks a lot for the reply @ryangjchandler!

I failed to see that this is in fact already documented in the `x-on` section: https://github.com/alpinejs/alpine#x-on. 

Do you think that we could improve the documentation to include a note about that in the `$dispatch` section? It wasn't clear to me at first just focusing on the dispatching part.

ryangjchandler:
@thiagomajesk feel free to make a PR with some documentation improvements.

PR #574, opened 2020-06-15
Make more clear how to use $dispatch with sibling nodes

thiagomajesk:
https://github.com/alpinejs/alpine/issues/573#issuecomment-644284789

### DIFF
--- a/README.md
+++ b/README.md
@@ -589,7 +589,7 @@ If you need to access $event inside of a JavaScript function you can pass it in 
 </div>
 ```
 
-**Note on Event propagation**
+**Note on Event Propagation**
 
 Notice that, because of [event bubbling](https://en.wikipedia.org/wiki/Event_bubbling), when you need to capture events dispatched from nodes that are under the same nesting hierarchy, you'll need to use the [`.window`](https://github.com/alpinejs/alpine#x-on) modifier:
 
@@ -602,12 +602,11 @@ Notice that, because of [event bubbling](https://en.wikipedia.org/wiki/Event_bub
 <div>
 ```
 
-Since both elements are nested under `div`, this won't work. 
-The `custom-event` is dispatched from the `button` and will propagate to its ancestor, the `div`.
+> This won't work because when `custom-event` is dispatched, it'll propagate to its common ancestor, the `div`.
 
-**Dispatching to components**
+**Dispatching to Components**
 
-You can take advantage of the previous technique to make your components talk to each other:
+You can also take advantage of the previous technique to make your components talk to each other:
 
 **Example:**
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -589,6 +589,20 @@ If you need to access $event inside of a JavaScript function you can pass it in
 </div>
 ```
 
+Notice that, because of [event bubling](https://en.wikipedia.org/wiki/Event_bubbling), you'll need to use the [`.window`](https://github.com/alpinejs/alpine#x-on) modifier if you are trying to propagate events between sibling nodes:
```
ryangjchandler, line 592: bubbling*

ryangjchandler, line 592: This should probably say "between components" instead of "sibling nodes" since that's really what is happening.

thiagomajesk: I actually haven't indicated those are two components... 
Piggybacking off of your previous explanation, I think "sibling nodes" is broader than "between components" because there are cases where you are not in another component and this problem may still arise. Using a modified version of the previous example which won't work:

```html
<div>
    <div @custom-event="console.log($event.detail.foo)">
        <button @click="$dispatch('custom-event', { foo: 'bar' })">
        <!-- When clicked, will console.log "bar" -->
    </div>
</div>
```
This, for instance, was something I tried and was confused about why it wouldn't work. 
But once I understood bubbling was taking place, it was easier to reason about (even for component usage).

SimoTod: It's not only about sibling, though. You could have a different component on a different point of your DOM which is not a direct ancestor of the other component and you would still need the window modofier.

thiagomajesk: Well noted @SimoTod. 
Any suggestions on how to change this without making it only about distinct components?
I want to make it clear enough to cover both cases, which is precisely what was my struggle.
I'm not against using @ryangjchandler suggestion if it's your preference though.

SimoTod: You can try to describe the opposite: you can say that you can capture events dispatched by node inside your component and, for all other cases you need to use the window or document modifier.

thiagomajesk: @SimoTod and @ryangjchandler I think I captured the right amount of information. 
Could you review the update? I'm closing this conversation because it's outdated.

thiagomajesk:
@SimoTod, @ryangjchandler I've updated the PR with some tweaks.

README.md
SimoTod: I find the first paragraph a bit confusing but I'm not a native speaker so I could be wrong so I'll leave it with the others to review. (I'm commenting here since I don't want to create noise on the main thread).

thiagomajesk: @SimoTod I'm also not a native speaker. 
If you have a suggestion on how to improve it I'll gladly take it 😄.

calebporzio:
Thanks for the helpful copy @thiagomajesk !

PR #577, opened 2020-06-16
Update babel/preset-env and fix audit warnings

SimoTod:
The current build is failing. It seems there's some incompatibilities between the version of babel/preset-env we use and the latest version of node. This happened to other libraries as well (For example Laravel mix, see https://github.com/JeffreyWay/laravel-mix/issues/2383). 🤷 

Updating babel/preset-env to the most recent version fixes the build process.

I've also run 'npm audit fix' since there were a lot of low lever security warnings
`found 1523 low severity vulnerabilities in 961 scanned packages`
All fixed without any BCs.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,12 +14,12 @@
             }
         },
         "@babel/compat-data": {
-            "version": "7.8.6",
-            "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.8.6.tgz",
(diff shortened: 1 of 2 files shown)


carlmjohnson:
Looks like I opened the same PR, #581. I'll close in favor of you, because yours also does security audit stuff.

carlmjohnson:
I explain in #580 that the core of the problem is that Node 12.17 turned on a JS module thing that breaks Babel. So solutions are either to pin Node at 12.16 or upgrade Babel.

PR #578, opened 2020-06-16
Remove flatMap since not supported by Edge 18

### LINKED ISSUE #576: flatMap usage cause issue on Edge
There is a `flatMap` use case in `getXAttrs` function which cause Edge to stop working on current master. The use case implemented when Caleb was working on `x-bind`. So i think it is still `WIP`. Just want to put it here to remind.

SimoTod:
flatMap is not supported in Edge 18 (market share 2.8%) but that version does support the module/nomodule syntax so it won't use the IE11 version of Alpine providing a polyfill.

As a result, the upcoming release would break when used with old Edge versions.

It's safer to use reduce in this case: `FlatMap` is just syntax sugar for `map+flat` and `flat` is syntax sugar for `reduce+concat` so the following are equivalent

```javascript
        .flatMap(i => {
            if (i.type === 'spread') {
                let directiveBindings = saferEval(i.expression, component.$data)

                return Object.entries(directiveBindings).map(([name, value]) => parseHtmlAttribute({ name, value }))
            } else {
                return i
            }
        })
```
```javascript
        .map(currentValue => {
            if (currentValue.type === 'spread') {
                let directiveBindings = saferEval(currentValue.expression, component.$data)

                return Object.entries(directiveBindings).map(([name, value]) => parseHtmlAttribute({ name, value }))
            } else {
                return currentValue
            }
        })
        .reduce((accumulator, currentValue) => accumulator.concat(currentValue), [])
```
```javascript
        .reduce((accumulator, currentValue) => {
            if (currentValue.type === 'spread') {
                let directiveBindings = saferEval(currentValue.expression, component.$data)

                return accumulator.concat(Object.entries(directiveBindings).map(([name, value]) => parseHtmlAttribute({ name, value })))
            } else {
                return accumulator.concat(currentValue)
            }
        }, [])
```

Closes #576

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -101,15 +101,15 @@ export function getXAttrs(el, component, type) {
     return Array.from(el.attributes)
         .filter(isXAttr)
         .map(parseHtmlAttribute)
-        .flatMap(i => {
-            if (i.type === 'spread') {
-                let directiveBindings = saferEval(i.expression, component.$data)
+        .reduce((accumulator, currentValue) => {
+            if (currentValue.type === 'spread') {
+                let directiveBindings = saferEval(currentValue.expression, component.$data)
 
-                return Object.entries(directiveBindings).map(([name, value]) => parseHtmlAttribute({ name, value }))
+                return accumulator.concat(Object.entries(directiveBindings).map(([name, value]) => parseHtmlAttribute({ name, value })))
             } else {
-                return i
+                return accumulator.concat(currentValue)
             }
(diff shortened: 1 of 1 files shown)


SimoTod:
Note: Test will fail until we merge #577

calebporzio:
Oh thanks @SimoTod! I actually just refactored it out myself though... sorry. Thanks again!

SimoTod:
No worries

PR #581, opened 2020-06-17
Fix build by upgrading Babel

### LINKED ISSUE #580: Broken build
https://github.com/alpinejs/alpine/runs/777172957

I think I know why the build is broken. The log says,

> [0] [!] (plugin babel) Error: Cannot find module '@babel/compat-data/corejs3-shipped-proposals'

I had a similar issue. I believe it was caused by Node 12.17 turning on modules in a way that breaks Babel. https://nodejs.org/en/blog/release/v12.17.0/ Honestly, the whole thing is such a typical JS dependency breaking 💩 show (why does v12.17 make a breaking change???), it makes me never want to use JS again (or at least switch to Deno), but it has a pretty simple fix in my experience. If you just upgrade Babel, it gets past whatever broken in Node 12.17.

carlmjohnson:
Fixes #580 .

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,37 +14,210 @@
             }
         },
         "@babel/compat-data": {
-            "version": "7.8.6",
-            "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.8.6.tgz",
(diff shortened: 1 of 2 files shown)


carlmjohnson:
CI tests pass. :-)

carlmjohnson:
Closing in favor of #577.

PR #582, opened 2020-06-17
[Proof of concept]: Alpine Lite

HugoDF:
Create `alpine-lite.js` weighing in under 4kb gzipped (~11kb minified)

It turns out that rollup is very good at dead code elimination so we can toggle features/code based on an variables defined in the rollup config.

Disables the following features:
- x-for
- x-if
- x-model
- x-spread
- boolean attribute detection
- transitions
- `class` attribute object/array binding
- `.away`, `.debounce` modifiers 
- observable-membrane (uses buggy hand-rolled observable but that's probably not an issue if we don't have x-for to use arrays of things)

Other things in this PR:

- refactors the IE11-Only code to use `process.env.IE11_only` (removes the need for strip-code rollup plugin)
- removes unused rollup plugins (terser was already unused, strip-code is newly unused)

**Alternative Approach(es) & Questions**

- This PR uses a single variable `LITE` to denote whether a feature is in the Lite version or not, we could go another way and add per feature, "feature toggles" eg. `X_FOR_ENABLED`, `X_IF_ENABLED`, `X_MODEL_ENABLED`, `X_TRANSITIONS_ENABLED` etc, which would mean that people can create custom builds of Alpine.js with the features they need, this would pave the way towards a Tailwind CLI style tool which can generate custom builds given a set of enabled/disabled features
  - should we go for this more granular approach?
  - should we go for `X_IF_ENABLED` or `X_IF_DISABLED` type of flags?
- What other features should we disable?
- Which currently disabled features should we re-enable?

### DIFF
--- a/dist/alpine-ie11.js
+++ b/dist/alpine-ie11.js
@@ -5515,7 +5515,7 @@
 
     return new Function(['dataContext'].concat(_toConsumableArray(Object.keys(additionalHelperVariables))), "with(dataContext) { ".concat(expression, " }")).apply(void 0, [dataContext].concat(_toConsumableArray(Object.values(additionalHelperVariables))));
   }
-  var xAttrRE = /^x-(on|bind|data|text|html|model|if|for|show|cloak|transition|ref|spread)\b/;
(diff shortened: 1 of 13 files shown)


calebporzio:
Holding off on this for now thanks @HugoDF !

PR #597, opened 2020-06-22
feature(add support for .camel binding modifier)

### LINKED ISSUE #594: x-bind converts attributes to lowercase
When given `:viewBox`, x-bind renders the attribute to lowercase which causes the issue for SVG design.

info

ryangjchandler:
Closes #594.

This PR adds a new `.camel` modifier that can be used alongside `x-bind`, in the same fashion as Vue's modifier.

A place where this would be needed is on `<svg>` elements when binding the `viewBox` attribute - using the `.camel` modifier, the attribute will be set `viewBox` instead of `viewbox`.

This does require people to bind using a kebab case syntax, so it would need to be `x-bind:view-box.camel` instead of `x-bind:viewBox.camel`, the behaviour matches that of Vue.

CodePen here using my fork: https://codepen.io/ryangjchandler/pen/PoZprNx

/cc @HugoDF


src/directives/bind.js
```diff
@@ -72,7 +74,7 @@ export function handleAttributeBindingDirective(component, el, attrName, express
     }
 }
 
-function setIfChanged(el, attrName, value) {
+function setIfChanged(el, attrName, value, namespaced = false) {
```
SimoTod, line 77: Great PR, I think this line slipped in from a different feature/implementation since it doesn't seem to be used.

ryangjchandler: Good spot! This was being used in my original implementation, but it isn't actually needed.

calebporzio:
Is there any way for us to just preserve the casing of the bound attribute? Without needing the extra .camel?

HugoDF:
@calebporzio with `setAttributeNS` with the right spec it's possible I think, it's the solution I proposed https://github.com/alpinejs/alpine/issues/594#issuecomment-647364328

ryangjchandler:
@calebporzio It's possible, but I felt that adding a new modifier would be more extensible and cover more use cases.

calebporzio:
Yeah, I suppose it just feels like something that should be taken care of automatically, although it WILL be easier to just merge this.

Is there a way to use .setAttributeNS and NOT have to check if we're inside an SVG, I imagine that would be expensive overhead to add to every other bind operation

ryangjchandler:
@calebporzio I can take a look at using `setAttributeNS` instead, but I think this `.camel` modifier could also be used for `x-on:` where the event name needs to be camel cased. An example is https://github.com/alpinejs/alpine/discussions/626

HugoDF:
> @calebporzio I can take a look at using `setAttributeNS` instead, but I think this `.camel` modifier could also be used for `x-on:` where the event name needs to be camel cased. An example is #626

that makes sense actually, this PR doesn't add the support for event bindings right?

ryangjchandler:
@HugoDF Not currently, but if @calebporzio likes the idea I'll update the PR and add some stuff to the docs.

ryangjchandler:
Hey all (@calebporzio mainly) - I've updated this PR to add support for the `.camel` modifier to both `x-on:` and `x-bind:` with a simple test for both, as well as README updates.

src/directives/bind.js
SimoTod: `namespaced` is back?

ryangjchandler:
Avoid merging this one - the merge with master seems to have undone a commit of mine. Will sort this tomorrow.

ryangjchandler:
Ready for review & merge.

calebporzio:
Thanks Ryan!

fnick851:
Is this modifier documented anywhere? On the x-bind page (https://alpinejs.dev/directives/bind) I don't see it.

PR #598, opened 2020-06-22
README v.2.4.1 Russian translation

evgeniyPP:
Full translation of documentation for version 2.4.1.

I saw that before me @demiurge-ash had already translated the part about x-spread and I sincerely wanted to leave it in his translation, but, unfortunately, he translated it very verbatim. I had to rewrite.


calebporzio:
Thank you so much!

PR #600, opened 2020-06-24
Fix: default x-data to '{}' to in mutation observer initialiser

### LINKED ISSUE #593: Naked x-data exception.
A naked `x-data` attributes, say `<div x-data></div>` results in an unmatched `}` exception being generated inside `saferEval`. I'm not sure if I cut/paste this from example code somewhere, decided on my own that it would be treated the same way as `x-data="{}"` or it was just a mistake on my part but it wasn't obvious where in my HTML the issue was.

Would be worth catching the error beforehand and explaining the naked x-data isn't possible, setting a default and/or catching the eval as well if there are other syntax errors in the evaled `$expression`?

HugoDF:
Closes #593


calebporzio:
Thanks @HugoDF!

PR #603, opened 2020-06-25
Fix x-spread example

### LINKED ISSUE #602: X-Spread not working
I was trying to test out the `x-spread` functionality as it is shown in the readme, but it does not work for me. Either I'm missing something which is causing it to not work. Or there is a bug or the documentation needs to be fixed with a working example.

I have copied the example from the readme, which can be seen here:
https://codepen.io/ericp_mrel/pen/XWXggqj

computerfr33k:
Fixes #602 

The x-spread example code had a typo with x-data where it was missing '()' at the end of the function name.

### DIFF
--- a/README.md
+++ b/README.md
@@ -529,7 +529,7 @@ These behave exactly like VueJs's transition directives, except they have differ
 ### `x-spread`
 **Example:**
 ```html
-<div x-data="dropdown">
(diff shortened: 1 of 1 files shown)


calebporzio:
Thanks!

PR #604, opened 2020-06-25
Avoid unnecessary loops

bekliev:
There's no need to filter the `directives` array if type parameter is not passed into the `getXAttrs` function


bekliev:
Moved into this PR: https://github.com/alpinejs/alpine/pull/607

PR #606, opened 2020-06-25
Implementing Clean Code principles

bekliev:
And maybe DRY, KISS, SOLID principles too.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -366,27 +366,32 @@ export function transitionClassesOut(el, component, directives, hideCallback) {
 export function transitionClasses(el, classesDuring, classesStart, classesEnd, hook1, hook2) {
     const originalClasses = el.__x_original_classes || []
 
+    // To not remove classes that were in the original class attribute
+    const getRidOfOriginalClasses = list => list.filter(i => !originalClasses.includes(i))
+
+    const add = list => el.classList.add(...list)
+    const remove = list => el.classList.remove(...getRidOfOriginalClasses(list))
+
     const stages = {
         start() {
-            el.classList.add(...classesStart)
+            add(classesStart)
         },
         during() {
-            el.classList.add(...classesDuring)
+            add(classesDuring)
         },
         show() {
             hook1()
         },
         end() {
-            // Don't remove classes that were in the original class attribute.
-            el.classList.remove(...classesStart.filter(i => !originalClasses.includes(i)))
-            el.classList.add(...classesEnd)
+            remove(classesStart)
+            add(classesEnd)
         },
         hide() {
             hook2()
         },
         cleanup() {
-            el.classList.remove(...classesDuring.filter(i => !originalClasses.includes(i)))
-            el.classList.remove(...classesEnd.filter(i => !originalClasses.includes(i)))
+            remove(classesDuring)
+            remove(classesEnd)
         },
     }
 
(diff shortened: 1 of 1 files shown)


ryangjchandler:
Hi @bekliev - thanks for opening this PR. I know that it's still a WIP but I'm wondering what benefits this brings to the project? Refactoring is nice but can sometimes make the logic harder to follow, micro optimisations aren't always the best. 

I'll keep an eye on the changes as you make them.

bekliev:
@ryangjchandler totally agree with you in this:
> Refactoring is nice but can sometimes make the logic harder to follow, micro optimisations aren't always the best.

I also think if too much identical logic in the code then it's better to reuse it to be less bugprone in the future.
In case somebody changes the implementation one of the duplicated code to fix some bug or introduce new feature and doesn't know/remeber that the same logic has to be changed in other places then it could lead to bugs that hard to find.

What do you think?

ryangjchandler:
@bekliev That's exactly what all of the tests are for - to catch a bug and ensure any fix is applied accordingly. If the bug isn't being covered by the current test suite, it should be added before the fix to reproduce the problem as closely as possible.

The other thing with PRs that refactor core logic is that people are likely to disagree on where things should go, what they should be called. It's a big pain point of open source software.

bekliev:
> The other thing with PRs that refactor core logic is that people are likely to disagree on where things should go, what they should be called, etc.

Let's leave the naming thing to you (of the hardest problem in programming IMO)  - you decide the namings )

All I care is clean and easy to read/undesrtand code. What the author(s) wanted to do "here" and "there". You can always point out If I made if harder to understand.

So... will you give it a chance?

ryangjchandler:
Yeah, I'm not an original author - that credit goes to Caleb. I also can't merge the PR in, only close it.

I'll keep an eye on the changes you make, but ultimately, it's down to Caleb whether it gets merged or not.

KevinBatdorf:
@bekliev  `getRidOfOriginalClasses` seems reversed. Maybe you want `transitionClasses` or `transitionClassesOnly` for that variable name? 

The original classes are what you want to keep.

> people are likely to disagree on where things should go, **what they should be called**, etc.

😄

HugoDF:
From what I've seen in the PRs, Caleb will only merge PRs where there's a clear benefit to the end users of Alpine (eg. better performance, smaller bundle size), refactoring beyond a couple of lines without any improvement on those fronts is unlikely to get merged.

Also, just an FYI there has been some talk of trying to extract the transition logic out, so not sure particularly useful to starting with that.

bekliev:
@HugoDF 
> Caleb will only merge PRs where there's a clear benefit to the end users of Alpine (eg. better performance, smaller bundle size), refactoring beyond a couple of lines without any improvement on those fronts is unlikely to get merged.

But what about Developer Experience and clean code? About making code more understandable?

I **do agree** with Caleb mindset of thinking about User Experience, I'm also a big fan on UX thing. But I also believe that we have to care about ourselves too while **no harm** to end users.


---
> Also, just an FYI there has been some talk of trying to extract the transition logic out, so not sure particularly useful to starting with that.

Thanks for advice! Gonna leave transition logic till it's done.

calebporzio:
Thanks for pitching in @bekliev! I agree those bits could be cleaned up a bit and I appreciate the contribution.

However, the benefit this PR brings doesn't outweigh the added risk of pulling in code I didn't write myself to be frank.

Thanks again.

PR #607, opened 2020-06-25
WIP boosting performance

bekliev:
work in progress

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -58,22 +58,22 @@ const Alpine = {
         const observer = new MutationObserver((mutations) => {
             if (this.pauseMutationObserver) return;
 
-            for (let i=0; i < mutations.length; i++){
-                if (mutations[i].addedNodes.length > 0) {
-                    mutations[i].addedNodes.forEach(node => {
-                        // Discard non-element nodes (like line-breaks)
-                        if (node.nodeType !== 1) return
+            mutations.forEach(mutation => {
+                if (mutation.addedNodes.length === 0) return;
 
-                        // Discard any changes happening within an existing component.
-                        // They will take care of themselves.
-                        if (node.parentElement && node.parentElement.closest('[x-data]')) return
+                mutation.addedNodes.forEach(node => {
+                    // Discard non-element nodes (like line-breaks)
+                    if (node.nodeType !== 1) return
 
-                        this.discoverUninitializedComponents((el) => {
-                            this.initializeComponent(el)
-                        }, node.parentElement)
-                    })
-                }
-              }
+                    // Discard any changes happening within an existing component.
+                    // They will take care of themselves.
+                    if (node.parentElement && node.parentElement.closest('[x-data]')) return
+
+                    this.discoverUninitializedComponents((el) => {
+                        this.initializeComponent(el)
+                    }, node.parentElement)
+                })
+            })
         })
 
         observer.observe(targetNode, observerOptions)
--- a/src/index.js
+++ b/src/index.js
@@ -51,7 +51,7 @@ const Alpine = {
 
         const observerOptions = {
             childList: true,
-            attributes: true,
+            attributeFilter: ['x-data'],
             subtree: true,
         }
 
--- a/src/utils.js
+++ b/src/utils.js
@@ -182,7 +182,7 @@ export function transitionIn(el, show, component, forceSkip = false) {
 
         transitionHelperIn(el, modifiers, show)
     // Otherwise, we can assume x-transition:enter.
-    } else if (attrs.filter(attr => ['enter', 'enter-start', 'enter-end'].includes(attr.value)).length > 0) {
+    } else if (attrs.some(attr => ['enter', 'enter-start', 'enter-end'].includes(attr.value))) {
         transitionClassesIn(el, component, attrs, show)
     } else {
     // If neither, just show that damn thing.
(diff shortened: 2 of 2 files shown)


src/utils.js
```diff
@@ -110,10 +110,9 @@ export function getXAttrs(el, component, type) {
         directives = directives.concat(Object.entries(spreadObject).map(([name, value]) => parseHtmlAttribute({ name, value })))
     }
     
-    // If no type is passed in for filtering, bypass filter
-    if (!type) return directives;
+    if (type) return directives.filter(i => i.type === type)
```
ryangjchandler, line 113: I don't think should change. The unhappy path is tested first and returns early, which is the preferred approach.

bekliev: First I thinked same way, but after a moment realized that `type` parameter of the function [getXAttrs](https://github.com/alpinejs/alpine/pull/607/commits/f761f02bccf476f6f13d7a459dee8c704aa627c8#diff-2b4ca49d4bb0a774c4d4c1672d7aa781R100) is optional accordingly to these use-cases in `components.js` file:
1. https://github.com/alpinejs/alpine/blob/a0fab06379b5427df82a8a32e495e373f51c7a16/src/component.js#L165
2. https://github.com/alpinejs/alpine/blob/a0fab06379b5427df82a8a32e495e373f51c7a16/src/component.js#L227
3. https://github.com/alpinejs/alpine/blob/a0fab06379b5427df82a8a32e495e373f51c7a16/src/component.js#L243

In case the `type` attribute was required then the happy/main case would be the filtering by type.
But our case is optional type and cause of it happy/main case is to just get `x-*` attributes.

src/index.js
```diff
@@ -58,22 +58,22 @@ const Alpine = {
...
+                    if (node.nodeType !== 1) return
+
+                    // Discard any changes happening within an existing component.
+                    // They will take care of themselves.
+                    if (node.parentElement && node.parentElement.closest('[x-data]')) return
+
+                    this.discoverUninitializedComponents((el) => {
+                        this.initializeComponent(el)
+                    }, node.parentElement)
+                })
```
bekliev, line 75: Much of it is just indentation changes.
Logic is not changed

src/index.js
```diff
@@ -58,22 +58,22 @@ const Alpine = {
...
-                        // They will take care of themselves.
-                        if (node.parentElement && node.parentElement.closest('[x-data]')) return
-
-                        this.discoverUninitializedComponents((el) => {
-                            this.initializeComponent(el)
-                        }, node.parentElement)
-                    })
-                }
-              }
+            mutations.forEach(mutation => {
```
bekliev, line 61: forEach is waaaay faster than our old-lovely and imperative for-loop.

Proof ([benchmark](https://www.measurethat.net/Benchmarks/Show/3862/array-loops-for-vs-for-in-vs-for-of-vs-foreach-vs-map-f))
image

src/index.js
```diff
@@ -58,22 +58,22 @@ const Alpine = {
         const observer = new MutationObserver((mutations) => {
             if (this.pauseMutationObserver) return;
 
-            for (let i=0; i < mutations.length; i++){
-                if (mutations[i].addedNodes.length > 0) {
```
bekliev, line 62: forEach is waaaay faster than our old-lovely and imperative for-loop.

Proof ([benchmark](https://www.measurethat.net/Benchmarks/Show/3862/array-loops-for-vs-for-in-vs-for-of-vs-foreach-vs-map-f))
image

src/index.js
```diff
@@ -51,7 +51,7 @@ const Alpine = {
 
         const observerOptions = {
             childList: true,
-            attributes: true,
+            attributeFilter: ['x-data'],
```
bekliev, line 54: We need to initialize only elements with `x-data` attribute on them to initialize the component.

By setting the option `attributeFilter` we're incredibly reducing complexity of the MutationObserver by triggering the callback only when we confident that element has `x-data` attribute and then we run nested loops.

We should avoid executing nested loops as much as we can especially when we they are useless.

src/index.js
bekliev: As you can see `for-of` loop is faster than `forEach` in Chrome.
But in Firefox `for-of` even slower than `for-loop` (see screenshot below)

So I've used `forEach` cause of that.

image

bekliev:
Waiting for feedback

carlmjohnson:
Do you have a benchmark of some sort? Doing an “optimization” without a benchmark is just guessing.

bekliev:
> Do you have a benchmark of some sort? Doing an “optimization” without a benchmark is just guessing.

I would be happy to do that, but don't know how to do that with alpine...
Is there some kinda boilerplate code I can use? Maybe some older benchmarks?
In benchmark for testing especially MutationObserver - there has to be some updates in the DOM itself (like appending into the DOM new html-tags with `x-data` attributes)

ryangjchandler:
@bekliev I typically benchmark "improvements" by looking at what components would be affected by the changes. For example, the `forEach` `for..of` benchmarks are for a generic loop - how much impact does that have on an *actual* Alpine component being mutated?

bekliev:
> @bekliev I typically benchmark "improvements" by looking at what components would be affected by the changes. For example, the `forEach` `for..of` benchmarks are for a generic loop - how much impact does that have on an _actual_ Alpine component being mutated?

Exactly! Are there some well-tested benchmarks for performance-checks of alpine components? Or there are not?

SimoTod:
@bekliev We don't have any benchmarks for now.

My personal opinion, It may be worth to split this PR in small micro PRs. 
There was a dev who sent a PR with a lot of changes in the past and it got rejected because Caleb did not agree with half of them. Small PR will make life easier for Caleb when he needs to review and increase the likelihood of some of your optimisation being merged.

For example, i can see 3 different areas here:
- `filter` to `some` in transitions
- getXAttrs
- mutation observer

bekliev:
@SimoTod oh, nice advice - will do in weekends! Thanks!

bekliev:
@SimoTod done:
https://github.com/alpinejs/alpine/pull/629 (mutation observer)
https://github.com/alpinejs/alpine/pull/630 (transitions)
https://github.com/alpinejs/alpine/pull/631 (getXAttrs)

Issue #612, opened 2020-06-28
2.4 transition behavior regression


KevinBatdorf:
Here's the setup:

Alpine 2.3.5: https://codepen.io/KevinBatdorf/pen/gOPxojN
Alpine 2.4.1: https://codepen.io/KevinBatdorf/pen/c89c2b1ea994646bca5cd254b4152611

Notice the flickering animations on 2.4.1

~I have a fix ready but will see if there's a test I can add.~

loweoj:
Adding another example of this bug. When alpine is pinned to 2.4.1 you can see error flashes on the background when typing continuously into the input for a little while - https://codepen.io/loweoj/pen/wvMreYx

PR #613, opened 2020-06-28
WIP: [RFC] Update glitches with transitons

KevinBatdorf:
This seems to be a pretty reliable way to verify the transitions, but I'm not sure if it's the correct approach.

📹 [Here's a video demo on YouTube](https://www.youtube.com/watch?v=LU7Vdg8BW04&feature=youtu.be)

I marked this as WIP because I've yet to do the following:
1. Test using x-if
1. Test with other transition situations
1. Add actual tests 🏎️ 

Would like to get some comments first to make sure this is on the right direction before investing too much time. 🚀 

👍 I've included the demo in the examples folder just for convenience, and to see the before/after

Fixes https://github.com/alpinejs/alpine/issues/612

### DIFF
--- a/dist/alpine-ie11.js
+++ b/dist/alpine-ie11.js
@@ -5699,25 +5699,6 @@
     return rawValue;
   }
 
-  function registerTransitionListeners(el) {
-    var _this8 = this;
-
-    el.addEventListener('transitionrun', function () {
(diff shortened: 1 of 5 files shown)


KevinBatdorf:
Discussed this is in the discord. This won't work because transitions need to be cancelled. Closing for now!

PR #619, opened 2020-06-29
Feature: `.passive` modifier for passive event listeners

### LINKED ISSUE #559: x-on support for passive event listeners
I would love to see support for [passive event listeners](https://developers.google.com/web/updates/2016/06/passive-event-listeners). I imagine it working like:
```
<div @mousedown.passive="interactive = true">
```

It looks like it should be pretty easy to implement as `x-on` is already creating an event listener for the element.
https://github.com/alpinejs/alpine/blob/5181242588436f58c3ea19cd0e45d6dcbe29f853/src/directives/on.js#L22

_Originally posted by @alex-page in https://github.com/alpinejs/alpine/discussions/557_

HugoDF:
Closes #559 

- adds `.passive` modifier which populates the `options` object for listener handlers
- adds tests to check that a listener is passive (ie. e.preventDefault() won't do anything)
- document `.passive` x-on modifier

IE11:
- switches custom event polyfill from `custom-event-polyfill` to `events-polyfill` (CustomEvent constructor)
- adds listener options polyfill

Note:
- since we pass a listener `options` now we could leverage `once` and `capture` (might allow us to get rid of some of the custom `once` modifier code), @calebporzio @ryangjchandler let me know what you think about that
  - see https://github.com/HugoDF/alpine/compare/feat-passive-listeners...HugoDF:refactor-native-once which ditches most of the `once` custom code and uses the `once` option instead (all tests pass)


ryangjchandler:
@HugoDF Looking at this now, does this new polyfill also support the `once` option?

HugoDF:
> @HugoDF Looking at this now, does this new polyfill also support the `once` option?

seems to, but I haven't actually taken it for a spin on IE11

calebporzio:
Thanks @HugoDF !

PR #623, opened 2020-06-30
Feature - Transition improvements

### LINKED ISSUE #376: x-show with x-transition with quick event may have weird behavior
If you fire fast events to show/hide elements with x-transition it ends up hiding. (sorry if it's too vague, i made a jsfiddle to help understand)

https://jsfiddle.net/pgn6om87/

If you see this fiddle example and move the mouse hover/out the button quickly you may reproduce the error.

### LINKED ISSUE #612: 2.4 transition behavior regression

### LINKED ISSUE #643: Transitions broken in v2.4.1
I have an app with a tree that uses `x-show.transition` to show/hide nodes.

This has worked great in "all" AlpineJS versions, but when upgrading from v2.4.0 to v2.4.1 the transitions happens on any update, even if the `x-show` boolean changes or not, which makes for a very busy tree.

I understand that v2.4.1 was a reversal of a commit that introduced a related issue, but the reversal does not look correct to me (in that they do not match that great):

https://github.com/alpinejs/alpine/pull/543/files
https://github.com/alpinejs/alpine/commit/91b057d4c4295c4edf459716e800c7fbffd0e211

SimoTod:
Closes #612 
Closes #376
Closes #643

v2.4.1 fixed an issue with x-show directives inside x-for but introduced a regression experienced by a few people in the discord channel and affecting couple of demos:
show animations are re triggered every time a component is repainted so, depending on the effect applied, it could result in element flickering or similar. 

2.4.1 also still represented the issue when show and hide overlap (See https://codepen.io/SimoTod/full/XWXzKaQ and try to double click on toggle: show variable will be set to true but the x-show div will be hidden and, from that point, the whole logic will be broken unless you refresh the page).

This PR aims to reinstantiate the fix in 2.4.0 but with support for x-show inside x-for and a few more changes:
- no logic added to show.js (which was breaking the memory card example)
- all logic in utils.js which should be easier to follow and understand
- sensible variables

~~If this PR is accepted, there will be a follow up PR fixing some additional edge cases with request animation frame. 
I kept them separated so it's easier to review and the second bunch of changes doesn't block this first PR which is probably more important.~~

Memory card game using fixed version: https://codepen.io/SimoTod/full/yLePaLW
Overlapping transition using fixed version: https://codepen.io/SimoTod/pen/gOPXwbv

-- NOTE -- 
`test('transition out not called when item is already hidden')` and `test('transition in not called when item is already visible)` have been updated to catch the regression in 2.4.1. Those tests were written with the original reactivity layer, the salesforce membrane by default does not trigger a repainting if value hasn't changed so those 2 tests were not valid anymore.

I've also added a test to cover the regression in 2.4.0 (it was a bit tricky to strip it down to a minimal state but it happens when x-show directives inside x-for have an x-bind:style directive).


src/utils.js
```diff
@@ -407,20 +431,40 @@ export function transition(el, stages) {
...
 export function isNumeric(subject){
     return ! isNaN(subject)
 }
+
+// Thanks @vuejs
+// https://github.com/vuejs/vue/blob/4de4649d9637262a9b007720b59f80ac72a5620c/src/shared/util.js
+export function once(callback) {
+    let called = false
+
+    return function () {
```
ryangjchandler, line 464: missing arguments parameter?

SimoTod: nope, https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments#:~:text=The%20arguments%20object%20is%20a,first%20entry's%20index%20at%200%20.

SimoTod: that function is always without arguments anyway, I can pass an empty array if clearer

ryangjchandler: omg, I never knew about this weird arguments variable - cheers!

HugoDF: could use `function(...args)` instead but I think the transpiled output would be larger & it probably makes sense to do it the same way Vue does

muzafferdede:
Great work @SimoTod! 
Quick question, i have a case which i need child elements to finish transitions then parent element do its transition thing. ex:

```html
<div x-data="{show: true}">
  <p>shown: <span x-text="show"></span></p>
  <button @click="show = !show">Toggle</button>
  <div x-show.transition.duration.1000ms="show">
    PARENT
    <div x-show.transition.duration.5000ms="show">
      CHILD
    </div>
  </div>
</div>
```

What do you expect in this case to happen? child needs 5000ms to complete the transition but parent needs 1000ms. 

I am expecting child element finish its transition and parent start once all child element finished. Or am i wrong?

SimoTod:
I need to test it but I believe that, with the current code, they start at the same time when showing but they are sequential when hiding, since the resolve callback is separate for the show part in x-show.

muzafferdede:
Yeah, I was not trying to link it to this PR but want to know if i make sense to expect this behavior. Maybe we should consider this behavior in transitions. Will be great for the component patent - child reactivity

src/directives/show.js
```diff
@@ -24,12 +24,14 @@ export function handleShowDirective(component, el, value, modifiers, initialUpda
 
     const handle = (resolve) => {
         if (value) {
-            transitionIn(el,() => {
-                show()
-            }, component)
+            if(el.style.display === 'none' || el.__x_transition) {
```
SimoTod, line 27: This line removed in v2.4.1 caused a regression experienced by several users.
We need to queue the animation only if the element is hidden or there is a transition running.
We don't check the type of transition and we don't add additional logic here, the low level functions will take care of ensuring that duplicated transition are discarded

SimoTod:
@calebporzio a couple of guys on discord confirmed that it works correctly for them. 
It also fixes an issue mentioned in #641
Feel free to have a spin when you have time, thanks.

calebporzio:
Thank you so much @SimoTod for all your hard work on this ❤️ ❤️ ❤️

PR #629, opened 2020-07-04
Boost mutation observer

bekliev:


src/index.js
```diff
@@ -58,22 +58,22 @@ const Alpine = {
         const observer = new MutationObserver((mutations) => {
             if (this.pauseMutationObserver) return;
 
-            for (let i=0; i < mutations.length; i++){
```
bekliev, line 61: forEach is waaaay faster than our old-lovely and imperative for-loop.

Proof ([benchmark](https://www.measurethat.net/Benchmarks/Show/3862/array-loops-for-vs-for-in-vs-for-of-vs-foreach-vs-map-f))
image

bekliev: As you can see `for-of` loop is faster than `forEach` in Chrome.
But in Firefox `for-of` even slower than `for-loop` (see screenshot below)

So I've used `forEach` cause of that.

image

src/index.js
```diff
@@ -51,7 +51,7 @@ const Alpine = {
 
         const observerOptions = {
             childList: true,
-            attributes: true,
+            attributeFilter: ['x-data'],
```
bekliev, line 54: We need elements with `x-data` attribute only - to initialize components.

By setting the option `attributeFilter` we're incredibly reducing complexity of the `MutationObserver` by triggering the callback only when we confident that element has `x-data` attribute.

You could say: "There are code that get elements with `x-data` attribute only:" https://github.com/alpinejs/alpine/blob/4aa31e2e4c24b4765389d3a63c461536033eb172/src/index.js#L39-L40
YES THERE IS!
But as you know JS is slow, and if we delegate this job to `MutationObserver` itself which is written in native code in all browsers (e.g. C++) then it's gonna check for that attribute's existence super-fast!

In `MutationObserver`'s callback we have nested loops `O(2*n)` - as you know they are slow too.
We should avoid executing nested loops as much as we can especially when they are useless.

ryangjchandler:
Interesting benchmark. I was always under the impression that `forEach` was slower since it has to run / invoke a function for each item.

src/index.js
SimoTod: There's another mutationObserver in component.js which would benefit from the same change.

bekliev: done in commit https://github.com/alpinejs/alpine/pull/629/commits/2d0ef071998021ce17b2217a7d8c4fed8770316c

ryangjchandler:
@bekliev - have you got any more test cases / benchmarks for these changes, preferably in the context of an Alpine component?

calebporzio:
Thank you for your contribution. I'm closing this because it changes a crucial part of Alpine's core and there aren't tangible enough benchmarks provided.

Thank you again.

PR #630, opened 2020-07-04
Boost checkings (filter > some)

bekliev:
Benchmark: https://www.measurethat.net/Benchmarks/Show/8658/alpinejs-speed-up-transitionin-transitionout

### DIFF
--- a/src/component.js
+++ b/src/component.js
@@ -282,7 +282,7 @@ export default class Component {
                 case 'if':
                     // If this element also has x-for on it, don't process x-if.
                     // We will let the "x-for" directive handle the "if"ing.
-                    if (attrs.filter(i => i.type === 'for').length > 0) return
+                    if (attrs.some(i => i.type === 'for')) return
 
                     var output = this.evaluateReturnExpression(el, expression, extraVars)
 
--- a/src/utils.js
+++ b/src/utils.js
@@ -182,7 +182,7 @@ export function transitionIn(el, show, component, forceSkip = false) {
 
         transitionHelperIn(el, modifiers, show)
     // Otherwise, we can assume x-transition:enter.
-    } else if (attrs.filter(attr => ['enter', 'enter-start', 'enter-end'].includes(attr.value)).length > 0) {
+    } else if (attrs.some(attr => ['enter', 'enter-start', 'enter-end'].includes(attr.value))) {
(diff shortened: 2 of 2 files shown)


src/utils.js
```diff
@@ -208,7 +208,7 @@ export function transitionOut(el, hide, component, forceSkip = false) {
             ? modifiers.filter((i, index) => index > modifiers.indexOf('out')) : modifiers
 
         transitionHelperOut(el, modifiers, settingBothSidesOfTransition, hide)
-    } else if (attrs.filter(attr => ['leave', 'leave-start', 'leave-end'].includes(attr.value)).length > 0) {
+    } else if (attrs.some(attr => ['leave', 'leave-start', 'leave-end'].includes(attr.value))) {
```
bekliev, line 211: In this cases it's faster to use `some` instead of `filter`

You can find the benchmark here:
https://www.measurethat.net/Benchmarks/Show/8658/alpinejs-speed-up-transitionin-transitionout

---

MDN about [some](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some)
> The some() method tests whether at least one element in the array passes the test implemented by the provided function. It returns a Boolean value.

---

MDN about [filter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter)
> The filter() method creates a new array with all elements that pass the test implemented by the provided function.

ryangjchandler:
Good work providing a benchmark - `some` could probably be used in a couple of other places if it hasn't already been done by others.

SimoTod:
Looks good. 👍

src/component.js
```diff
@@ -282,7 +282,7 @@ export default class Component {
                 case 'if':
                     // If this element also has x-for on it, don't process x-if.
                     // We will let the "x-for" directive handle the "if"ing.
-                    if (attrs.filter(i => i.type === 'for').length > 0) return
+                    if (attrs.some(i => i.type === 'for')) return
```
bekliev, line 285: @ryangjchandler after [your message](https://github.com/alpinejs/alpine/pull/630#issuecomment-653805505) just checked out the whole project with regex that matches:
```js
filter\(.*\)\.length
```

And found one 😃

calebporzio:
Nice! .some is more readable than .filter.length anyways to me.

PR #631, opened 2020-07-04
Boost getXAttrs

bekliev:


src/utils.js
```diff
@@ -110,10 +110,9 @@ export function getXAttrs(el, component, type) {
         directives = directives.concat(Object.entries(spreadObject).map(([name, value]) => parseHtmlAttribute({ name, value })))
     }
     
-    // If no type is passed in for filtering, bypass filter
-    if (!type) return directives;
+    if (type) return directives.filter(i => i.type === type)
 
-    return directives.filter(i => i.type === type)
+    return directives;
```
bekliev: After a moment realized that `type` parameter of the function [getXAttrs](https://github.com/alpinejs/alpine/pull/607/commits/f761f02bccf476f6f13d7a459dee8c704aa627c8#diff-2b4ca49d4bb0a774c4d4c1672d7aa781R100) is optional accordingly to these use-cases in `components.js` file:
1. https://github.com/alpinejs/alpine/blob/a0fab06379b5427df82a8a32e495e373f51c7a16/src/component.js#L165
2. https://github.com/alpinejs/alpine/blob/a0fab06379b5427df82a8a32e495e373f51c7a16/src/component.js#L227
3. https://github.com/alpinejs/alpine/blob/a0fab06379b5427df82a8a32e495e373f51c7a16/src/component.js#L243

Our case is: `type` param is optional and happy/main case is to just get `x-*` attributes.
In case the `type` parameter was required then the happy/main case would be the filtering-by-type.

So it's better to return optional case wrapped in optional condition itself and return happy/main case by default.

SimoTod:
Benchmark (no type): https://www.measurethat.net/Benchmarks/Show/8753/0/test-getxattrs
Benchmark (with type): https://www.measurethat.net/Benchmarks/Show/8754/0/test-getxattrs-with-type

They seem equivalent. I ran the benchmarks several times in chromes and sometimes the current version is quicker, sometimes is slower. The order of magnitudo is about the same. 

Am I doing something wrong @bekliev ?

bekliev:
I have a PC (Windows) and benchmark with no type - this PR always faster per [~10k ops/second](https://www.measurethat.net/Benchmarks/ShowResult/116230).
But benchmark with type - seems faster too but not that much [~4k ops/second](https://www.measurethat.net/Benchmarks/ShowResult/116233)

I have to admit that in first 2 or 3 tries of second benchmark (with type) I've got PR slower than original code.
And then I figured that I was clicking or scrolling after I hit the "Run tests" button especially while benchmark was working with first test (the PR version). And cause of that PR version performance dropped significantly and was slower per ~30k ops/second.
If I don't do anything after hitting the button then results always become in favor of the PR version.

SimoTod:
Weird, I'm not scrolling and I'm still getting alternate results (for example https://www.measurethat.net/Benchmarks/ShowResult/116235 and generally https://www.measurethat.net/Benchmarks/ListResults/8753). 
I assume it depends on what the browser is processing in the background at that giving time (extensions, other tabs, etc). 
The order of magnitude is still the same though, ~200K op/sec so if there is one, it's probably a micro optimisation.
Between the 3 PRs, it's probably the one with less benefits but good to see some benchmark and improvements. 👍

bekliev:
@SimoTod agree with you.
At least now it's better to read the code in my opinion 👀 

**before:**
https://github.com/alpinejs/alpine/blob/2af7688adbc326e3ed97daae19d036f7c1876599/src/utils.js#L113-L118

**after:**
https://github.com/alpinejs/alpine/blob/1f9ad11d95e641ec38b142c460a44d0c3241467e/src/utils.js#L113-L115

calebporzio:
Not sure if this is more performant or not, but I like style better so I'm merging it! Thanks.

PR #636, opened 2020-07-05
Ensure event object is passed to handlers defined as functions using x-spread

### LINKED ISSUE #627: $event object not accessible in event handlers added via x-spread
When adding event handlers via `x-spread` it doesn't appear that the native event object is accessible to the callback function:

```js
function Example(){
  return {
    button: {
      ['@click'](){
         // No way to access event object? No arguments are supplied
      }
    }
  }
}
```

I was expecting that the event handler would be called with the `$event` object as it's first argument - analogous to supplying a handler name in an `x-on`/`@` directive - for example `@click="myHandler"`

I opened a discussion topic about this a few days ago here: https://github.com/alpinejs/alpine/discussions/617 however after further investigation this seems to me that it should be raised as an issue.

In the discussion I also mentioned access to the `$dispatch` utility (and I guess by extension `$watch`) - That would be nice but at least it's consistent with the non-`x-spread` behaviour. However calling an event handler with the Event object as its first argument seems like a common (and expected) pattern.

Is this a bug? It would be really nice to be able to define event handlers via x-spread for more complex components but not having access to the Event object seems to limit its usefulness.

Simple jsfiddle test here: https://jsfiddle.net/9dLv24zp/16/

allmarkedup:
This pull request ensures that the `$event` special variable is passed to event handlers defined as functions when using x-spread, as an alternative to having to reference another function by name in the event handler definition.

Closes #627 - more context in that issue.

### DIFF
--- a/dist/alpine-ie11.js
+++ b/dist/alpine-ie11.js
@@ -884,59 +884,395 @@
 
   })();
 
-  // Polyfill for creating CustomEvents on IE9/10/11
+  var ApplyThisPrototype = (function() {
+    return function ApplyThisPrototype(event, target) {
+      if ((typeof target === 'object') && (target !== null)) {
+        var proto = Object.getPrototypeOf(target);
+        var property;
 
-  // code pulled from:
-  // https://github.com/d4tocchini/customevent-polyfill
-  // https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent#Polyfill
-
-  (function() {
-    if (typeof window === 'undefined') {
-      return;
-    }
-
-    try {
-      var ce = new window.CustomEvent('test', { cancelable: true });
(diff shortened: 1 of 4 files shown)


ryangjchandler:
Would you be able to compile the dist/ files and maybe add a test too?

HugoDF:
The file you want to add to is https://www.github.com/alpinejs/alpine/tree/master/test%2Fspread.spec.js

You probably just want to add a `click` handler that increments a variable or something using `x-spread`

allmarkedup:
Sorry chaps I was far too hasty with the initial commit - the test caught my mistake. All fixed now. If it needs anything else just let me know :-)

calebporzio:
This is great. Thanks for the clean fix with a test!

PR #640, opened 2020-07-07
docs(note-show-style-order)

### LINKED ISSUE #434: x-show gets overridden by :style
See [this 2.3.1 example](https://flems.io/#0=N4IgzgpgNhDGAuEAmIBcIB0ALeBbKIANCAGYCWMYaA2qAHYCGuEamO+RIsA9nYn6wA8YWACcyAB3gACMKNgBeADogc8CWFQB6LbCR0MAKzBJoZAG6iMdCPC0BzLFoZQJZG8eev3EAALmAJgwAZgwARi0kMjA7FzcbDFx3IzAVAD5BLRFxKTSlOnzBKPN86WkADwBaJAZ4BmUQYGksMlNUaXhRAFcIaQBfFVKKytgobgYAayGqsCxuAHcGgEIW00G6MtQYgE8YBoADHjHRdtFkffW8jekATW4u2TmuqH0AchlIXuZCyIsrzkgMAQZF4VHQADZUAEAiA+oR6EwWOgUpweHwIAJ0LD4SBGMxWBhYGAqMQ0fx4KxqFVRuMJgBdaRNKJgCRQBjbdp0XgQADc-XyAOgcHgILoYJAYVQAAZYXS+kA). The text shows up when it shouldn't. If you delete the `:style` line, or empty its value, or remove the backticks, it works as expected.

HugoDF:
Closes #434

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,6 +215,9 @@ If you wish to run code AFTER Alpine has made its initial updates to the DOM (so
 
 `x-show` toggles the `display: none;` style on the element depending if the expression resolves to `true` or `false`.
 
+> Note: when `style` is also being bound, the `x-show` directive should appear after the `x-bind:style` (or `:style`) directive to avoid issues, since x-show toggles `display: none` in the inline styles (which is also being manipulated by the style binding).
+> eg. `<div x-data="{ hide: true }" :style="`color: red`" x-show="!hide">You shouldn't see me</div>`
+
 **x-show.transition**
 
 `x-show.transition` is a convenience API for making your `x-show`s more pleasant using CSS transitions.


SimoTod:
I wish we could reconsider https://github.com/alpinejs/alpine/pull/236, to be fair (without the object support if Caleb is not sure). I think it's annoying that x-on:style overrides any values in pre existing style attributes.

HugoDF:
I can revive it

I think the biggest issue was added bulk from that but if we just do string merging it should be ok.

calebporzio:
Ok, I think Alpine should force x-show to override any :show bindings.

Is there a simple way to accomplish this without digging into the object syntax support (which I think we will likely end up adding after all)?

HugoDF:
Well the only bit from that PR that we would need is the "style rule merging", _or_ we can just order the operations so that `style` binding runs first and `x-show` second, always.

HugoDF:
Closing since we're going to order the things

PR #646, opened 2020-07-10
feature(remove x cloak last)

ryangjchandler:
This PR updates the removal of `x-cloak` to ensure that it is the very last thing that happens when initialising / updating elements in a component.

This one has caught quite a few people out since the README doesn't specify that directives are evaluated in the order they're defined, so just a small change so less people get stuck when `x-cloak` doesn't work as expected.


calebporzio:
Hey @ryangjchandler, can you expound on this a bit?

Also, why not put it at the end of "resolveBoundAttributes" so we aren't repeating ourselves, but even still, I'm unsure why this needs to exist on update.

I'm also unclear when the current setup doesn't work? For transitions and things that use setTimeout?

Thanks for clarifying.

ryangjchandler:
> 
> 
> Hey @ryangjchandler, can you expound on this a bit?
> 
> Also, why not put it at the end of "resolveBoundAttributes" so we aren't repeating ourselves, but even still, I'm unsure why this needs to exist on update.
> 
> I'm also unclear when the current setup doesn't work? For transitions and things that use setTimeout?
> 
> Thanks for clarifying.

Good point, it probably makes more sense putting it at the end of `resolveBoundAttributes`.

The problem here is `x-cloak`'s purpose in Alpine is to hide the element until it has been initialised by Alpine. Technically speaking, it does do that but it is down to the developer to ensure that they put `x-cloak` at the end of their HTML attributes.

An example:

```html
<div x-data="{ show: false }">
    <p x-cloak x-show="show"></p>
</div>
```

With the current setup, `x-cloak` will be evaluated / processed by Alpine first. This means that it will be removed before the `x-show` on the element has been processed, resulting in a flash of content still in some cases, especially if the `x-show` evaluation is defined after a load of other directives.

My changes will always make sure that `x-cloak` is removed last, so that the rather simplified example above doesn't cause any problems. It just means that the `x-cloak` functionality actually does what it says, rather than relying on the developer to always put that define that attribute on the element last.

ryangjchandler:
Hey guys, I'm going to close this PR and open a new one that works with the new order of execution, etc.

PR #647, opened 2020-07-10
Tests x-model binding for all input types

### LINKED ISSUE #407: Model binding tests for all input types

Marvinified:
I left out writing test for these input types `<input type=["file", "image" | "submit" | "button"]>` because I believe these input types don't require two-way data binding for any use case scenario and `(type=" file | image ")` were also not specifically implemented. 

Closes #407

### DIFF
--- a/test/model.spec.js
+++ b/test/model.spec.js
@@ -444,23 +444,254 @@ test('x-model bind datetime-local input', async () => {
 
 // <input type="email"> 
 test('x-model bind email input', async () => {
+})
+
+// <input type="month"> 
+test('x-model bind month input', async () => {
     document.body.innerHTML = `
-    <div x-data="{ key: 'alphine@example.com' }">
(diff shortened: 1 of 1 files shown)


calebporzio:
Thanks for picking this up @Marvinified! Great job!

PR #663, opened 2020-07-21
remove unecessary directive ordering clause

HugoDF:
Some changes that were part of https://github.com/alpinejs/alpine/pull/315/files aren't needed any more.

The relevant use-case's tests still pass without the guard since it's a specific instance of "bind should be evaluated before model"


calebporzio:
Oh this is fantastic, I hated that section. Thanks Hugo!

PR #664, opened 2020-07-21
fix(.camel modifier for namespaced events)

ryangjchandler:
This PR updates / changes the RegEx used for the `.camel` modifier so that it supports namespaced events.

Previously, `htmx:fetch-event` would get changed to `htmxFetchEvent`, now it will be changed to `htmx:fetchEvent`.

Also, this is the same RegEx pattern that Vue uses.


jreviews:
@ryangjchandler Tested this and it works great. Thanks!

calebporzio:
Thanks Ryan!

PR #667, opened 2020-07-22
fixes numeric attributes and adds test

### LINKED ISSUE #666: Incorrect binding for attributes with numbers in name
```html
<svg x-data="{ x: 10, y: 10 }">
  <line x1="0" y1="0" :x2="x" :y2="y" />
</svg>
```

Produces the following html:

```js
<svg>
  <line x1="0" y1="0" x="10" y="10" />
</svg>
```

The `x2` and `y2` attributes have the trailing number stripped. The same happens if you bind `x1` and `y1`.

A little bit of experimentation shows that Alpine strips all numeric content and everything that follows it from attribute names. Guessing there's an overly strict regex somewhere?

```html
<line :test123foo="4" />

<!-- becomes -->

<line test="4" />
```

danprince:
Just added numbers to the attribute parsing regex so that it doesn't silently remove them.

Motivating example is SVG line elements that take `x1`, `y1`, `x2`, `y2` attributes. Alpine has been trimming those down to `x` and `y`.

Closes #666

### DIFF
--- a/test/utils.spec.js
+++ b/test/utils.spec.js
@@ -1,7 +1,18 @@
-import { arrayUnique } from '../src/utils'
+import { arrayUnique, parseHtmlAttribute } from '../src/utils'
 
 test('utils/arrayUnique', () => {
     const arrayMock = [1, 1, 2, 3, 1, 'a', 'b', 'c', 'b']
     const expected = arrayUnique(arrayMock)
     expect(expected).toEqual([1, 2, 3, 'a', 'b', 'c'])
+})
+
+test('utils/parseHtmlAttribute', () => {
+    const attribute = { name: ':x1', value: 'x' };
+    const expected = parseHtmlAttribute(attribute);
+    expect(expected).toEqual({
+        type: 'bind',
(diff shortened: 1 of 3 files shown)


test/utils.spec.js
```diff
@@ -1,7 +1,18 @@
-import { arrayUnique } from '../src/utils'
+import { arrayUnique, parseHtmlAttribute } from '../src/utils'
 
 test('utils/arrayUnique', () => {
     const arrayMock = [1, 1, 2, 3, 1, 'a', 'b', 'c', 'b']
     const expected = arrayUnique(arrayMock)
     expect(expected).toEqual([1, 2, 3, 'a', 'b', 'c'])
 })
+
+test('utils/parseHtmlAttribute', () => {
```
HugoDF, line 9: Could we get a more end to end test in the `bind.js` suite as well as this?

danprince: All set!

calebporzio:
Thanks!

PR #673, opened 2020-07-23
Transition .delay modifier  

muzafferdede:
Currently we can not **delay** the transitions for the root or child elements where there are some reasonable use cases needs it. With this PR which uses almost the same approach as `.duration` modifier, we can delay the transitions for example like `x-show.delay.1000ms`. 

Also currently if parent element has `x-show.transition` while transitioning out, it will not wait for child to finish and just transition out and back to display block to wait child finish then hide with the child again which is seems like a bug. With this PR we are able to define a delay to wait for the child and do the transitioning for the parent as well.

Also it will open more opportunities for the `nested chain animations`. A sample codepen link down bellow if you want to check it out.

> Thanks to @SimoTod for the sample

https://codepen.io/muzafferdede/pen/jOWdZZE

### DIFF
--- a/README.md
+++ b/README.md
@@ -234,11 +234,11 @@ If you wish to run code AFTER Alpine has made its initial updates to the DOM (so
 | `x-show.transition.scale` | Only use the scale. |
 | `x-show.transition.scale.75` | Customize the CSS scale transform `transform: scale(.75)`. |
 | `x-show.transition.duration.200ms` | Sets the "in" transition to 200ms. The out will be set to half that (100ms). |
-| `x-show.transition.delay.200ms` | Sets the "in" transition to 200ms. The out will be set to half that (100ms). |
+| `x-show.transition.delay.200ms` | Sets 200ms delay on transition. |
 | `x-show.transition.origin.top.right` | Customize the CSS transform origin `transform-origin: top right`. |
 | `x-show.transition.in.duration.200ms.out.duration.50ms` | Different durations for "in" and "out". |
 
-> Note: All of these transition modifiers can be used in conjunction with each other. This is possible (although ridiculous lol): `x-show.transition.in.duration.100ms.origin.top.right.opacity.scale.85.out.duration.200ms.origin.bottom.left.opacity.scale.95`
+> Note: All of these transition modifiers can be used in conjunction with each other. This is possible (although ridiculous lol): `x-show.transition.in.duration.100ms.origin.top.right.opacity.scale.85.out.duration.200ms.delay.200ms.origin.bottom.left.opacity.scale.95`
 
 > Note: `x-show` will wait for any children to finish transitioning out. If you want to bypass this behavior, add the `.immediate` modifer:
 ```html
--- a/README.md
+++ b/README.md
@@ -234,6 +234,7 @@ If you wish to run code AFTER Alpine has made its initial updates to the DOM (so
 | `x-show.transition.scale` | Only use the scale. |
 | `x-show.transition.scale.75` | Customize the CSS scale transform `transform: scale(.75)`. |
 | `x-show.transition.duration.200ms` | Sets the "in" transition to 200ms. The out will be set to half that (100ms). |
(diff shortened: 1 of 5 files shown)


ryangjchandler:
Hey @muzafferdede - instead of adding a new modifier, would something like transition events help at all? You could listen for an event from the child element, then trigger the display change on the parent - this would seem like a more reliable approach since the delay might not always be perfect.

muzafferdede:
i tried that approach as well. Listening transition events are not consistent. Sometimes they get missed like 1 out of 10. Safari does more often. I think `.modifier` is best approach for now considering amount of implementing it requires.

SimoTod:
I believe that, because we remove classes (or styles in the x-show.transition version) using a timeout callback that has the same length of the transition itself, `transitionend` is not totally reliable. Sometimes it gets fired and sometimes it gets cancelled because classes and styles are removed 1ms before the transition ends.

Regardless the specific use case and whether or not we want to add that modifier, I think we should consider a possible transition delay when setting the callback timeout otherwise transition like https://codepen.io/SimoTod/pen/jOWgjMe would not work.

muzafferdede:
hmm, @SimoTod I might have another approach. Will try when i get time.

SimoTod:
@muzafferdede My comment was addressed to @ryangjchandler to discuss. What do you want to change?

I don't dislike this feature. I see the modifier as a nice to have for those who use the x-show.transition syntax but I think we should consider delays regardless (which you did at line https://github.com/alpinejs/alpine/pull/673/files#diff-2b4ca49d4bb0a774c4d4c1672d7aa781R464 to line 469)

muzafferdede:
We are on the same page @SimoTod. I think that line is the bare bone of the delay feature. I was thinking to not add a modifier if it is not worth it. Instead auto detect delay on element and sum to timeout. And considering your point on 1ms gap, maybe by default add 1ms delay?? Not sure... This is why i want to play around and see if i can come back with anything useful.

ryangjchandler:
@muzafferdede - when you come back to look at this, would you be able to just write a couple of bullet points on the changes that you have made and why you have made them or how they are intended to work. This just makes a bit easier for us to review the PRs and hopefully merge them. Thanks!

muzafferdede:
@ryangjchandler I don't get what you mean. This PR is not clear or simple enough to you? I have mentioned most of the things you ask like what it is and how it works etc in the first comment. Also @SimoTod and me worked together on this from idea to PR in the discussion section. I don't know how much more clear it could be.

muzafferdede:
@SimoTod i think i got another approach. So, if we list out both approach as;

### With `.delay` modifier
This is what we implemented with this PR. 
```html

<div x-data="{show: false}">
    <button x-on:click="show=!show">Toggle</button>
    <div x-show.transition.delay.1000="show">
        Parent
        <div x-show.transition.duration.1000="show">
            Child 1
        </div>
    </div>
</div>
```


### Detect `transitionDelay` css value on el
```html

<div x-data="{show: false}">
    <button x-on:click="show=!show">Toggle</button>
    <div x-show.transition="show" class="transition-delay-1000">
        Parent
        <div x-show.transition.duration.1000="show">
            Child 1
        </div>
    </div>
</div>
```
This requires few line of changes as:
```js
...
// utils.js line 457 ~ 478

        let transitionDuration = Number(getComputedStyle(el).transitionDuration.replace(/,.*/, '').replace('s', ''))
        let transitionDelay = getComputedStyle(el).transitionDelay ? Number(getComputedStyle(el).transitionDelay.replace(/,.*/, '').replace('s', '')) : 0
        
        if (transitionDuration + transitionDelay === 0) {
            transitionDuration = Number(getComputedStyle(el).animationDuration.replace('s', ''))
            transitionDelay = getComputedStyle(el).animationDelay ? Number(getComputedStyle(el).animationDelay.replace('s', '')) : 0
        }
        
        let duration = (transitionDuration + transitionDelay) * 1000
...
```
So we don't need to add `.delay` modifier. Transition delays can assign with css or any possible way. All we need to do is check if there is delay and sum to duration. 

What do you think?

muzafferdede:
Closing this PR after our discussion on discord.

PR #713, opened 2020-08-19
Missing transition delay + bug fix

muzafferdede:
This PR adds support for transition delay. Currently **transitions break** if the element has a **delay** defined [as in this sample](https://codepen.io/muzafferdede/pen/GRZjEry)
The code just adds the **transitionDelay** to the timeout value from the inline `.delay` modifier or computed styles, allowing transitions to complete after a delay.

You can check this [sample](https://codepen.io/muzafferdede/pen/bGpwRrm) which fixes bug with the new code **without any breaking changes**

- adds `.delay` modifier ex: `x-show.transition.delay.500`
- supports css class defined delays ex: `class="delay-500"`
- has inline and x-transition tests
- adds related info to docs
- uses almost same approach as `.duration`
- improved extracting time related computed properties for better performance

### DIFF
--- a/README.md
+++ b/README.md
@@ -239,10 +239,11 @@ If you wish to run code AFTER Alpine has made its initial updates to the DOM (so
 | `x-show.transition.scale` | Only use the scale. |
 | `x-show.transition.scale.75` | Customize the CSS scale transform `transform: scale(.75)`. |
 | `x-show.transition.duration.200ms` | Sets the "in" transition to 200ms. The out will be set to half that (100ms). |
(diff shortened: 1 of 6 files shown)


calebporzio:
Thank you for all the effort on this PR.

I think at some point we may want to support transition delay (either through the CSS property or the .delay modifier), but for now, I'd like to wait and see if this feature is more in demand.

Thanks again.

fabianwurk:
Would be great to have this **Delay** transition added to Alpine please 🙏🏻

PR #715, opened 2020-08-19
IE11 template polyfill issue leads to false warnings

### LINKED ISSUE #714: Invalid multiple element root error on IE11
I'm not sure if I am doing something wrong, but for some reason every `x-for` on IE causes the following warning to be shown in the console even though I only have one child node in my loop.  The error is not displayed in Chrome.

```
Alpine: <template> tag with [x-for] encountered with multiple element roots. Make sure <template> only has a single child node.
```

I can replicate the problem even with a very basic application like this:

```html
<!DOCTYPE html>
<html lang="en">
    <head>
        <meta charset="UTF-8" />
        <meta name="viewport" content="width=device-width, initial-scale=1.0" />
        <title>Document</title>
        <script
            type="module"
            src="https://cdn.jsdelivr.net/gh/alpinejs/alpine@v2.6.0/dist/alpine.min.js"
        ></script>
        <script
            nomodule
            src="https://cdn.jsdelivr.net/gh/alpinejs/alpine@v2.6.0/dist/alpine-ie11.min.js"
            defer
        ></script>
    </head>
    <body>
        <div x-data="test()">
            <template x-for="link in links">
                <div>
                    <span x-text="link.name"></span>
                </div>
            </template>
        </div>
        <script>
            function test() {
                return {
                    links: [
                        {
                            name: 'link1',
                        },
                        {
                            name: 'link2',
                        },
                    ],
                };
            }
        </script>
    </body>
</html>
```

HugoDF:
Closes #714 

- Switch to el.content.childNodes.length as per @ryangjchandler suggestion

### DIFF
--- a/dist/alpine-ie11.js
+++ b/dist/alpine-ie11.js
@@ -1276,7 +1276,28 @@
   })(EventListenerInterceptor);
 
   // For the IE11 build.
-  SVGElement.prototype.contains = SVGElement.prototype.contains || HTMLElement.prototype.contains;
+  SVGElement.prototype.contains = SVGElement.prototype.contains || HTMLElement.prototype.contains // .childElementCount polyfill
+  // from https://developer.mozilla.org/en-US/docs/Web/API/ParentNode/childElementCount#Polyfill_for_IE8_IE9_Safari
+  ;
+
+  (function (constructor) {
+    if (constructor && constructor.prototype && constructor.prototype.childElementCount == null) {
+      Object.defineProperty(constructor.prototype, 'childElementCount', {
+        get: function get() {
+          var i = 0,
+              count = 0,
+              node,
+              nodes = this.childNodes;
+
+          while (node = nodes[i++]) {
+            if (node.nodeType === 1) count++;
+          }
+
+          return count;
+        }
+      });
+    }
+  })(window.Node || window.Element);
 
   var check = function (it) {
     return it && it.Math == Math && it;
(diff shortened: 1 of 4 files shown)


ryangjchandler:
@HugoDF nice one 🤙🏼

ryangjchandler:
@HugoDF did you run this through on IE at all?

HugoDF:
> @HugoDF did you run this through on IE at all?

Nope, I don't have IE handy 👀

ryangjchandler:
> Nope, I don't have IE handy 👀

Okay, I'll use your branch as the CDN and test it out on IE to see if it's all good.

HugoDF:
@ryangjchandler done

ryangjchandler:
Nice one, I'll test this in a bit. FWIW, I put up a little HTML tester (https://previewer.ryangjchandler.co.uk) for easy testing in different browsers, specifically IE11 because CodePen and others dont' support IE.

ryangjchandler:
@HugoDF - I tested this and it still shows that same error.. I'm wondering if there is something else that we can use..

HugoDF:
> @HugoDF - I tested this and it still shows that same error.. I'm wondering if there is something else that we can use..

Ok we could just strip the warning for IE11?

HugoDF:
Converting to draft until I add the IE11 stripping comments to the warning function

SimoTod:
according to MDN, safari could have issues as well since it's not supported for documentFragment -> https://developer.mozilla.org/en-US/docs/Web/API/ParentNode/childElementCount

SimoTod:
childNodes is supported but it contains all nodes (including text nodes if you have spaces and stuff like that in between tags).
i think we can wither include the polyfill (https://developer.mozilla.org/en-US/docs/Web/API/ParentNode/childElementCount#Polyfill_for_IE8_IE9_Safari) ot just apply the same logic to the check (looping on all nodes and counting the ones that are real elements)

calebporzio:
I say we add the polyfill. The less we change the actual source for IE11 the better.

calebporzio:
Also, like Simone said, childNodes includes non-element nodes, so we should avoid that.

HugoDF:
Put it back to `childElementCount` and added the polyfill for IE11 🙂

calebporzio:
Great, thanks Hugo!

PR #719, opened 2020-08-21
Feature - $watch with arrays

SimoTod:
Adds a special case for arrays so watcher will trigger every time their elements changes (support pop, push, shift, unshift, sort, etc)


calebporzio:
Thanks @SimoTod!!

PR #726, opened 2020-08-23
docs(x-model `.number` modifier)

HugoDF:
I noticed that the `x-model.number` modifier is not mentioned anywhere, this adds a short example for it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -403,7 +403,7 @@ The `camel` modifier will attach an event listener for the camel case equivalent
 **`.number` modifier**
 **Example:** `<input x-model.number="age">`
 
-The `number` modifier will convert the input's value to a number.
+The `number` modifier will convert the input's value to a number. If the value cannot be parsed as a valid number, the original value is returned.
 
 **`.debounce` modifier**
 **Example:** `<input x-model.debounce="search">`
--- a/README.md
+++ b/README.md
@@ -405,7 +405,6 @@ The `camel` modifier will attach an event listener for the camel case equivalent
 
 The `number` modifier will convert the input's value to a number.
 
-
 **`.debounce` modifier**
 **Example:** `<input x-model.debounce="search">`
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -400,6 +400,12 @@ The `camel` modifier will attach an event listener for the camel case equivalent
 
 > Note: `x-model` is smart enough to detect changes on text inputs, checkboxes, radio buttons, textareas, selects, and multiple selects. It should behave [how Vue would](https://vuejs.org/v2/guide/forms.html) in those scenarios.
 
+**`.number` modifier**
+**Example:** `<input x-model.number="age">`
+
+The `number` modifier will convert the input's value to a number.
+
+
```
ryangjchandler, line 408: One too many blank lines I think ;)

ryangjchandler (changes requested):

README.md
HugoDF: Haha, removed the pesky extra line.

README.md
```diff
@@ -400,6 +400,11 @@ The `camel` modifier will attach an event listener for the camel case equivalent
 
 > Note: `x-model` is smart enough to detect changes on text inputs, checkboxes, radio buttons, textareas, selects, and multiple selects. It should behave [how Vue would](https://vuejs.org/v2/guide/forms.html) in those scenarios.
 
+**`.number` modifier**
+**Example:** `<input x-model.number="age">`
+
+The `number` modifier will convert the input's value to a number.
```
SimoTod, line 406: Shall we mention that the value is kept as it is if it's not a valid number?

HugoDF: I was going to add that it's using `parseFloat` under the hood but decided against it, we're failing gracefully but I dunno how valuable it is that users know that (I think they would just assume it fails gracefully)

SimoTod: we can write something which doesn't mention parseFloat (I copied Vue but we can use something like "If the value cannot be parsed as a valid number, the original value is returned.")
If we don't say anything, as a user I would assume that not valid number will be converted to 0 or NaN (which is what Alpine used to do before a PR changed that to match Vue)

HugoDF: Updated to that 👍

calebporzio:
Thanks Hugo!

Issue #724, opened 2020-08-23
x-spread example on documentation does not work


oguzhancelikarslan:
the example related to x-spread on documentation does not work.

oguzhancelikarslan:
@HugoDF I contacted with Simone Todaro yesterday. I prepared and shared the first Alpine JS Course in Turkish. and It almost completely contains every piece of documentation. To talk about it can you send me mail! oguzhancelikarslan@gmail.com.

The link: https://www.youtube.com/watch?v=CZloxR3pGwg&list=PLNqIPgpISSa8Yl04SvV2zTCnbp79A4mtH

ryangjchandler:
@oguzhancelikarslan Could you please discuss this elsewhere.

ryangjchandler:
@HugoDF I wonder if there is anything that we can automate for this, similar to how we replace the version in the distribution files, perhaps something on release. Or we go back to the old 2.x.x version in the README and make the note about pinning to a version in production more important.

oguzhancelikarslan:
> @oguzhancelikarslan Could you please discuss this elsewhere.

Sure. But I could not find the email address of @HugoDF.  Sorry if I disturb you.

HugoDF:
> @HugoDF I wonder if there is anything that we can automate for this, similar to how we replace the version in the distribution files, perhaps something on release. Or we go back to the old 2.x.x version in the README and make the note about pinning to a version in production more important. 

As part of the `build` script we could replace the pinned version with what's in the package.json

ryangjchandler:
> > @HugoDF I wonder if there is anything that we can automate for this, similar to how we replace the version in the distribution files, perhaps something on release. Or we go back to the old 2.x.x version in the README and make the note about pinning to a version in production more important.
> 
> As part of the `build` script we could replace the pinned version with what's in the package.json

Yeah, that could work - just a simple string or RegEx replace I guess.

HugoDF:
Closing this since `x-spread` is available as long as you're beyond `2.4.0` (have also sent in a PR that will update the README to latest package version so people grabbing the latest version from README should get all the documented features).

PR #735, opened 2020-08-28
Replace pinned version in readme on build

HugoDF:
Replace pinned version number in READMEs to be the latest per `package.json` on every `build` run.

This should reduce confusion (although people should really check what the latest is before pinning) for example #724 

Also from the CDN stats there's overwhelming use of `2.3.5` which is probably from README documenting it as "the version to pin to"


calebporzio:
Solid. Thanks Hugo!

PR #736, opened 2020-08-29
Make x-text work on SVG elements

### LINKED ISSUE #734: x-text doesn't work with SVG elements
I have an SVG image that is embedded in an HTML page inside an Alpine component. As part of the SVG there's a text element that I would like to bind to some Alpine data. Unfortunately, `x-text` does not seem to work for SVG elements. 

[Example](https://codepen.io/philippbosch/pen/qBZjBRJ?editors=1000)

I believe this is because [`x-text`tries to set `.innerText` on the element](https://github.com/alpinejs/alpine/blob/master/src/directives/text.js#L7), but SVG elements, or at least `<text>` don't have an `.innerText` attribute. 

Something like this could work:

```diff
-     el.innerText = output
+     if (el.hasAttribute('innerText') {
+       el.innerText = output;
+     } else if (el.hasAttribute('textContent') {
+       el.textContent = output;
+     }
```

It's not the most elegant solution, but unless somebody has a better approach, I can put together a PR for this.

philippbosch:
The `x-text` directive sets `innerText` on the element. The problem is that `innerText` is [only available on elements based on `HTMLElement`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/innerText) but not `SVGElement` which is the base for e.g. the `<text>` SVG element. 

~~So instead of blindly setting `el.innerText` we now first check if it's actually an HTML element and otherwise fall back to setting `el.textContent` which is [part of the `Node` interface](https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent) that all elements are based on.~~

Based on the discussion in #734, this PR swaps `innerText` out for `textContent` because they behave identically (at least when setting the value) and the latter is [part of the `Node` interface](https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent) that all elements – HTML and SVG – are based on.

Fixes #734

### DIFF
--- a/test/ref.spec.js
+++ b/test/ref.spec.js
@@ -10,22 +10,22 @@ test('can reference elements from event listeners', async () => {
         <div x-data="{}">
             <span x-ref="bob"></span>
 
-            <button x-on:click="$refs['bob'].innerText = 'lob'"></button>
+            <button x-on:click="$refs['bob'].textContent = 'lob'"></button>
         </div>
     `
 
     Alpine.start()
 
-    expect(document.querySelector('span').innerText).toEqual(undefined)
+    expect(document.querySelector('span').textContent).toEqual('')
 
     document.querySelector('button').click()
 
-    await wait(() => { expect(document.querySelector('span').innerText).toEqual('lob') })
+    await wait(() => { expect(document.querySelector('span').textContent).toEqual('lob') })
 })
 
 test('can reference elements from data object methods', async () => {
     document.body.innerHTML = `
-        <div x-data="{ foo() { this.$refs.bob.innerText = 'lob' } }">
+        <div x-data="{ foo() { this.$refs.bob.textContent = 'lob' } }">
             <span x-ref="bob"></span>
 
             <button x-on:click="foo()"></button>
@@ -34,9 +34,9 @@ test('can reference elements from data object methods', async () => {
 
     Alpine.start()
 
-    expect(document.querySelector('span').innerText).toEqual(undefined)
+    expect(document.querySelector('span').textContent).toEqual('')
 
     document.querySelector('button').click()
 
-    await wait(() => { expect(document.querySelector('span').innerText).toEqual('lob') })
+    await wait(() => { expect(document.querySelector('span').textContent).toEqual('lob') })
 })
--- a/src/directives/text.js
+++ b/src/directives/text.js
@@ -4,5 +4,9 @@ export function handleTextDirective(el, output, expression) {
         output = ''
     }
 
-    el.innerText = output
+    if (el instanceof HTMLElement) {
+        el.innerText = output
+    } else {
+        el.textContent = output
+    }
 }
--- a/src/directives/text.js
+++ b/src/directives/text.js
@@ -4,9 +4,5 @@ export function handleTextDirective(el, output, expression) {
         output = ''
     }
 
-    if (el instanceof HTMLElement) {
-        el.innerText = output
-    } else {
-        el.textContent = output
-    }
+    el.textContent = output
 }
--- a/test/constructor.spec.js
+++ b/test/constructor.spec.js
@@ -32,7 +32,7 @@ test('auto-detect new components at the top level', async () => {
 
     fireEvent.input(document.querySelector('input'), { target: { value: 'bar' }})
 
-    await wait(() => { expect(document.querySelector('span').innerText).toEqual('bar') })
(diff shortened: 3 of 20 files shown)


HugoDF:
There's some other failing tests that were checking innerText, a search and replace of innerText to textContent should fix them

philippbosch:
@HugoDF I'm on it. There's much more to it.

philippbosch:
So, I changed all the tests to work on `textContent` instead of `innerText`, and they all pass now. 

There are three kinds of changes to the tests:

1. **Substitute `innerText` with `textContent`**
   ```diff
   - await wait(() => { expect(document.querySelector('span').innerText).toEqual('bar') })
   + await wait(() => { expect(document.querySelector('span').textContent).toEqual('bar') })
   ```
   This one is straighforward and a no-brainer. 

2. **Additional change of type from int to string** 
   ```diff
   - expect(document.querySelector('span').innerText).toEqual(0)
   + expect(document.querySelector('span').textContent).toEqual('0')
   ```
   This kinda makes sense that because the thing is called `textContent` what you get from it is text, or a string. 

3. **Array to string conversion**
   ```diff
   - expect(document.querySelector('span').innerText).toEqual(['bar'])
   + expect(document.querySelector('span').textContent).toEqual('bar')
   ```
   This is also understandable, because `['bar'].toString() === 'bar'` but it has to potential to throw people off when reading the test code.

I am torn about whether this is what we want. Especially no. 3 is really confusing I find. 

One idea I had about that was that instead of testing for equality (`expect(el.textContent).toEqual('bar')`) we could have a custom Jest matcher specifically for testing `textContent` (`expect(el).toHaveATextContentOf(['bar'])`). This way we could hide the fact that the value is converted to a string and still allow people to pass in non-string values. 

What do you guys think? @HugoDF @ryangjchandler @SimoTod

ryangjchandler:
@philippbosch - I personally think that having a custom matcher is a bit overkill. The tests are pretty clear in what they do, since we are actually making assertions against a string so it's all regular JavaScript behaviour, I can't see any need for the custom matcher. See what the other guys think though.

SimoTod:
My 2 pence: I think the current tests are not real because jest doesn't support innerText. If you use it in a real browser `x-text="['foo','bar']`will print `foo,bar' not the js array.
Ideally we should use your tests because they are better.
textContent is also more performant but it only affect us marginally since we don't use it to read values.
That being said, the amount of changes to tests may require a longer review. Maybe ask Caleb directly on that PR explaining why you need to update them and if he doesn't want to change those tests, you can re-PR the initial version.

philippbosch:
@calebporzio Here's the tl;dr:

I want to make `x-text` work on SVG elements which is currently broken because `SVGElement`s don't have an `innerText` property. But there is also the `textContent` property which all elements have and which behaves identically and is actually standardized. So my approach was `s/innerText/textContent/g`. But after that tons of tests were broken. Turns out that jsdom does not implement `innerText` at all (https://github.com/jsdom/jsdom/issues/1245) so the existing tests that work on `innerText` are not realistic because if they ran in a real browser they would break – mainly because `innerText` (as well as `textContent`) always returns a string. For example, this one would break because the value would be `"0"` instead of the expected `0`: 

https://github.com/alpinejs/alpine/blob/12b2da3a40d086011d0c30a8ede7f8a2bf2f1fce/test/if.spec.js#L92

I propose to adjust all the tests and swap out `innerText` for `textContent` and make the necessary changes to the expected values like I have done in e72737f. It's a lot of changes but I'm conviced it's the right thing to do. What do you think?

calebporzio:
Totally agree. Good PR.

.textContent ftw!

test/ref.spec.js
```diff
@@ -10,22 +10,22 @@ test('can reference elements from event listeners', async () => {
         <div x-data="{}">
             <span x-ref="bob"></span>
 
-            <button x-on:click="$refs['bob'].innerText = 'lob'"></button>
+            <button x-on:click="$refs['bob'].textContent = 'lob'"></button>
```
HugoDF, line 13: this test could stick with `innerText` as well

HugoDF (commented):
found 2-3 instances where the test didn't need to be updated, I think it's fine to update them to be fair, just thought I would point them out.

philippbosch:
@HugoDF yeah, I saw these but thought it made sense to also change them for the sake of consistency

HugoDF:
> @HugoDF yeah, I saw these but thought it made sense to also change them for the sake of consistency

Makes sense

PR #743, opened 2020-09-02
Correct wording of warning message

cie:
This warning message is a bit misleading, the template[x-if] tag requires a single child element, not a single child node (because text nodes are also nodes)

- It can have some empty text nodes around the single child element
- It's not good if it has a single text node as a child.


HugoDF:
Do you have code samples of where the message would have been inaccurate?

HugoDF:
Found one on Discord

```html
<div x-data="{texts: ['hi', 'bye', 'sup']}">
    <template x-for="text in texts">
        Hola
        <div x-text="text"></div>
    </template>
</div>
```

Guess this PR makes sense

philippbosch (approved):

HugoDF:
Yep thanks @cie 

This change makes sense, I think @calebporzio will merge it when he gets a chance

calebporzio:
Thanks @cie!

PR #748, opened 2020-09-04
feature(x for over range)

ryangjchandler:
This PR adds support for the `x-for="i in 10"` syntax that can be found in Vue. Lots of people have come to us with questions lately about how to replicate this in userland, but honestly the syntax is quite verbose and ugly.

The equivalent of `i in 10` would be `i in [...Array(10).keys()]` which, in my opinion, isn't too nice to look at and a bit of a pain. 

The only thing to note is how it compares to Vue, **Vue will start the count at 1**. So, the first value of `i` in `i in 10` would be `1`. 

~~With this implementation in Alpine, **the first value of `i` will be `0`**, since the `Array` itself starts at `0`. Happy to change this though and make it more in line with Vue's implementation if it means that the PR will get merged, I can imagine a lot of people being happy with this syntax in the core.~~

The first item in the loop would be the **number 1**, up through 10. 

* [x] Implement `i in 10` syntax
* [x] Add note to docs about new syntax

### DIFF
--- a/README.md
+++ b/README.md
@@ -516,8 +516,6 @@ Alpine supports the `i in n` syntax, where `n` is an integer, allowing you to it
 </template>
 ```
 
-This will output 10 spans, starting with an `i` value of 0.
-
 ---
 
 ### `x-transition`
--- a/README.md
+++ b/README.md
@@ -506,6 +506,18 @@ You can nest `x-for` loops, but you MUST wrap each loop in an element. For examp
 </template>
 ```
 
+#### Iterating over a range
+
+Alpine supports the `i in n` syntax, where `n` is an integer, allowing you to iterate over a fixed range of elements.
+
+```html
+<template x-for="i in 10">
+    <span x-text="i"></span>
+</template>
+```
+
+This will output 10 spans, starting with an `i` value of 0.
+
 ---
 
 ### `x-transition`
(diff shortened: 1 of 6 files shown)


calebporzio:
Makes sense. Thanks Ryan!

carlmjohnson:
This issue is closed now, but for Alpine 3, could it be changed to be `i in $range(10)`? I think making a magic `$range()` function that works like Python's range ($range(2) = [1, 2], $range(2, 4)= [2, 3], $range(0, 4, 2) = [0, 2]) would be simpler to implement and simpler to understand, as well as allowing more flexibility (e.g. `$range({ from: 1, to: 2, inclusive: true})` = [1, 2]).

ryangjchandler:
@carlmjohnson I'm going to tag @KevinBatdorf here for this too, but the current range syntax is something that a lot of people had asked about in the past and is something that works out of the box in Vue.

A `$range` helper would only be useful, in my opinion, if you wanted those options as you said.

HugoDF:
> This issue is closed now, but for Alpine 3, could it be changed to be `i in $range(10)`? I think making a magic `$range()` function that works like Python's range ($range(2) = [1, 2], $range(2, 4)= [2, 3], $range(0, 4, 2) = [0, 2]) would be simpler to implement and simpler to understand, as well as allowing more flexibility (e.g. `$range({ from: 1, to: 2, inclusive: true})` = [1, 2]).

We could implement this outside of Alpine.js `$range` would be a couple lines of code

KevinBatdorf:
Sounds good to me. Added a discussion on the Helper's repo

https://github.com/alpine-collective/alpine-magic-helpers/discussions/30

PR #758, opened 2020-09-10
Support :indeterminate=true on checkboxes

philippbosch:
checkbox elements can have an `indeterminate` property which means its value is neither `true` or `false` (check [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input/checkbox#Indeterminate_state_checkboxes) for details). This is not an HTML attribute (~`<input type="checkbox" indeterminate>`~) but can only be set through JS. 

This does not work currently:
```html
<input type="checkbox" :indeterminate="true">
```

This PR adds a special case for the `bind` directive that covers this case.

### DIFF
--- a/src/directives/bind.js
+++ b/src/directives/bind.js
@@ -65,6 +65,8 @@ export function handleAttributeBindingDirective(component, el, attrName, express
             const newClasses = convertClassStringToArray(value)
             el.setAttribute('class', arrayUnique(originalClasses.concat(newClasses)).join(' '))
         }
+    } else if (el.tagName === 'INPUT' && el.type == 'checkbox' && attrName == 'indeterminate') {
+        el.indeterminate = value
     } else {
         attrName = modifiers.includes('camel') ? camelCase(attrName) : attrName
 
--- a/test/bind.spec.js
+++ b/test/bind.spec.js
@@ -436,6 +436,32 @@ test('radio values are set correctly', async () => {
     expect(document.querySelector('#list-test').checked).toBeFalsy()
 });
 
+test('indeterminate property is set correctly for checkboxes', async () => {
+    document.body.innerHTML = `
+        <div x-data="{foo: true, bar: false}">
+            <input type="checkbox" id="checkbox1" :indeterminate="foo" />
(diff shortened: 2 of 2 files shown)


SimoTod:
My personal opinion but I'm not sure that adding exceptions for specific cases it's a good idea. We would need to update it everytime we want to support a new DOM property. 
I'd rather see a generic prop modifier for x-bind like the one in Vue.

ryangjchandler:
> My personal opinion but I'm not sure that adding exceptions for specific cases it's a good idea. We would need to update it everytime we want to support a new DOM property.
> I'd rather see a generic prop modifier for x-bind like the one in Vue.

Yeah, I agree here. If there was a `prop` modifier for the `x-bind` directive, that would be better and more future-proof. 

@philippbosch Let me know if you need any help.

philippbosch:
Ah, that sounds very reasonable. I'll give it a try. Thanks for your feedback, @SimoTod @ryangjchandler

ryangjchandler:
@philippbosch Would you be able to open a new PR with the `.prop` modifier changes that reference this one?

philippbosch:
Sure, will do!

ryangjchandler:
> Sure, will do!

Wicked, cheers!

calebporzio:
Hey @philippbosch, let's hold off on that PR. I want to see more of a demand in the community before adding anything to the API for this case is it's fairly uncommon.

In the meantime, here's a CodePen I threw together to demonstrate how you would currently accomplish this in Alpine.

Thanks to the others for weighing in.

philippbosch:
Yeah, ok. I understand your concern. Thanks for the Codepen. It's basically what I have in my code right now.

newtonjob:
@calebporzio Just needed this right now and here's what I have in my code ATM;

```js
<div x-data="{ indeterminate: true }">
    <input type="checkbox" x-effect="$el.indeterminate = indeterminate">
    ...
</div>
```

Although it'd be really nice to simply do this;
```js
<div x-data="{ indeterminate: true }">
    <input type="checkbox" :indeterminate="indeterminate">
    ...
</div>
```

Works pretty well in Vue and so I had assumed it'd work the same way in Alpine.

PR #759, opened 2020-09-11
Change x-data docs JSON object to say object literal

KevinBatdorf:
A Discord user passed in a JSON object by reference and it caused unwanted side effects. This just adds more clarity to the docs.

Note, I updated the languages where this wasn't translated, but some were translated and need to be updated still. Translators can look here if they are unsure of the difference:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#Object_literal_notation_vs_JSON


KevinBatdorf:
Also, just for anyone ending up here via Google, you should be using a function if not inlining the object literal:

```html
Bad:
<div x-data="data"></div>
<script> window.data = { one: 'two' } </script>
```

```html
Bad, better to use ():
<div x-data="data"></div>
<script> window.data = function() { return { one: 'two' } }</script>
```

```html
Good:
<div x-data="data()"></div>
<script> window.data = function() { return { one: 'two' } }</script>
```

```html
Good:
<div x-data="{ one: 'two' }"></div>
```

calebporzio:
Yeah, this is an important distinction. Thanks @KevinBatdorf!

Also, damn dude - PRing to every single language? Baller move.

PR #761, opened 2020-09-12
feature(prop x for range)

### LINKED ISSUE #760: Iterating over a range has to be a Number literal
I'm not sure if this is intended, but you can't iterate over a range using a variable that holds an integer:
```
x-for="i in 10" works
x-for="i in j" where j is 10 doesn't work
```
Here's a pen:
https://codepen.io/mokhosh/pen/mdPpvZR

ryangjchandler:
Closes #760. Adds support for iterating over range from data property.

```html
<div x-data="{ count: 10 }">
    <template x-for="i in count" :key="i">
        <span x-text="i"></span>
    </template>
    <button type="button" @click="count++">Increment count</button>
</div>
```


calebporzio:
Perfect, thanks!

PR #767, opened 2020-09-14
feature(allow watching for changes on magic properties)

ryangjchandler:
This pull request replaces the `self.getUnobservedData()` method used inside of the watcher logic with the `self.unobservedData` property on the component.

The reason for this change is that when the membrane was being unwrapped, any magic properties on the `$data` object would be removed, meaning magic properties could not be watched for changes.

An example of this might be with Spruce, where somebody would want to watch for changes on the `$store.example.data` property, but because the `comparisonData` object did not have a `$store` key, it would simply fail. This opens up a lot of opportunity for cooler magic properties too (despite the fact that it's not a documented API) and also means that these libraries don't need to implement their own watcher logic.

This also isn't a huge change and does not produce any failing tests, so no regression is expected.

### DIFF
--- a/dist/alpine-ie11.js
+++ b/dist/alpine-ie11.js
@@ -23,11 +23,10 @@
   	return module = { exports: {} }, fn(module, module.exports), module.exports;
   }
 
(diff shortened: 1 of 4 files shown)


HugoDF:
Would it make sense to add a test ensuring you can watch magic properties?

calebporzio:
Sounds good

ryangjchandler:
YES!!!

PR #770, opened 2020-09-15
 fix(allow binding non string values to checkboxes)

ryangjchandler:
This PR fixes an issue where non-string values couldn't be bound to a checkbox's value. An example would be trying to bind an integer / number:

<input type="checkbox" x-data :value="100">

The value of the checkbox would actually be "on" due to the typeof value === 'string' check inside of the checkbox handler. This PR replaces that check with a value truth check and makes sure that it's not a boolean, then casts the value to a string using String(value).

This could be done in user-land using x-bind:value="String(100)" or template literals x-bind:value="${value}" but I'd say that it's a bit of a bug with the framework.

### DIFF
--- a/dist/alpine-ie11.js
+++ b/dist/alpine-ie11.js
@@ -6466,8 +6466,8 @@
         // If we are explicitly binding a string to the :value, set the string,
         // If the value is a boolean, leave it alone, it will be set to "on"
         // automatically.
-        if (typeof value === 'string' && attrType === 'bind') {
-          el.value = value;
+        if (value && typeof value !== 'boolean' && attrType === 'bind') {
+          el.value = String(value);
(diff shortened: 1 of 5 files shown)


ryangjchandler:
@calebporzio Hold off on this one for a second, just found an issue with these changes.

ryangjchandler:
@calebporzio Ready for review now - there was a problem where trying to bind an integer `0` wouldn't work because it was considered "falsy", sorted that.

@KevinBatdorf originally found this problem too btw, in case he wanted to check if this fix works for him. From the new test, it looks like it will sort it out.

calebporzio:
Ok, thanks Ryan.

Yeah, this seems like a bug and a good fix. Thanks!

PR #771, opened 2020-09-15
Add magic properties support in IE11

SimoTod:
`$wire` and other custom magic properties are currently broken in IE11.
IE11 uses a proxy polyfil that "seal" the object, meaning that no properties can be defined/redefined after the proxy is created so they need to be set earlier.


calebporzio:
Thanks @SimoTod!

PR #778, opened 2020-09-18
German translation for version 2.7.0

kkukelka:
- Added full German translation of the main Alpine documentation for version 2.7.0
- Added reference to German translation in main documentation


calebporzio:
Wow, thank you so much for this!

PR #788, opened 2020-09-22
x-bind:aria-* booleans enabled for true and false

davidleininger:
This PR fixes an issue where `x-bind` is used on an `aria-*` attribute as a boolean and the `false` state is removed.

Booleans for `aria-*` attributes require a `false` to exist so screen readers can understand the state of an object as per [W3C Recommendations](https://www.w3.org/TR/wai-aria-1.1/#state_prop_def).

#### `false` Inspector Screenshot
aria false example

### DIFF
--- a/src/directives/bind.js
+++ b/src/directives/bind.js
@@ -68,7 +68,7 @@ export function handleAttributeBindingDirective(component, el, attrName, express
     } else {
         attrName = modifiers.includes('camel') ? camelCase(attrName) : attrName
 
-        // If an attribute's bound value is null, undefined or false and does not include aria, remove the attribute
+        // If an attribute's bound value is null, undefined or false and does not start with aria, remove the attribute
         if ([null, undefined, false].includes(value) && !attrName.startsWith('aria')) {
             el.removeAttribute(attrName)
         } else {
--- a/src/directives/bind.js
+++ b/src/directives/bind.js
@@ -69,7 +69,7 @@ export function handleAttributeBindingDirective(component, el, attrName, express
         attrName = modifiers.includes('camel') ? camelCase(attrName) : attrName
 
         // If an attribute's bound value is null, undefined or false and does not include aria, remove the attribute
-        if ([null, undefined, false].includes(value) && !attrName.includes('aria')) {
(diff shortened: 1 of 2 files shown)


davidleininger:
No idea why that last check failed when the one before it didn't and in the last commit, I only added a line space in the readme. I don't have a way to rerun the test. Let me know if there is anything I need to do.

KevinBatdorf:
@davidleininger You can use `:aria-expanded="open.toString()"`

src/directives/bind.js
```diff
@@ -68,8 +68,8 @@ export function handleAttributeBindingDirective(component, el, attrName, express
     } else {
         attrName = modifiers.includes('camel') ? camelCase(attrName) : attrName
 
-        // If an attribute's bound value is null, undefined or false, remove the attribute
-        if ([null, undefined, false].includes(value)) {
+        // If an attribute's bound value is null, undefined or false and does not include aria, remove the attribute
+        if ([null, undefined, false].includes(value) && !attrName.includes('aria')) {
```
mikemcbride, line 72: You would likely want `!attrName.startsWith('aria')`, right? Otherwise you'd match on attributes containing `aria` inadvertently, like `data-variable`

davidleininger:
Fair point.

I guess I just thought it should be smart enough to do it for you. Either way, that a work around for me for now.

KevinBatdorf:
A note added to the documentation might be useful.

src/directives/bind.js
davidleininger: Done.

KevinBatdorf: Since the prefix is `aria-` maybe better to include the `-` as well?

ryangjchandler:
I feel as though there shouldn't be attribute specific exceptions built into the library. 

As @KevinBatdorf said, most people would just do this using `prop.toString()`, which isn't a bad way of doing it since it's more explicit which is in favour of the Alpine way. 

Perhaps maybe a note in the README about using `toString` makes more sense than adding an exception to the library for aria attributes. I can imagine there being a few edge cases that make this change a bit tricky.

davidleininger:
Closing this PR, opened another adding a note to the docs: #789

PR #789, opened 2020-09-22
Docs: Add note to chain .toString() to persist  false

davidleininger:
Adding a note to the docs for `x-bind` to make it clear to chain `.toString()` to the value if you need to persist a false value on a bound attribute.


calebporzio:
Thanks!

PR #796, opened 2020-09-30
docs: switch li to div

sagalbot:
The example for a modal uses a `ul`, with content, but with no children `li` wrapping the content. As far as I know this is invalid HTML/semantics. Swapped out the `ul` for a `div`.


KevinBatdorf:
Unless you add in aria labels, etc. a dropdown list should default to a `ul` ([source](https://www.w3.org/WAI/tutorials/menus/structure/)). A more correct change would be to just make that a comment inside.

```html
    <button @click="open = true">Open Dropdown</button>

    <ul
        x-show="open"
        @click.away="open = false"
    >
        <!-- Dropdown Body -->
    </ul>
</div>
```

PR #812, opened 2020-10-08
Remove garbled symbol on Japanese document

kimulaco:
Thank you for developing a great framework!
I found a garbled symbol in a [Japanese document](https://github.com/alpinejs/alpine/blob/master/README.ja.md#x-transition).

image

There was one unnecessary character that I couldn't see in the source code, so I deleted it.
If there are no mistakes, merge them!


calebporzio:
Thanks!

PR #815, opened 2020-10-11
Improve consistency of `checked` attribute value comparisons

### LINKED ISSUE #814: Cannot uncheck checkboxes with pre-set values
First I thought it has to do with Livewire, but after a lot of head scratching I checked Alpine without Livewire and the problem is still there.

Created a pen to demonstrate the problem. As you can see unchecked checkboxes can be checked and unchecked, but the ones that are set on load cannot:

https://codepen.io/mokhosh/pen/rNLVmvq

HugoDF:
Closes #814 

Some `el.checked` are being generated with loose equality (`==`), but in the x-model checkbox array case, unchecking was impossible due to a strict equality check (`!==`) in the "uncheck value" branch.

Extract the `el.checked` code to a util & use it where relevant (should help us keep it more consistent).


calebporzio:
Thanks @HugoDF!

PR #840, opened 2020-10-22
fix(x-for): issue with range iteration and numeric arrays

### LINKED ISSUE #839: x-for issue with numbers
Hi, 

I just noticed a bug with the latest alpine build (2.7.1) that doesn't affect versions prior.

If you x-for over and array of numerical values the first item always has a value of 1 when the array length is 1. Once the Array length is over 1 behaviour goes back to normal.

Here a code pen for it:

https://codepen.io/mariusaustr/pen/vYKydLy?editors=1001

Add a value that isn't 1 into input field and see that its rendered at 1. On the second addition the numbers return to normal.

ryangjchandler:
Closes #839. The problem was being caused by the `isNumeric` utility using `isNaN()` on arrays. 

When an array only contains 1 item, that is numeric, it will converted to a `Number` object that holds that value of the first item, therefore passing the `isNumeric` check. This PR just changes it to check that the `subject` being passed through is **not** an array using `Array.isArray`.

### DIFF
--- a/dist/alpine-ie11.js
+++ b/dist/alpine-ie11.js
@@ -6270,7 +6270,7 @@
     }.bind(this));
   }
   function isNumeric(subject) {
-    return !isNaN(subject);
+    return !Array.isArray(subject) && !isNaN(subject);
   } // Thanks @vuejs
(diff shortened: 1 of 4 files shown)


calebporzio:
Thanks @ryangjchandler!

Issue #834, opened 2020-10-20
x-for adjacent to another x-for throws not defined error


kevinvayro:
I am trying to make a datepicker that uses Alpine work but I am having some issues when I place two x-for's together.

I have two arrays, one holds the dates for a month and the second holds the number of blank days required to make sure the first day aligns with the correct day of the week.

Everything works and I can loop over both arrays and display the contents of the array however when I put the two x-for templates into the same div, after changing the month a couple of times I get a `Uncaught ReferenceError: date is not defined`. The template is `<template x-for="date in no_of_days" :key="date">`

If I reverse the order of the two templates and make the blank days show at the end of the month, which isn't what I want, then everything works fine and there is no error.

Here is a demo of the issue I'm having. [https://codepen.io/kevinvayro/pen/KKMgyXg](https://codepen.io/kevinvayro/pen/KKMgyXg). There are 4 examples that show the scenarios where it is working and not. In the scenario that doesn't work, the number of days will get stuck at 30 after pressing 'Next month' twice.

I have a workaround which is to just add the blank days to the days array as an empty string but I'd like to know if this is an actual issue or if I'm doing something wrong.

calebporzio:
Interesting, so let's refine the issue.

Can you confirm that any adjacent template/x-for tags cause a breakage?

```html
<template x-for="foo in bar">
    <div>...</div>
</template>
<template x-for="bar in baz">
    <div>...</div>
</template>
```

??

kevinvayro:
I've done some more experimenting and I've replaced the original arrays with two new arrays that are just filled to a random length using foo and bar as my template variable names and the issue still is the same.

I do now have some more information though which might be of use.

Before the error appears, the HTML generated looks as follows. The generated divs sit below the template.
  
```
<div class="flex">
    <template x-for="bar in random_array2">
        <div x-text="bar" style="width: 14.28%" class="px-1 mb-1 text-red-400"></div>
    </template>
    <div x-text="bar" style="width: 14.28%" class="px-1 mb-1 text-red-400">0</div>
    
    <template x-for="foo in random_array" :key="foo">
        <div style="width: 14.28%" class="px-1 mb-1">
            <div x-text="foo"></div>
        </div>
    </template>
    <div style="width: 14.28%" class="px-1 mb-1">
        <div x-text="foo">0</div>
    </div>
    <div style="width: 14.28%" class="px-1 mb-1">
        <div x-text="foo">1</div>
    </div>
</div>
```

[...]

PR #851, opened 2020-10-26
Add check for template tag when there are sibling x-for

KevinBatdorf:
Likely fixes https://github.com/alpinejs/alpine/issues/834

I'm not totally familiar with this part of the code to be fair, but all the tests are passing, and the test I added fails otherwise.


KevinBatdorf:
The title isn't totally accurate. It will now just ignore anything that isn't keyed, not just `template` tags

calebporzio:
Yep, this makes good sense. Thanks so much for pitching in on this @KevinBatdorf!

Issue #835, opened 2020-10-21
On clone, magic helpers don't have access to the freshly cloned component


KevinBatdorf:
Is there intent behind passing in the `canonicalComponentElementReference` to the magic helpers instead of passing in the freshly cloned component? 

```js
let canonicalComponentElementReference = componentForClone ? componentForClone.$el : this.$el
Object.entries(Alpine.magicProperties).forEach(([name, callback]) => {
  Object.defineProperty(dataExtras, `$${name}`, { get: function () { return callback(canonicalComponentElementReference) } });
})
```

See here: https://github.com/alpinejs/alpine/blob/master/src/component.js#L25

This limits magic helpers that add properties or keep track of components as only the component that's about to be destroyed is accessible.

@SimoTod ran a test on `$wire` with the following and says it works all the same. 
```js
Object.entries(Alpine.magicProperties).forEach(([name, callback]) => {
  Object.defineProperty(dataExtras, `$${name}`, { get: function () { return callback(this.$el) } });
})
```

If the `canonicalComponentElementReference` _is_ required, we could pass `this.$el` in as a second param instead.

calebporzio:
Can you paste in a bit of code that demonstrates why this is a hangup for you?

I believe I NEEDED to pass the canonical element reference, so that `$wire` could access the ACTUAL runtime Livewire component when an Alpine component is cloned for morphdom.

Has someone tested that passing just `this.$el` still works with Livewire's `$wire` across morphdom diffs?

calebporzio:
Copying from closed issue:

> Can you paste in a bit of code that demonstrates why this is a hangup for you?
> 
> I believe I NEEDED to pass the canonical element reference, so that `$wire` could access the ACTUAL runtime Livewire component when an Alpine component is cloned for morphdom.
> 
> Has someone tested that passing just `this.$el` still works with Livewire's `$wire` across morphdom diffs?

calebporzio:
Let's move this discussion to #859

PR #859, opened 2020-10-31
Helpers to receive access to cloned component

KevinBatdorf:
Opening this PR to make it easier to understand my intent of #835 

The `canonicalComponentElementReference` may be needed to compute the initial state but it doesn't seem needed after that. More importantly, magic helpers in some cases need access to the most current cloned component, not the canonical component (which is about to be deleted).

Relevant part of the code:
https://github.com/alpinejs/alpine/blob/master/src/component.js#L25

### DIFF
--- a/src/component.js
+++ b/src/component.js
@@ -22,10 +22,8 @@ export default class Component {
             $el: this.$el,
         }
 
-        let canonicalComponentElementReference = componentForClone ? componentForClone.$el : this.$el
-
         Object.entries(Alpine.magicProperties).forEach(([name, callback]) => {
-            Object.defineProperty(dataExtras, `$${name}`, { get: function () { return callback(canonicalComponentElementReference) } });
+            Object.defineProperty(dataExtras, `$${name}`, { get: function () { return callback(this.$el) } });
         })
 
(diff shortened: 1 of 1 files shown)


calebporzio:
Posted from closed issue. We can continue the discussion here now.

> Can you paste in a bit of code that demonstrates why this is a hangup for you?
> 
> I believe I NEEDED to pass the canonical element reference, so that `$wire` could access the ACTUAL runtime Livewire component when an Alpine component is cloned for morphdom.
> 
> Has someone tested that passing just `this.$el` still works with Livewire's `$wire` across morphdom diffs?

KevinBatdorf:
I wrote a [helper](https://github.com/alpine-collective/alpine-magic-helpers/blob/bf88f7055266d59f70dbf3e71fdf907bee1c3e1a/src/undo.js) to track component state over time and undo changes ([demo](https://codepen.io/KevinBatdorf/pen/c674633a3dcf8c0729345ab8ab86791f?editors=1010)).

To avoid any chance of breaking something, can we pass in the clone as a second param? I've updated this PR.

calebporzio:
Sounds good!

And WOW, that is an insane helper @KevinBatdorf - nice work!

PR #862, opened 2020-11-03
Feat: better (eval) error messages/context

### LINKED ISSUE #783: Error context is lost
This is Alpine v2.7.0.

If I do something like this:

```
 <template x-for="item in thisdoesnotexist" :key="item">
<div></div>
</template>
```

I get this not very useful error in the console:

```html
[Error] TypeError: undefined is not an object (evaluating 'items.forEach')
	(anonymous function) (alpine.js:1762)
```

In non-trivial setups, guessing where the faulty template code is, is extremely hard.

I remember in a not-so-distant Alpine version that I saw some more context (as in: the actual script string that failed to evaluate).

HugoDF:
Closes #783 (or maybe is just a starting point to better error messages)

tl;dr: 
- surface what expression is being evaluated and why it's throwing
- surface the component from which the error is being thrown

Contents of the PR:
- handling of eval errors.
- handling of initialization errors.
- handling of listener errors.
- error unit tests.
- better error-handling for x-for (throw if collection is not an array)
- error examples (x-data and x-on)

### DIFF
--- a/examples/index.html
+++ b/examples/index.html
@@ -48,9 +48,10 @@
             </thead>
             <tbody>
                 <tr>
-                    <td>Broken Component</td>
+                    <td>Broken Components</td>
                     <td>
                         <div x-data="some.bad.expression()">I'm a broken component</div>
+                        <button x-data x-on:click="something()">I break on click</button>
                     </td>
                 </tr>
 
--- a/src/directives/for.js
+++ b/src/directives/for.js
@@ -9,6 +9,10 @@ export function handleForDirective(component, templateEl, expression, initialUpd
 
     let items = evaluateItemsAndReturnEmptyIfXIfIsPresentAndFalseOnElement(component, templateEl, iteratorNames, extraVars)
 
+    if (! Array.isArray(items)) {
+        throw new TypeError(`collection "${iteratorNames.items}" in x-for="${expression}" is not an array, got "${JSON.stringify(items)}"`)
+    }
+
     // As we walk the array, we'll also walk the DOM (updating/creating as we go).
     let currentEl = templateEl
     items.forEach((item, index) => {
--- a/src/directives/on.js
+++ b/src/directives/on.js
@@ -1,4 +1,4 @@
-import { kebabCase, camelCase, debounce, isNumeric } from '../utils'
+import { kebabCase, camelCase, debounce, isNumeric, rethrowErrorWithContext } from '../utils'
 
 export function registerListener(component, el, event, modifiers, expression, extraVars = {}) {
     const options = {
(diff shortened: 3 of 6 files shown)


calebporzio:
Thanks @HugoDF - better error handling has been on the TODO for me. I'll take a look at this soon

HugoDF:
I think inevitably we're going to hitting some limits to what we can do.

eg. It will take some additional research/work to find the line of the HTML file where the error is happening

But this PR forwards eval errors to the user instead of swallowing them.

HugoDF:
@calebporzio any chance we could look at merging this? The current errors aren't any more helpful than what this introduces

calebporzio:
Ok Hugo.

Sorry, I haven't merged this because I'm uncomfortable with the solution. It's a pretty big change to the interface of Alpine from a dev's perspective, so I'm extra cautious.

Couple things: (note: I haven't FULLY wrapped my head around your implementation)
* Is there an implementation where you wrap the eval in a try catch in ONE place (or two because of saferEvalNoReturn) - rather than peppering error handling in the x-for and other places that use saferEval at some level? I'd be more comfortable with this if it was isolated
* Rather than adding .outerHTML to the error message, I think we should still throw the error, just add a console.warn above it with a reference to the offending element. This way, DevTools will handle showing the element AND allowing the developer to click on the element to go to it's place in the DOM. Here's a sample code snippet of how this would work:

```js
function tryCatch(el, callback, ...args) {
    try {
        return callback(...args)
    } catch (e) {
        console.warn('Alpine Expression Error: '+e.message, el)

        throw e
    }
}
```

HugoDF:
> Is there an implementation where you wrap the eval in a try catch in ONE place (or two because of saferEvalNoReturn) - rather than peppering error handling in the x-for and other places that use saferEval at some level? I'd be more comfortable with this if it was isolated

the situation here is that:
- `saferEval*` doesn't have access to the relevant element (so no way to know which component/directive caused the eval error)
- where we have access to the element we don't have access to the expression causing the error

> Rather than adding .outerHTML to the error message, I think we should still throw the error, just add a console.warn above it with a reference to the offending element

Yeah I'm fine with removing this particular stuff, I guess we can use `console.error` (since it is an error)

With the above in mind, I think maybe a good spot to do this logic would be in component.js https://github.com/alpinejs/alpine/blob/b2ad786f374074b7563ec00e83772d4af696474d/src/component.js#L353-L365? And maybe around saferEval use in component initialisation ? (https://github.com/alpinejs/alpine/blob/master/src/component.js#L31)

That should get rid of most of the try/catch-ing in saferEval itself.

Note that this change https://github.com/alpinejs/alpine/pull/862/files#diff-27401b46ee2dd7c5623b592a3b513742553b0cd6261fd2430222771e94ba7c72R12-R15 technically doesn't need to be in the PR, it's a new error/check that's better than `cannot read property forEach of ...`

HugoDF:
@calebporzio I've implemented with the `tryCatch` + console.error in this PR https://github.com/alpinejs/alpine/pull/887

It's not just 2 spots in saferEval but it's 3 instances, all in component.js (x-data, evaluateCommand, evaluateReturn)

HugoDF:
I'm closing this in favour of the other PR, we can always come back and use relevant code when/if that's ever needed.

PR #873, opened 2020-11-12
docs: add link to MDN "Event" on `x-on`

HugoDF:
A lot of people are asking which events they can use (not aware that the types of events are not Alpine.js-specific).

This should clear that up by linking to MDN "Event Reference"


calebporzio:
Thanks Hugo!

Issue #756, opened 2020-09-09
Alpine.clone results in click handler being called twice


dvic:
Reported in https://github.com/phoenixframework/phoenix_live_view/issues/1140 with an example repo.

Short summary: I have a list view that gets three items added after 1 second. Each list item has a `@click` handler. The weird thing is: it seems that `Alpine.clone` results in the click handler of the list item being called twice. If I remove that line in `app.js`, the click handlers still work and they are just called once. Also putting `x-data` on the list item affects this bug, with it, it never gets called twice.

I tried to isolate the issue with a JavaScript only example, but I was not able to (because in that example I'm not calling / don't need to call Alpine.clone). I'm not sure what a "proper" javascript-only example would be that tests whether `Alpine.clone` is doing the right thing. Any suggestions?

**update:** the problem is not specific to Phoenix, the problem also exists with Livewire: https://github.com/dvic/hello-livewire (click on the button on the home page, then on the list item, `console.log` will be called twice)

dvic:
**update** I managed to reproduce the issue, not sure if it makes sense though (with the `Alpine.clone` call):

https://jsbin.com/zucomuguza/1/edit?html,js,console,output

![image](https://user-images.githubusercontent.com/1214337/92623130-27005100-f2c6-11ea-88db-eb3b0bd08471.png)

HugoDF:
From debugging the source it seems that the `li` listener is somehow getting initialised twice, once by the mutation observer and once by the existing `ul` component re-initialising itself on post-DOM-insertion

dvic:
@HugoDF Thanks for looking into it. Is the mutation observer attached to each scope? Because if I remove the `x-data` on the outer div, the problem disappears.

KevinBatdorf:
@HugoDF cc: @SimoTod @ryangjchandler 

I'm not 100% sure this line in Alpine is correct. It might need to be reversed. It's a pretty important line to change I think so might want to make sure it's tested.
https://github.com/alpinejs/alpine/blob/master/src/component.js#L378

For example, in the helpers repo, I had to reverse that to get things working properly under the conditions I gave it.
https://github.com/alpine-collective/alpine-magic-helpers/blob/master/src/component.js#L26

It also addresses the above. Here's a codepen:
https://codepen.io/KevinBatdorf/pen/762e34ccc866ac26beb355bca363745e

KevinBatdorf:
[...]

PR #882, opened 2020-11-18
WIP: Clean up event listeners on clone

KevinBatdorf:
Currently, under some circumstances listeners will linger after a component is cloned. 

Fixes https://github.com/alpinejs/alpine/issues/756

### DIFF
--- a/test/on.spec.js
+++ b/test/on.spec.js
@@ -560,3 +560,31 @@ test('.camel modifier correctly binds event listener with namespace', async () =
         expect(document.querySelector('p').textContent).toEqual('bob');
     });
 })
+
+test('event listeners are removed from source component when cloned', async () => {
+    window.cloneNestedComponent = function(el) {
+        const x = el.__x;
+        delete el.__x
+        Alpine.clone(x, el)
+    }
+
+    document.body.innerHTML = `
+        <div x-data="{ count: 0 }" x-on:increment="count = count+1">
+            <span x-bind:foo="count"></span>
+
+            <button x-data id="increment" x-on:click="$dispatch('increment')"></button>
+        </div>
+    `
+
+    Alpine.start()
+
+    window.cloneNestedComponent(document.querySelector('button#increment'))
+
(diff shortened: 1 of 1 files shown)


calebporzio:
Interesting.

In your test case, you are registering a listener on an element via Alpine, destroying the Alpine component reference and re-cloning the component on to that element.

I WOULD expect that the listener would be fired twice.

.clone is meant to be used on different elements. And in those cases you want the listener to be run on both the original component element AND the cloned one.

KevinBatdorf:
I was under the impression that when you use `clone()`, the original item was to be removed completely after the cloning. Maybe I misunderstood its use (I have little Livewire experience still =/)

Someone added a [test case in Livewire](https://github.com/dvic/hello-livewire/blob/master/resources/views/livewire/counter.blade.php) though. Maybe someone else here can take a look and can chime in whether they are just doing it wrong?

Side note, I spent today trying to fix actually this and ran into issues because `removeEventListener` expects the EXACT SAME function to be passed in as you used with `addEventListener`, and that's not a trivial refactor.

calebporzio:
Ah thanks for clarifying. No, the idea with clone is to initialize a new component with a different component's state.

It's primary use is for Livewire's DOM diffing tool "morphdom" to initialize the new HTML with the existing Alpine state BEFORE performing the DOM-diff.

I see that referenced component. Whoever that was should post a GH issue on Livewire's repo or ideally, PR a failing test for that case, and we can tackle it on that repo because it's Livewire specific.

Thanks for digging into this Kevin. Much appreciated!

Closing this.

dvic:
> Whoever that was should post a GH issue on Livewire's repo or ideally, PR a failing test for that case, and we can tackle it on that repo because it's Livewire specific.

@calebporzio Hi! I made the example that @KevinBatdorf mentioned. I'm not a PHP developer / Livewire user so I'm not sure I can create a PR with a failing test case but I've submitted an issue for it: https://github.com/livewire/livewire/issues/2126. One more thing: this exact issue occurs also with the Elixir Phoenix Liveview library (that's how I encountered it).

PR #887, opened 2020-11-23
Feat: better (eval) error messages/context

### LINKED ISSUE #783: Error context is lost

HugoDF:
Closes #783 

from these discussion points on the other PR:
- https://github.com/alpinejs/alpine/pull/862#issuecomment-732275009 
- https://github.com/alpinejs/alpine/pull/862#issuecomment-732293276

Implement the suggestion from the latter comment.

Errors are logged as follows:

Screenshot 2020-11-23 at 18 14 42

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -67,7 +67,7 @@ export function debounce(func, wait) {
 
 const handleError = (el, expression, error) => {
     console.error(`Alpine: error in expression "${expression}" in component:`, el, `due to "${error}"`);
-    if (!isTesting()) {
+    if (! isTesting()) {
         throw error;
     }
 }
--- a/src/utils.js
+++ b/src/utils.js
@@ -66,7 +66,7 @@ export function debounce(func, wait) {
 }
 
 const handleError = (el, expression, error) => {
-    console.error(`Alpine: error in expression "${expression}" in component: `, el, `due to "${error}"`);
+    console.error(`Alpine: error in expression "${expression}" in component:`, el, `due to "${error}"`);
     if (!isTesting()) {
         throw error;
     }
--- a/src/utils.js
+++ b/src/utils.js
@@ -65,6 +65,24 @@ export function debounce(func, wait) {
     }
 }
 
+const handleError = (el, expression, error) => {
+    console.error(`Alpine: error in expression "${expression}" in component: `, el, `due to "${error}"`);
+    if (!isTesting()) {
+        throw error;
+    }
+}
+
+export function tryCatch(cb, { el, expression }) {
+    try {
+        const value = cb();
+        return value instanceof Promise
+            ? value.catch((e) => handleError(el, expression, e))
+            : value;
+    } catch (e) {
+        handleError(el, expression, e)
+    }
+}
+
 export function saferEval(expression, dataContext, additionalHelperVariables = {}) {
     if (typeof expression === 'function') {
         return expression.call(dataContext)
--- a/src/utils.js
+++ b/src/utils.js
@@ -142,8 +142,7 @@ export function getXAttrs(el, component, type) {
     let spreadDirective = directives.filter(directive => directive.type === 'spread')[0]
 
     if (spreadDirective) {
-        const { expression } = spreadDirective
-        let spreadObject = saferEval(el, expression, component.$data)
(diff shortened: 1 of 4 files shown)


src/utils.js
```diff
@@ -65,6 +65,24 @@ export function debounce(func, wait) {
     }
 }
 
+const handleError = (el, expression, error) => {
+    console.error(`Alpine: error in expression "${expression}" in component:`, el, `due to "${error}"`);
+    if (!isTesting()) {
```
HugoDF, line 70: Did this so that the app doesn't throw during tests eg. when testing event handlers

src/utils.js
```diff
@@ -65,6 +65,24 @@ export function debounce(func, wait) {
     }
 }
 
+const handleError = (el, expression, error) => {
+    console.error(`Alpine: error in expression "${expression}" in component:`, el, `due to "${error}"`);
+    if (! isTesting()) {
+        throw error;
+    }
```
HugoDF, line 72: Avoid throwing during tests eg. when testing event handlers if we throw it's basically untestable

src/utils.js
```diff
@@ -65,6 +65,24 @@ export function debounce(func, wait) {
...
+}
+
+export function tryCatch(cb, { el, expression }) {
+    try {
+        const value = cb();
+        return value instanceof Promise
+            ? value.catch((e) => handleError(el, expression, e))
+            : value;
+    } catch (e) {
+        handleError(el, expression, e)
```
HugoDF, line 82: handles sync errors

src/utils.js
```diff
@@ -65,6 +65,24 @@ export function debounce(func, wait) {
...
+    if (! isTesting()) {
+        throw error;
+    }
+}
+
+export function tryCatch(cb, { el, expression }) {
+    try {
+        const value = cb();
+        return value instanceof Promise
+            ? value.catch((e) => handleError(el, expression, e))
```
HugoDF, line 79: handle async errors eg. `saferEvalNoReturn` when it's an AsyncFunction

calebporzio:
Thanks @HugoDF, this is looking good. I like how you are also reporting the expression that caused the breakage.

One thought. In the case of x-spread (I think), sometimes the `expression` passed to `saferEval` is a function, and not a string. In that case, the concatenation will likely fail. Can you check on that?

Sorry for putting you through the ringer on this.

Thanks again!

HugoDF:
> Thanks @HugoDF, this is looking good. I like how you are also reporting the expression that caused the breakage.
> 
> One thought. In the case of x-spread (I think), sometimes the `expression` passed to `saferEval` is a function, and not a string. In that case, the concatenation will likely fail. Can you check on that?
> 
> Sorry for putting you through the ringer on this.
> 
> Thanks again!

I think it should work, let me add a unit test and example though 😄

HugoDF:
🤔 we're going to be up to 4 `tryCatch`-es, at that point I reckon it might be more future-proof to pass `el` into `saferEval`/`saferEvalNoReturn` and do the error reporting there.

That's a pretty trivial refactor in the end, a case of removing `tryCatch` callsites and instead passing `el` into the eval functions.

HugoDF:
@calebporzio I've moved the `tryCatch` into saferEval/saferEvalNoReturn see this PR https://github.com/HugoDF/alpine/pull/2/files (it's on my repo but will get reflected into this PR if you think it makes sense)

we save 0.1kb but more importantly we only have the `tryCatch` calls in 2 spots (downside is that we have to pass `el` to saferEval/saferEvalNoReturn)

calebporzio:
No, I like the direction. Good call

HugoDF:
> No, I like the direction. Good call

cool, should be ready for review then 👍

calebporzio:
Thanks for this Hugo! Going to be extremely helpful for users

PR #890, opened 2020-11-25
Fuse filter calls when combined with forEach, map...

diasbruno:


calebporzio:
Closing because there is no description or tests.

diasbruno:
@calebporzio Sorry. Updated the description.

PR #891, opened 2020-11-25
Fuse filter when combined with forEach, map...

diasbruno:
`.filter` calls when combined with `.forEach` or `.map` results in a 2 traversal and requires to allocate an intermediate array.

`.filter().map()` can be fused using `.reduce()`.
`.filter().forEach()` can be fused using only the `.forEach` and return if the current element don't pass the filter predicate.

In this refactor, all tests are done indirectly.

### DIFF
--- a/src/component.js
+++ b/src/component.js
@@ -149,8 +149,9 @@ export default class Component {
                 // Let's walk through the watchers with "dot-notation" (foo.bar) and see
                 // if this mutation fits any of them.
                 Object.keys(self.watchers)
-                    .filter(i => i.includes('.'))
                     .forEach(fullDotNotationKey => {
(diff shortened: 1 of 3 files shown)


ryangjchandler:
@diasbruno The chance of this sort of change being merged is are minuscule, unless it makes a _really_ big difference to performance, which in this case, it won't I don't think.

I'd also say that the original code is more readable, especially that `.reduce` call. Of course, it's up to @calebporzio, but based on previous PRs doing a similar thing, I can't see this being merged in.

diasbruno:
@ryangjchandler No problem. 👍

diasbruno:
Ok, closing it. I'll try to benchmark to see if there is any significant improvement.

PR #903, opened 2020-11-30
Update japanese Readme translation

g-kenkun:
Japanese Readme was missing the $watch and the security and V3 roadmap sections, so I added them.

Please let me know if you have any suggestions.


calebporzio:
Thanks!

PR #919, opened 2020-12-07
Update README.md Add Example2 for x-ref 

EOM:
Add Example 2 for x-ref and focus() in <input>


calebporzio:
Thanks for being willing to contribute, however, I'm going to hold off on this.

PR #925, opened 2020-12-10
Add module JS export

### LINKED ISSUE #166: Create a modules version of Alpine
See https://www.pika.dev/search?q=alpine and https://github.com/rollup/rollup/wiki/pkg.module for an explanation. 

Essentially, now that JavaScript modules are defined and there are adaptors to use them anywhere, there’s no good reason to distribute a CommonJS library anymore besides a legacy build tool chain, but Alpine is a new project and shouldn’t have that problem.

carlmjohnson:
This saves 1.5KB (300 bytes gzip, lol) after minification by removing unnecessary Babel definitions for defineProperty, ownKeys, and _objectSpread2.

Fixes #166

### DIFF
--- a/dist/alpine.mjs
+++ b/dist/alpine.mjs
@@ -1,0 +1,2002 @@
+// Thanks @stimulus:
+// https://github.com/stimulusjs/stimulus/blob/master/packages/%40stimulus/core/src/application.ts
+function domReady() {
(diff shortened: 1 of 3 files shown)


calebporzio:
Thank you for this!

This will be available in the next version of Alpine though. No plans for this version.

Thanks!

PR #958, opened 2020-12-26
Cast x-spread expression to string

### LINKED ISSUE #957: Using x-text inside x-spread doesn't work. Throws error!
JS Code:
```
loginForm = function() {
    return {
        errors: {},
        errorsMessage:{
            ['x-text']() {
                return this.errors.message;
            },
            ['x-show']() {
                return this.errors.message != '';
            }
        }
    }
}
```

HTML:
```
<div x-spread="errorsMessage"></div>
```

throws this error:
image

If I remove the `[x-text]`, it works fine.

KevinBatdorf:
This breaks because expressions are functions in x-spread, and when the output is defined, it attempts to run a String operation on it.

```js
 if (output === undefined && expression.match(/\./)) {
     output = ''
 }
```

...which causes an error. This fix will just cast the function there to a string so it won't error out.

Fixes #957

### DIFF
--- a/README.md
+++ b/README.md
@@ -621,6 +621,7 @@ The object keys are the directives (Can be any directive including modifiers), a
 > Note: There are a couple of caveats to x-spread:
 > - When the directive being "spread" is `x-for`, you should return a normal expression string from the callback. For example: `['x-for']() { return 'item in items' }`.
 > - `x-data` and `x-init` can't be used inside a "spread" object.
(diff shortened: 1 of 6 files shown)


KevinBatdorf:
Also snuck a little note on `x-ref` but can remove that if you want

calebporzio:
Thanks Kevin!

PR #988, opened 2021-01-07
corrected template roots error message

markfirmware:


src/utils.js
```diff
@@ -28,7 +28,7 @@ export function warnIfMalformedTemplate(el, directive) {
     if (el.tagName.toLowerCase() !== 'template') {
         console.warn(`Alpine: [${directive}] directive should only be added to <template> tags. See https://github.com/alpinejs/alpine#${directive}`)
     } else if (el.content.childElementCount !== 1) {
-        console.warn(`Alpine: <template> tag with [${directive}] encountered with multiple element roots. Make sure <template> only has a single child element.`)
+        console.warn(`Alpine: template with [${directive}] encountered with unexpected number of root elements. Make sure <template> has a single child element. `)
```
HugoDF, line 31: Any reason to remove the `<>`?
```suggestion
        console.warn(`Alpine: <template> tag with [${directive}] encountered with an unexpected number of root elements. Make sure <template> has a single child element. `)
```

markfirmware: One more change coming ...k

src/utils.js
```diff
@@ -28,7 +28,7 @@ export function warnIfMalformedTemplate(el, directive) {
     if (el.tagName.toLowerCase() !== 'template') {
         console.warn(`Alpine: [${directive}] directive should only be added to <template> tags. See https://github.com/alpinejs/alpine#${directive}`)
     } else if (el.content.childElementCount !== 1) {
-        console.warn(`Alpine: <template> tag with [${directive}] encountered with multiple element roots. Make sure <template> only has a single child element.`)
+        console.warn(`Alpine: <template> tag with [${directive}] encountered with an unexpected number of root elements. Make sure <template> has a single child element. `)
```
ryangjchandler, line 31: Should `child` be changed to `root` here so that it's inline with the README? Or perhaps the README should be changed to `child`.

ryangjchandler, line 31:
```suggestion
        console.warn(`Alpine: <template> tag with [${directive}] encountered with an unexpected number of root elements. Make sure <template> has a single root element. `)
```

ryangjchandler (commented):
Just a small one, wondering whether the terminology here should be changed so that it matches the README.

markfirmware:
> Just a small one, wondering whether the terminology here should be changed so that it matches the README.

I changed a bit of the readme to match the error message. Did I miss some?

markfirmware:
> > Just a small one, wondering whether the terminology here should be changed so that it matches the README.
> 
> I changed a bit of the readme to match the error message. Did I miss some?

@ryangjchandler @HugoDF are these what you mean?

    gitpod /workspace/alpine $ grep -i root README.md
    | [`$el`](#el) |  Retrieve the root component DOM node. |
    > Note: `x-if` must have a single element root inside the `<template></template>` tag.
    > Note: `x-for` must have a single element root inside of the `<template></template>` tag.
    `$el` is a magic property that can be used to retrieve the root component DOM node.

KevinBatdorf:
I think that comment came along with his suggested change just above it.

markfirmware:
Yeah sorry I am having a little trouble navigating the gh presentation. Does this mean there are no pending actions for me to take?

markfirmware:
@calebporzio ready to go. I wanted to compress to one commit but after thirty minutes of sesrching aimgavr up on that.

calebporzio:
Thanks everyone!

PR #1000, opened 2021-01-10
Attach el + expression to eval error

HugoDF:
The error can then be read using:

```js
// might need to listen to "error" as well for non-async eval errors
window.addEventListener('unhandledrejection', (event) => {
  console.log(event.reason.el, event.reason.expression, event.reason.toString())
})
```


calebporzio:
Thanks Hugo!

PR #1013, opened 2021-01-15
Change direction to be right to left

IMM9O:


calebporzio:
Thank you for this! Unfortunately, we just merged another PR that does this exact thing. Thanks again!

PR #1021, opened 2021-01-18
Allow integer n to be 0 in for-loop

pminne:
Not sure if there's a specific reason why `x-for="i in n"` isn't allowed with n = 0. Allowing it to be 0 doesn't break anything, corresponds with the documentation (0 = integer), and makes if more convient to dynamically create/remove DOM elements where at some point n can become 0 and all DOM elements should get removed.

Simple example case:
```javascript
<div x-data="
    {
        maxSeats: 3,
        seats: [],

        availableSeats() {
            return this.maxSeats - this.seats.length;
        },

        takeSeat(seatNumber) {
            this.seats.push(seatNumber);
        }
    }"
>

    <template x-for="seat in seats">
        <button type="button" style="width: 70px; height: 50px; border: 2px solid red; margin: 1rem;">TAKEN</button>
    </template>
    <template x-for="i in availableSeats()">
        <button type="button" @click="takeSeat(i)" style="width: 70px; height: 50px; border: 2px solid black; margin: 1rem;">FREE</button>
    </template>
</div>
```
This currently fails once `availableSeats` is 0. Wrapping the templates with a `x-if="availableSeats()"` doesn't work either since the inner templates don't get re-rendered when 0, leaving the previous situation in place.

Even though there may be an elegant way of implementing a case like the one above, I still think it makes sense to allow 0 in a for-loop.

### DIFF
--- a/src/directives/for.js
+++ b/src/directives/for.js
@@ -98,7 +98,6 @@ function evaluateItemsAndReturnEmptyIfXIfIsPresentAndFalseOnElement(component, e
     if (isNumeric(items)) {
         items = items > 0 ? Array.from(Array(items).keys(), i => i = 1) : []
     }
-    }
 
     return items
 }
--- a/src/directives/for.js
+++ b/src/directives/for.js
@@ -95,8 +95,8 @@ function evaluateItemsAndReturnEmptyIfXIfIsPresentAndFalseOnElement(component, e
     let items = component.evaluateReturnExpression(el, iteratorNames.items, extraVars)
 
     // This adds support for the `i in n` syntax.
-    if (isNumeric(items)) {
-        items = items > 0 ? Array.from(Array(items).keys(), i => i = 1) : []
+    if (isNumeric(items) && items >= 0) {
+        items = Array.from(Array(items).keys(), i => i + 1)
     }
 
     return items
--- a/src/directives/for.js
+++ b/src/directives/for.js
@@ -96,11 +96,8 @@ function evaluateItemsAndReturnEmptyIfXIfIsPresentAndFalseOnElement(component, e
 
     // This adds support for the `i in n` syntax.
     if (isNumeric(items)) {
-        if(items > 0) {
-            items = Array.from(Array(items).keys(), i => i + 1)
-        } else if(items === 0) {
-            items = []
-        }
+        items = items > 0 ? Array.from(Array(items).keys(), i => i = 1) : []
+    }
     }
 
     return items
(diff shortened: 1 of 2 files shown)


src/directives/for.js
```diff
@@ -95,8 +95,12 @@ function evaluateItemsAndReturnEmptyIfXIfIsPresentAndFalseOnElement(component, e
...
 
     // This adds support for the `i in n` syntax.
-    if (isNumeric(items) && items > 0) {
-        items = Array.from(Array(items).keys(), i => i + 1)
+    if (isNumeric(items)) {
+        if(items > 0) {
+            items = Array.from(Array(items).keys(), i => i + 1)
+        } else if(items === 0) {
+            items = []
+        }
```
ryangjchandler, line 103: You could probably turn this into a ternary instead, because the `else if` is irrelevant since we don't actually support negative numbers either.
```suggestion
    if (isNumeric(items)) {
        items = items > 0 ? Array.from(Array(items).keys(), i => i = 1) : []
    }
```

ryangjchandler (changes requested):
Only a small bit of feedback here - guess it's my fault for not adding support for 0 in the first place!

src/directives/for.js
```diff
@@ -95,8 +95,8 @@ function evaluateItemsAndReturnEmptyIfXIfIsPresentAndFalseOnElement(component, e
     let items = component.evaluateReturnExpression(el, iteratorNames.items, extraVars)
 
     // This adds support for the `i in n` syntax.
-    if (isNumeric(items) && items > 0) {
-        items = Array.from(Array(items).keys(), i => i + 1)
+    if (isNumeric(items)) {
+        items = items > 0 ? Array.from(Array(items).keys(), i => i = 1) : []
```
calebporzio, line 99: Why not: `items = Array.from(Array(items).keys(), i => i + 1)`?

Shouldn't need the ternary at all.

Also doesn't changing `+` to `=` cause a breaking change where each item in the iteration isn't incremented, but just set to the number `1`?

Please fix, then I'll merge

calebporzio: Also, surprised the tests passed with the `=`, maybe we need an extra test to ensure that ever item is an incrementing value

ryangjchandler: @calebporzio It should be `i + 1`, my laptop's keyboard is broken so I must have hit `=` instead of `+` in the suggestion.

The `i => i + 1` logic is just to start the counter from 1 instead of 0 so I'm guessing tests are still passing because we're not checking that the first item is `1`.

pminne: That was my initial thought, just make it `items >= 0`. But I found it cleaner with the if/else or ternary. Personal preference. Will make the change as suggested. Negative value will error then, which is fine I guess.

calebporzio:
Thanks!

PR #1054, opened 2021-01-29
feature: add support for debouncing x-on.away

ryangjchandler:
This PR is in response to #1053. It definitely seems like something that should be supported, so here you go.

The test suite wasn't playing very nice with me, I couldn't get the `.debounce` tests to work with the `.away` modifier, even though it's all working correctly in the browser.

I'll give it another go later on, but for now, here is the relevant logic.


calebporzio:
Thanks Ryan!

PR #1058, opened 2021-02-01
Don't register event listeners on clone - livewire/livewire#1995

austenc:
This was originally reported in the following Livewire issue:
https://github.com/livewire/livewire/issues/763

When components are cloned, event listeners are currently duplicated. This is because they're registered once in `initializeElement()` via `registerListeners()`, and again when the mutation observer picks them up. 

This PR changes the behavior of `Alpine.clone()` to no longer register event listeners.

**Questions:**
- Would it be preferred to create a new `cloneWithoutListeners()` function instead? I chose not to do this because adding another parameter to the `Alpine` constructor seemed gross. 
- Should we include a jest test for this? _(If so, direction / pointers welcomed!)_

If this is merged, we'll be able to bump the Alpine version used in Livewire's tests, but no changes will be required to Livewire itself. Thanks for considering!


calebporzio:
Thanks Austen!

Mushr0000m:
Hi, sorry but I think the x-init called multiples times with Livewire is still there. I updated my test from my livewire issue (https://github.com/livewire/livewire/issues/2520) to demonstrate it, just look at the console x-init is called each time the model changes by typing in the input : https://laravelplayground.com/#/snippets/47308c24-0b8f-445a-9fe1-641eeba4ef14

PR #1077, opened 2021-02-07
Add custom modifiers to x-model

PsyLogic:
My suggestion to add an API modifier for x-model, so we can add custom modifiers.

I put an example of the changes and their tests, any suggestion on how to use a global modifier so we can extend them as we need.

I didn't think too much to name the parse function so I can focus on the results.

### DIFF
--- a/src/directives/model.js
+++ b/src/directives/model.js
@@ -34,30 +34,52 @@ function generateModelAssignmentFunction(el, modifiers, expression) {
         } else if (el.type === 'checkbox') {
             // If the data we are binding to is an array, toggle its value inside the array.
             if (Array.isArray(currentValue)) {
-                const newValue = modifiers.includes('number') ? safeParseNumber(event.target.value) : event.target.value
+                const newValue = parseModifiers(modifiers,event.target.value)
                 return event.target.checked ? currentValue.concat([newValue]) : currentValue.filter(el => !checkedAttrLooseCompare(el, newValue))
             } else {
                 return event.target.checked
             }
         } else if (el.tagName.toLowerCase() === 'select' && el.multiple) {
-            return modifiers.includes('number')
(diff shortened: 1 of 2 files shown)


SimoTod:
I think the reverse modifier will also update the input when the component refreshes so after 2 times, it will be back to the original string, am I wrong? 
You'll probably need to hook into bind as well if you want to support this kind of modifiers so you can provide the right value when needed.

PsyLogic:
@SimoTod 
Well, if the init data of `foo` is `'bar'`, it will display `'bar'` until you type something (I don't know actually if it is a good thing or not to reverse the init value at the beginning )

For the `reverse` modifier, I would like to add  `lazy` modifier too (you know, that will prevent reversing the wrong string while typing)

```html
<div x-data="{ foo: 'bar' }">
    <input type="text" x-model.reverse.trim.lazy="foo"></input>
</div>
```

SimoTod:
What I mean is that you start from `bar`, you type f, it gets reverted to `frab` but it also shows `frab` in the input. If you then type o, I'll get converted to `obarf`. Is it the behaviour you wanted? 

I think it's a bit confusing and hard to handle, not sure it should be in the core (but you can add it using your api if the PR gets merged in). I was expecting something like you type and only the data model gets reverted. 

Just my 2 cents. I think Caleb is working on v3 so it may be hard to get a new feature into v2 (I like it though, it would open up to stuff like formatters).

PsyLogic:
Yes, that is right. Reverse example isn't a good example it was just for demonstration but as you say we can do string formating CC,IBAN,...
without actually importing third-party to do that kinf of stuff.
We could also use this kind of transformation using custom directives.

Giving that choice to devs to extend any part is a good thing.

calebporzio:
Thank you for the contribution. Going to hold off on adding this API though. Thank you.

Issue #909, opened 2020-12-01
readme guidance on putting x-if inside of x-for


markfirmware:
https://github.com/alpinejs/alpine#nesting-x-fors

Improve this to include x-if nested in x-for

Perhaps a generic statement on nesting \<template\> and also explicitly stating whether \<template\> is not-an-element even though it looks like one.

PR #1083, opened 2021-02-09
fix: misleading example in docs

### LINKED ISSUE #830: Mistake in x-for documentation?
In the README under  [x-for](https://github.com/alpinejs/alpine#x-for) it says:

> Note: x-for must have a single element root inside of the `<template></template>` tag.

However right above that it has an example that has multiple div tags immediately within the template tag:

```html
<template x-for="(item, index, collection) in items" :key="index">
    <!-- You can also reference "collection" inside the iteration if you need. -->
    <!-- Current item. -->
    <div x-text="item"></div>
    <!-- Same as above. -->
    <div x-text="collection[index]"></div>
    <!-- Previous item. -->
    <div x-text="collection[index - 1]"></div>
</template>
```

Am I understanding the Note incorrectly or is this a mistake in the docs?

Thanks

Seb

ryangjchandler:
This resolves #830. I think it's a valid point, anybody who wants to copy and paste this example into a fiddle to see how each thing works would be given an error / warning about multiple root nodes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -502,13 +502,15 @@ If you want to access the array object (collection) of the iteration, use the fo
 
 ```html
 <template x-for="(item, index, collection) in items" :key="index">
-    <!-- You can also reference "collection" inside the iteration if you need. -->
-    <!-- Current item. -->
-    <div x-text="item"></div>
-    <!-- Same as above. -->
-    <div x-text="collection[index]"></div>
-    <!-- Previous item. -->
-    <div x-text="collection[index - 1]"></div>
+    <div>
+        <!-- You can also reference "collection" inside the iteration if you need. -->
+        <!-- Current item. -->
+        <div x-text="item"></div>
+        <!-- Same as above. -->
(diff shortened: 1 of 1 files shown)


dtcooper:
Nice catch. I ran into this when I first tried `x-if` myself! Alternatively, it could be,

```html
<!-- You can also reference "collection" inside the iteration if you need. -->
<template x-for="(item, index, collection) in items" :key="index">
    <!-- Current item. -->
    <div x-text="item"></div>
</template>
<template x-for="(item, index, collection) in items" :key="index">
    <!-- Same as above. -->
    <div x-text="collection[index]"></div>
</template>
<template x-for="(item, index, collection) in items" :key="index">
    <!-- Previous item. -->
    <div x-text="collection[index - 1]"></div>
</template>
```

HugoDF:
Does this tangentially solve https://github.com/alpinejs/alpine/issues/909?

calebporzio:
Good call, thanks Ryan

PR #1085, opened 2021-02-11
Optimize isBooleanAttr with Set()

EricPKerr:
Set() is more performant than Array.includes().  Also supported on older browsers than some other built-ins used in Alpine like Array.from() per https://caniuse.com

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -181,18 +181,17 @@ export function parseHtmlAttribute({ name, value }) {
     }
 }
 
-export function isBooleanAttr(attrName) {
-    // As per HTML spec table https://html.spec.whatwg.org/multipage/indices.html#attributes-3:boolean-attribute
-    // Array roughly ordered by estimated usage
-    const booleanAttributes = [
-        'disabled','checked','required','readonly','hidden','open', 'selected',
-        'autofocus', 'itemscope', 'multiple', 'novalidate','allowfullscreen',
-        'allowpaymentrequest', 'formnovalidate', 'autoplay', 'controls', 'loop',
-        'muted', 'playsinline', 'default', 'ismap', 'reversed', 'async', 'defer',
-        'nomodule'
-    ]
+// As per HTML spec table https://html.spec.whatwg.org/multipage/indices.html#attributes-3:boolean-attribute
+const BOOLEAN_ATTRIBUTES = new Set([
+    'disabled', 'checked', 'required', 'readonly', 'hidden', 'open', 'selected',
+    'autofocus', 'itemscope', 'multiple', 'novalidate', 'allowfullscreen',
+    'allowpaymentrequest', 'formnovalidate', 'autoplay', 'controls', 'loop',
+    'muted', 'playsinline', 'default', 'ismap', 'reversed', 'async', 'defer',
+    'nomodule'
+])
 
-    return booleanAttributes.includes(attrName)
+export function isBooleanAttr(attrName) {
+  return BOOLEAN_ATTRIBUTES.has(attrName)
 }
 
 export function replaceAtAndColonWithStandardSyntax(name) {


HugoDF:
Hey @EricPKerr have you got benchmarks for this?

I've seen a lot of PRs doing this kind of work not get merged because the performance benefit isn't measurable/a benchmark isn't shared.

EricPKerr:
Yup - it depends what the average position of the accessed attribute is since it's linear vs log.  If the average position is 5th (index 4) or less the Array.includes is better, otherwise Set is better.

https://gist.github.com/EricPKerr/3283c71ee06e90daca4cc071662f1acf

KevinBatdorf:
I pasted that into my browser console and got this:
```
Test: withSet 	 Index: 0 	 Time: 2067 ms
Test: withIn 	 Index: 0 	 Time: 2031 ms
Test: withOut 	 Index: 0 	 Time: 1808 ms
Test: withSet 	 Index: 1 	 Time: 2985 ms
Test: withIn 	 Index: 1 	 Time: 2086 ms
Test: withOut 	 Index: 1 	 Time: 2126 ms
Test: withSet 	 Index: 2 	 Time: 2914 ms
Test: withIn 	 Index: 2 	 Time: 2084 ms
Test: withOut 	 Index: 2 	 Time: 2250 ms
Test: withSet 	 Index: 3 	 Time: 3945 ms
Test: withIn 	 Index: 3 	 Time: 2680 ms
Test: withOut 	 Index: 3 	 Time: 2784 ms
Test: withSet 	 Index: 4 	 Time: 2378 ms
Test: withIn 	 Index: 4 	 Time: 2498 ms
Test: withOut 	 Index: 4 	 Time: 2662 ms
Test: withSet 	 Index: 5 	 Time: 2407 ms
Test: withIn 	 Index: 5 	 Time: 2773 ms
Test: withOut 	 Index: 5 	 Time: 2845 ms
Test: withSet 	 Index: 6 	 Time: 3421 ms
Test: withIn 	 Index: 6 	 Time: 3342 ms
Test: withOut 	 Index: 6 	 Time: 3549 ms
Test: withSet 	 Index: 7 	 Time: 2875 ms
Test: withIn 	 Index: 7 	 Time: 3118 ms
Test: withOut 	 Index: 7 	 Time: 3365 ms
Test: withSet 	 Index: 8 	 Time: 3278 ms
Test: withIn 	 Index: 8 	 Time: 3695 ms
```

EricPKerr:
The advantage of the Set approach is a lot more noticeable when you're looking up attributes at the end of the list..

```
Test: withSet 	 Index: 24 	 Time: 1387 ms
Test: withIn 	 Index: 24 	 Time: 8785 ms
Test: withOut 	 Index: 24 	 Time: 8798 ms
```

HugoDF:
> The advantage of the Set approach is a lot more noticeable as you get to attributes near the end of the list..
> 
> ```
> Test: withSet 	 Index: 24 	 Time: 1387 ms
> Test: withIn 	 Index: 24 	 Time: 8785 ms
> Test: withOut 	 Index: 24 	 Time: 8798 ms
> ```

Yeah I think that's compelling, thanks for adding a benchmark.

This code is probably quite hot (gets called on every binding update).

Up to @calebporzio to pitch in and merge or not now 😁.

ryangjchandler:
@EricPKerr I'm pretty sure the IE version needs a polyfill for this still. Based on the compatibility tables on MDN, IE11 doesn't support `new Set(iterable)` which is exactly what's being used here.

SimoTod:
Set is more performant but we needed  200 millions of iterations to appreciate the difference. I haven't checked but if we need a polyfill, it won't be worth to add more bytes for something that users won't notice imo.

PR #1106, opened 2021-02-17
fix: x-cloak css in README

### LINKED ISSUE #1091: [x-cloak] should be reccomended as { display: none !important; } no?
While using tailwindcss for my project, I just realized that my usage of [x-cloak] had been meaningless because any element that had `.flex` or `.grid` applied to them would override the x-cloak atttribute.

ryangjchandler:
Closes #1091. Think this is a fair change since `x-cloak` should always have precedence over other CSS `display` properties.

### DIFF
--- a/README.md
+++ b/README.md
@@ -633,7 +633,9 @@ The object keys are the directives (Can be any directive including modifiers), a
 
 ```html
 <style>
-    [x-cloak] { display: none; }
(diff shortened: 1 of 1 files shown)


calebporzio:
Yep, makes sense. Thanks Ryan!

PR #1117, opened 2021-02-19
Add README translation of Simplified Chinese

noahziheng:
Thanks for @calebporzio and all contributors to create the amazing JS Framework.
I’m a Chinese native speaker and I have translated README to simplified Chinese(简体中文).
Hope to make Chinese Developer to learn Alpine.js easily~


calebporzio:
That is so kind of you! Thanks @noahziheng!

PR #1160, opened 2021-03-05
Fix nested alpine with nested livewire

### LINKED ISSUE #1144: Feedback wanted: Support for components added using Livewire
Right now, Alpine only works with Livewire components that were present on the initial load. Alpine components added using Livewire components which were requested later seem to **initialize before Livewire**, which makes the entire JS runtime break and stop executing.

I found a way to solve this, by essentially blocking Alpine component updates/inits while Livewire does its job. I think this is the right approach.

However, a few changes are needed to both project's JS suite, and I have no idea how to write tests for this, so I'm submitting this as an issue.

### Waiting for Livewire

```diff
updateElements(rootEl, extraVars = () => {}) {
+   let update = () => {
        this.walkAndSkipNestedComponents(rootEl, el => {
            // Don't touch spawns from for loop (and check if the root is actually a for loop in a parent, don't skip it.)
            if (el.__x_for_key !== undefined && ! el.isSameNode(this.$el)) return false
    
            this.updateElement(el, extraVars)
        }, el => {
            el.__x = new Component(el)
        })
    
        this.executeAndClearRemainingShowDirectiveStack()
    
        this.executeAndClearNextTickStack(rootEl)
+   }

+    if (window.LivewireManagingDOM) {
+        window.LivewireManagingDOM.then(update)
+    } else {
+        update();
+    }
}

initializeElements(rootEl, extraVars = () => {}, shouldRegisterListeners = true) {
+    let initialize = () => {
        this.walkAndSkipNestedComponents(rootEl, el => {
                // Don't touch spawns from for loop
                if (el.__x_for_key !== undefined) return false; // Don't touch spawns from if directives
  
                if (el.__x_inserted_me !== undefined) return false;
  
                this.initializeElement(el, extraVars, shouldRegisterListeners);
            }, el => {
                el.__x = new Component(el);
            });
            this.executeAndClearRemainingShowDirectiveStack();
            this.executeAndClearNextTickStack(rootEl);
+    }

+    if (window.LivewireManagingDOM) {
+        window.LivewireManagingDOM.then(initialize)
+    } else {
+        initialize()
+    }
}
```

### Skipping transitions

To support `$wire` updates etc, Livewire runs Alpine's `updateElements()` after it finishes making changes to the DOM.

This results in `show="false"` transitions thinking that they were user-initiated, so as the element is removed from the DOM, it runs `transitionOut()`.

```diff
function transitionIn(el, show, reject, component, forceSkip = false) {
+    if (window.skipAlpineTransitions) return show();

function transitionOut(el, hide, reject, component, forceSkip = false) {
+    if (window.skipAlpineTransitions) return hide();
```

## Livewire changes

The `refreshAlpineAfterEveryLivewireRequest()` method should be replaced with logic that works like this:

```js
window.Livewire.hook('message.sent', () => {
    window.LivewireManagingDOM = new Promise((resolve, reject) => {
        window.Livewire.hook('message.processed', () => {
            window.LivewireManagingDOM = undefined;

            resolve(true);
        });

        window.Livewire.hook('message.failed', () => {
            window.LivewireManagingDOM = undefined;

            reject();
        });
    }).then(() => {
        window.skipAlpineTransitions = true;
        window.LivewireManagedAlpineComponents.forEach(component => {
            component.updateElements(component.$el);
        });
        window.skipAlpineTransitions = false;
    });
});

window.LivewireManagedAlpineComponents = [];

window.Alpine.onComponentInitialized(component => {
    let livewireEl = component.$el.closest('[wire\\:id]')

    if (livewireEl && livewireEl.__livewire) {
        window.LivewireManagedAlpineComponents.push(component);
    }
})
```

This creates a Promise when a LW message is sent, and resolves it when a response is **fully processed**. The Alpine logic was moved to `.then()`, meaning that **Alpine will only execute after LW is done with its changes**.

It also runs the original `refreshAlpineAfterEveryLivewireRequest()` logic, but it does so *after* the components have been properly initialized first.

## More context

The variable names used in this code are obviously POC. I think it'd make sense to make this non-LW specific in Alpine, and then hook into that logic from LW.

And the `window.skipAlpineTransitions` should probably be a property on `window.Alpine`.

cc @joshhanley

joshhanley:
This fixes an issue where if you have a Livewire component which has an Alpine component in it, that has nested Livewire components that also have Alpine in them. Adding a new nested Livewire component dynamically triggers an Alpine error
> Uncaught (in promise) TypeError: Cannot read property '$wire' of undefined

See this Livewire PR livewire/livewire#2517, which is a failing test, and issues linked in that PR for more details.

What is happening, is Livewire calls clone on Alpine's elements in `alpinifyElementsForMorphdom()` here
[livewire/livewire/js/component/SupportAlpine.js#L129](https://github.com/livewire/livewire/blob/b647ac63378ea02bdac975f70a4e57aea84bda0b/js/component/SupportAlpine.js#L129)

But this method is called in the `onBeforeElUpdated` morphdom hook here 
[livewire/livewire/js/component/index.js#L459](https://github.com/livewire/livewire/blob/b647ac63378ea02bdac975f70a4e57aea84bda0b/js/component/index.js#L459)

When Alpine's component is cloned it then tries to initialise any nested components (which can contain $wire calls), but this happens before the nested Livewire components are ready (as it's running in the `onBeforeElUpdated` hook above). This then causes the above error.

This PR stops nested components from initialising at the clone stage, so that they are initialised later.

✅ I've run all the Alpine tests locally after this change and they all pass.
✅ I also ran all the Livewire tests (using this Alpine modification) including the failing on in PR livewire/livewire#2517 and they all passed.

I checked through both code bases and the only call to the clone method is in Livewire's `SupportAlpine.js` listed above, so I see no implications for not newing up nested component's during a clone, as the dusk test proves that they do get initialised later.

Fixes #1144 

Hope this helps!

### DIFF
--- a/dist/alpine-ie11.js
+++ b/dist/alpine-ie11.js
@@ -7358,7 +7358,7 @@
 
       this.initializeElements(this.$el, function () {
         _newArrowCheck(this, _this);
-      }.bind(this), componentForClone ? false : true); // Use mutation observer to detect new elements being added within this component at run-time.
+      }.bind(this), componentForClone); // Use mutation observer to detect new elements being added within this component at run-time.
       // Alpine's just so darn flexible amirite?
 
       this.listenForNewElementsToInitialize();
@@ -7511,7 +7511,7 @@
         var extraVars = arguments.length > 1 && arguments[1] !== undefined ? arguments[1] : function () {
           _newArrowCheck(this, _this10);
         }.bind(this);
-        var shouldRegisterListeners = arguments.length > 2 && arguments[2] !== undefined ? arguments[2] : true;
+        var componentForClone = arguments.length > 2 && arguments[2] !== undefined ? arguments[2] : false;
         this.walkAndSkipNestedComponents(rootEl, function (el) {
           _newArrowCheck(this, _this10);
 
@@ -7519,11 +7519,11 @@
           if (el.__x_for_key !== undefined) return false; // Don't touch spawns from if directives
 
           if (el.__x_inserted_me !== undefined) return false;
-          this.initializeElement(el, extraVars, shouldRegisterListeners);
+          this.initializeElement(el, extraVars, componentForClone ? false : true);
         }.bind(this), function (el) {
           _newArrowCheck(this, _this10);
 
-          el.__x = new Component(el);
+          if (!componentForClone) el.__x = new Component(el);
         }.bind(this));
         this.executeAndClearRemainingShowDirectiveStack();
         this.executeAndClearNextTickStack(rootEl);
--- a/dist/alpine.js
+++ b/dist/alpine.js
@@ -1514,7 +1514,7 @@
       // event listeners are registered (the mutation observer will take care of them)
 
 
-      this.initializeElements(this.$el, () => {}, componentForClone ? false : true); // Use mutation observer to detect new elements being added within this component at run-time.
+      this.initializeElements(this.$el, () => {}, componentForClone); // Use mutation observer to detect new elements being added within this component at run-time.
       // Alpine's just so darn flexible amirite?
 
       this.listenForNewElementsToInitialize();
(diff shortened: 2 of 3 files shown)


ryangjchandler:
@calebporzio Is there any update on this PR? It fixes a major issue with nested Alpine component's that use `$wire` in Filament and we're currently using Josh's dist/ files to work around it.

danharrin:
This fixes a ton of serious Livewire issues, thanks a lot Josh! 👍

PH7-Jack:
this solves https://github.com/alpinejs/alpine/pull/1160

thanks

calebporzio:
Thanks @joshhanley for this fix!

joshhanley:
No worries!

PR #1165, opened 2021-03-09
Add error handler override

patinthehat:
This PR adds the ability to override the default error handler:
```js
const handleError = (el, expression, error) => {
    console.warn(`Alpine Error: "${error}"\n\nExpression: "${expression}"\nElement:`, el);

    if (! isTesting()) {
        Object.assign(error, { el, expression })
        throw error;
    }
}
```
Not having any way to override this error handler is an issue as I'd like to intercept the error and display it elsewhere.  Namely, within a debugging app with an [integration library for Ray](https://github.com/permafrost-dev/alpinejs-ray).

This PR adds `Alpine.setErrorHandler(handler)`, allowing external applications and scripts to use their own error handler.  Calling `Alpine.setErrorHandler(null)` causes the default error handler to be used again.

I feel that this is a worthwhile addition to alpine.js, without changing its default behavior.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,5 +1,5 @@
 import Component from './component'
-import { domReady, isTesting } from './utils'
+import { domReady, isTesting, setErrorHandler } from './utils'
 
 const Alpine = {
     version: process.env.PKG_VERSION,
@@ -107,6 +107,10 @@ const Alpine = {
 
     addMagicProperty: function (name, callback) {
         this.magicProperties[name] = callback
+    },
+
+    setErrorHandler: function(handler) {
+        setErrorHandler(handler);
     },
 
     onComponentInitialized: function (callback) {
--- a/src/utils.js
+++ b/src/utils.js
@@ -1,4 +1,6 @@
 
+let errorHandler = null;
+
 // Thanks @stimulus:
 // https://github.com/stimulusjs/stimulus/blob/master/packages/%40stimulus/core/src/application.ts
 export function domReady() {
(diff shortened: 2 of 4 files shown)


HugoDF:
@patinthehat the errors have all the data you need, no need for a custom error handler (although up to Caleb if he wants to support this)

See this PR in devtools

https://github.com/alpine-collective/alpinejs-devtools/pull/167

You need to attach listeners for unhandled rejection and error and you'll have access to the element and expression

Feel free to tag me on the alpine ray repo if you need a hand

patinthehat:
@HugoDF Doesn't `tryCatch()` always call `handleError()`, regardless if an error is thrown? Am I missing something?

HugoDF:
> @HugoDF Doesn't `tryCatch()` always call `handleError()`, regardless if an error is thrown? Am I missing something?

It does but the error is re-thrown (after adding `el` and `expression`, & logging a warning), so a global error does get thrown

HugoDF:
I've replied to you on the alpine-ray repo, in JS you could just overwrite the console and ignore Alpine errors using a regex

calebporzio:
Thanks for weighing in with Alternatives @HugoDF 

Holding off on this for now. Thanks for contributing!

HugoDF:
> Thanks for weighing in with Alternatives @HugoDF 
> 
> Holding off on this for now. Thanks for contributing!

I think @patinthehat managed to get part of this working without the changes in this PR 🙂

PR #1166, opened 2021-03-09
Fetch x-init after running onBeforeComponentInitialized

stancl:
This makes it possible to add x-init logic to all elements programmatically.

No behavior is broken or changed, we just read the attribute after running the "before initialized" callbacks.

Also makes sense for "on before initialized" to be able to affect how the initialization process works.

### DIFF
--- a/src/component.js
+++ b/src/component.js
@@ -16,7 +16,6 @@ export default class Component {
 
         const dataAttr = this.$el.getAttribute('x-data')
         const dataExpression = dataAttr === '' ? '{}' : dataAttr
-        const initExpression = this.$el.getAttribute('x-init')
 
         let dataExtras = {
             $el: this.$el,
@@ -82,6 +81,8 @@ export default class Component {
         this.showDirectiveLastElement
 
         componentForClone || Alpine.onBeforeComponentInitializeds.forEach(callback => callback(this))
+        
+        const initExpression = this.$el.getAttribute('x-init')
 
         var initReturnedCallback
         // If x-init is present AND we aren't cloning (skip x-init on clone)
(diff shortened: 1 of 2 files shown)


ryangjchandler:
> This makes it possible to add x-init logic to all elements programmatically.
> 
> No behavior is broken or changed, we just read the attribute after running the "before initialized" callbacks.
> 
> Also makes sense for "on before initialized" to be able to affect how the initialization process works.

Have you got an example of how you might use this?

ryangjchandler:
> Hmm, that's interesting. Would that run before the initial render or after it? It's true that I could run a function at any point in the lifecycle, but `init()` specifically is helpful since it can set data that will be used in the first render.

Yeah, it runs _after_ the first call to `this.initialElements` so doesn't do quite the same thing as `x-init`.

SimoTod:
You can use onBeforeComponentInitialised to simulate x-init="init()" and onComponentInitialised to simulate x-init ="() => init()". I personally prefer the latter since it runs quick enough for users not be noticing the difference and you have full access to other helpers but it's up to you.

stancl:
> quick enough for users not be noticing the difference

Makes sense. I guess there are still edge cases where you need usual `x-init` behavior, since you may require some data from that expression to be available in the template. But also seems good enough for most cases.

calebporzio:
@stancl - can you write a test for this? I'm comfortable merging if so.

stancl:
@calebporzio Done. Here it's failing without the patch: https://github.com/stancl/alpine/runs/2170112142#step:4:143

calebporzio:
Thanks @stancl !

Issue #1171, opened 2021-03-11
Duplicate event handlers being added to elements with ids during LiveView update.


cthulahoops:
I've been running into a weird bug with event handlers being added more than once to elements when using AlpineJS with LiveView.

This is a fairly minimal demonstration:

```
<script>
function add_message(dispatch) {
  event.preventDefault()
  let message = document.getElementById("message").value;
  dispatch('add_message', {message: message});
}
</script>

<ul id="to-input-div" x-data="{}" phx-hook="PushEvent" phx-push-event="add_message">
<%= for message <- @messages do %>
  <li><%= message %></li>
<% end %>
  <input id="message" x-on:keydown.enter="add_message($dispatch)">
  <button x-on:click="add_message($dispatch)">Add (no id)</button>
  <button id="unreferenced" x-on:click="add_message($dispatch)">Add (with id)</button>
</ul>
```
When you enter a message and hit enter it gets added once, and a second copy of the event handler is added. The second message is added twice, the third four times, the fourth eight.... Clicking the button with an id does the same thing, but clicking the button without id only adds it once. (Checking registered event handlers confirms that an event handler has been added for every message added to the list.)

This is the `phx-hook`:

```
Hooks.PushEvent = {
  mounted() {
    const eventName = this.el.attributes['phx-push-event'].value;
    this.el.addEventListener(eventName, (event) => { this.pushEvent(eventName, event.detail) })
  }
}
```

And I have the callback calling `Alpine.clone` from `onBeforeElUpdated`.

(Running alpinejs 2.8.1 and phoenix_live_view 0.15.4).

cthulahoops:
The difference in behaviour occurs because morphdom uses the node's id to identify nodes that should be moved rather than deleted and reinserted. I think the bug occurs because element moves aren't distinguished from element creations.

With this in mind, I can write an version that triggers the issue without liveview:

```
<div id="to-input-div" x-data="{'x': 0}">
  <span x-text="x">0</span>
  <button id="a" x-on:click="x += 1; $el.appendChild(document.getElementById('a'))">A</button>
  <button id="b" x-on:click="x += 1; $el.appendChild(document.getElementById('b'))">B</button>
  <button id="c" x-on:click="x += 1; $el.appendChild(document.getElementById('c'))">C</button>
</div>
```

I expect that clicking a button will increment the counter and move that button to the end. When the buttons get moved duplicate event handlers get attached and the counter starts to go up in steps of more than one.

[...]

PR #1175, opened 2021-03-14
Remove stale event handlers when nodes are readded to the document.

cthulahoops:
Fixes: #1171

I'm not familiar enough with the project to know what I need to do to test this or if it's the right approach, but it does appear to resolve the problem for my simple example.

### DIFF
--- a/dist/alpine-ie11.js
+++ b/dist/alpine-ie11.js
@@ -6929,6 +6929,8 @@
       event = camelCase(event);
     }
 
+    var node_add_count = el.__x_node_add_count;
+
     var _handler2, listenerTarget;
 
     if (modifiers.includes('away')) {
@@ -6964,6 +6966,11 @@
             listenerTarget.removeEventListener(event, _handler2, options);
             return;
           }
+        }
+
+        if (el.__x_node_add_count !== node_add_count) {
+          listenerTarget.removeEventListener(event, _handler2, options);
+          return;
         }
 
         if (isKeyEvent(event)) {
@@ -7806,6 +7813,7 @@
                   return;
                 }
 
+                node.__x_node_add_count = (node.__x_node_add_count || 0) + 1;
                 this.initializeElements(node);
               }.bind(this));
             }
--- a/dist/alpine.js
+++ b/dist/alpine.js
@@ -863,6 +863,7 @@
       event = camelCase(event);
     }
 
+    const node_add_count = el.__x_node_add_count;
     let handler, listenerTarget;
 
     if (modifiers.includes('away')) {
@@ -892,6 +893,11 @@
             listenerTarget.removeEventListener(event, handler, options);
             return;
           }
+        }
+
+        if (el.__x_node_add_count !== node_add_count) {
+          listenerTarget.removeEventListener(event, handler, options);
+          return;
         }
 
         if (isKeyEvent(event)) {
@@ -1805,6 +1811,7 @@
                 return;
               }
 
+              node.__x_node_add_count = (node.__x_node_add_count || 0) + 1;
               this.initializeElements(node);
             });
           }
--- a/src/component.js
+++ b/src/component.js
@@ -411,6 +411,8 @@ export default class Component {
                             return
                         }
 
+                        node.__x_node_add_count = (node.__x_node_add_count || 0) + 1
+
                         this.initializeElements(node)
                     })
                 }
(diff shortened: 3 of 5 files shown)


HugoDF:
@cthulahoops you can add a test with your example from the issue in the `for.spec.js` file I think

cthulahoops:
`for.spec.js` rather than `on.spec.js`?

HugoDF:
> @HugoDF `for.spec.js` rather than `on.spec.js`?

Oops yeah sorry `on.spec.js`

cthulahoops:
At the top of `on.spec.js` we replace the `MutationObserver`with an empty class so the bug doesn't happen in the tests. It looks like the tests still pass if I remove that.

HugoDF:
> At the top of `on.spec.js` we replace the `MutationObserver`with an empty class so the bug doesn't happen in the tests. It looks like the tests still pass if I remove that. That is the existing tests pass, and it becomes possible to write a test to reproduce my bug.

That makes a lot of sense.

Are you saying you can write a failing test if mutation observer isn't faked? As far as I remember, I don't think the mutation observer works properly in tests, it's not implemented 1-1 with browsers.

I think I've seen other tests where we "fake" the mutation observer triggering, that might be an option as well.

cthulahoops:
Yes! I can write the failing test, so the mutation observer works well enough for this test at least.

I can't get the passing test to run locally. It seems to be refusing to run my changed code, it just runs the original whatever I do. It seems to run and pass here though.

HugoDF:
> Yes! I can write the failing test, so the mutation observer works well enough for this test at least.
> 
> I can't get the passing test to run locally. It seems to be refusing to run my changed code, it just runs the original whatever I do. It seems to run and pass here though.

Yeah for it to work locally you need to rebuild the files (& maybe link? I forget)

If they're working here it's good enough though 🙂

cthulahoops:
This branch is working well for me. Is there anything else needed?

calebporzio:
Thanks for the contribution.

For me to feel comfortable merging I will need a more clear and succinct explanation/demonstration of the problem. Ideally without needing to understand LiveView or morphdom.

HugoDF:
@calebporzio I think this sample code illustrates it pretty well

https://github.com/alpinejs/alpine/issues/1171#issuecomment-798204177

Might be an edge case for Alpine by itself (I'm not sure why you would reinsert existing Nodes with Alpine) but might cause issues with libraries that attempt to reuse DOM nodes?

cthulahoops:
I expect you'd encounter this issue in pure alpine in applications that involve reordering a list. I can imagine a todo list that allows the user to drag and drop nodes to order them. I was also working on an (unrelated) app with a list of conversations and conversations moved to the top of the list when new messages arrive, which I think would have triggered this issue if it were in alpine.

I do agree this is going to be a more common problem when using liveview or the like. I ran into it with a multi-input box for entering e-mail addresses. The container has a div for each e-mail address and then an input box at the end. When you add a new e-mail address, liveview readds all the elements in the correct order - including the input box.

It's not too hard to restructure things to work around this problem if you understand it, but it's very confusing and hard to understand.

The bug occurs because MutationObserver doesn't tell you whether a node is being added for the first time or re-added, so alpine will always add the event handler even if it was there already. I think removing stale event handlers is the correct fix rather than trying to detect readded nodes in order to avoid re-adding the event handlers. This way things should continue to work if event handlers are added or removed from nodes.

Is that helpful? I'm happy to try and explain further.

calebporzio:
Thanks for the reference @HugoDF 

Yeah, this makes sense. I ran into this when re-implementing the mutation observer for V3 of Alpine.

Here's the implementation I came up with: 
image

Seeing as this will be fixed in the next version in a more ideal way, I think I'm comfortable pulling this in for now. Thank you!

cthulahoops:
:tada: :smile: 

I think replacing the callback with the new one in this case is likely to be more correct than skipping.

For example, I have code that looks a bit like this:

```
<ul x-data="active = 0">
<%= for {suggestion, idx} <- Enum.with_index(@suggestions) do %>
   <li x-bind:mouseover="active = <%= idx %>" id="suggestion-<%= suggestion.id %>">
<% end %>
</ul>
```

When suggestions are added or removed the index of each element changes, and the callback needs replacing. I could implement this in lots of ways to avoid this, but I think things are likely to get confusing if callbacks aren't updated.

Thanks for merging!

PR #1189, opened 2021-03-17
Add charset meta tag to the examples

JunaidQadirB:
Due to the missing charset meta tag, the browser (Firefox in my case) throws this error:
`The character encoding of the HTML document was not declared. The document will render with garbled text in some browser configurations if the document contains characters from outside the US-ASCII range. The character encoding of the page must be declared in the document or in the transfer protocol.`

[Here's an explanation](https://www.w3.org/International/questions/qa-html-encoding-declarations) from W3C regarding declaring the `charset` meta tag.

### DIFF
--- a/examples/index.html
+++ b/examples/index.html
@@ -1,5 +1,6 @@
 <html>
     <head>
+        <meta charset="utf-8"/>
         <style>
             .hidden { display: none; }
             [x-cloak] { display: none; }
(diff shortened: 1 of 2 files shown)


calebporzio:
Thanks!

JunaidQadirB:
Pleasure!

PR #1195, opened 2021-03-25
TypeScript definitions

rbeard0330:
This PR includes a basic set of TS type defs intended to be used for developers creating Alpine components in TypeScript.  This is problematic currently because calls like `this.$el` are understood by `tsc` to be references to non-existent properties, since TS does not see those properties being added to the component by Alpine.  As a result, in addition to specifying the types of the magic properties, there needs to be a workaround to convince the compiler that those properties exist.  The approach taken here is to add an empty object as a property of the `Alpine` object, and then to assert in the type definitions that the object includes the magic properties.  The intended usage is for the user to spread the dummy object into their component, thereby tricking the compiler into believing that the properties exist.  This is a bit of a hack, but I think the opportunity for mischief is minimal.  (The obvious ways to encounter problems would be to try to use the dummy object directly or else to create a component in JS and try to use it before Alpine has a chance to finish initializing it--neither of these seem like normal things to do.)

The other part of this contribution is to provide extensible types for `$event` and `$dispatch`.  The intended use for these types is in the signatures of functions that get passed `$event` or `$dispatch` from HTML expressions.  There are generic types for `$event` and `$dispatch` that reflect the actual behavior of those properties, but users can also specify more specific types to reflect the types of events that are expected to be received/dispatched by the component.  For example, if a method is supposed to dispatch a `foo-event` event with a `number` in the `detail` field, and the type of that event is defined by the user as `FooEvent`, then the type assigned to `$dispatch` in the method parameter list can be given as `$Dispatch<FooEvent>`, which allows TypeScript to type-check the way `$dispatch` is invoked to match the intended semantics.  There's an example in the README.

A couple notes:

1.  There's another package ([alpine-typescript](https://github.com/LeanAdmin/alpine-typescript)) that provides somewhat similar functionality.  My reasons for not sending this PR in their direction are: 1) I think it would be nice to have TS support as part of the core framework rather than an external library, and 2) that project also includes some syntactic reworking that ideally should be separate from good typing.
2.  Testing is not fully baked yet.  For development purposes, I put together a sample component that runs through a lot of things that should/should not be type errors.  Running tsc from the project root should successfully compile that file using strict settings.  I'm sure there's a smoother way to handle this long-term, but I didn't want to start adding a bunch of test dependencies without getting feedback first.

Of course, I'm happy to rework any of this PR as needed--just point me in the right direction.  I'm a big fan of the framework and hope I can make a contribution.

Best,
Robert


rbeard0330:
I just pushed some changes that remove the dummy object `Alpine._BASE` and the associated type definition.  This approach didn't work in practice, since the code defining components wouldn't have necessarily have access to the Alpine object.  The suggested alternative is for the user to provide an empty object and the type assertion.  This is a bit less ergonomic, but it's probably clearer to the user what's going on.

calebporzio:
Thanks for the PR @rbeard0330, however, as a non TypeScript user myself I don't want to be responsible for checking the validity of what you've done and maintaining it ongoing. Therefore, I'm closing this. Thanks for being willing to contribute though!

snmishra:
@rbeard0330 It's unfortunate that Typescript support was rejected. How about uploading to the [DefinitelyTyped](https://definitelytyped.org/)?

lbineau:
Any news on typescript support? https://definitelytyped.org/ would be fine.

wtho:
I gave it a shot and created a PR on definitelytyped
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/58130

I have to wait for this DefinitelyTyped-tools PR to be merged before to reuse types from `@vue/reactivity`.
https://github.com/microsoft/DefinitelyTyped-tools/pull/376

Anyway, in the meantime it would be great if some of you could review or even take ownership of the types with me. I am open for improvements in the types. As the **typescript-bot** in the PR phrases it:

> :bell: @wtho — you're the only owner, but it would still be good if you find someone to review this PR in the next few days, otherwise a maintainer will look at it. (And if you do find someone, maybe even recruit them to be a second owner to make future changes easier...)

Cheers!

PR #1204, opened 2021-03-28
Add: tabindex example in x-bind directive

kantbtrue:


calebporzio:
Thanks for the contribution! However, I don't think this adds enough unique value to be included in the README. Thanks again.

Issue #566, opened 2020-06-11
Performance issues when used with lots of DOM elements


SebastianSchoeps:
**Update after some digging of different people (thanks!) into the issue:**
Just by embedding Alpine.js (without even using it) the performance gets slow when using "many" DOM elements. "Many" depends on the computer's performance - in my case the rendering took 11s (with Alpine) vs. 3s (without Alpine) for 3,000 DOM elements on an old i5-3210M I had available.
This is caused by Alpine using `querySelectorAll` and iterating over all elements.
A workaround is possible by using `deferLoadingAlpine`.

-----------------

I run into strange issues when using Alpine with bpmn-js (https://github.com/bpmn-io/bpmn-js). Just by embedding Alpine the performance for loading a tree model goes down from 2s to 8s (on a slow computer).

I created two examples:
https://nanuc.io/alpine-performance/with-alpine.html
https://nanuc.io/alpine-performance/without-alpine.html

The only difference is that at the first example Alpine is embedded (see below).

I know that this is a very specific use case that probably no one else on this world has encountered yet... I'm reporting it anyway because it might be the case that Alpine affects the performance of other libraries too.

Unfortunately the only solution for me seems to replace Alpine by something else for now because my knowledge to dive into this and solve this is far too little.

Sebastian

```
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="utf-8" />
    <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/bpmn-js@7.2.0/dist/assets/diagram-js.css">
    <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/bpmn-js@7.2.0/dist/assets/bpmn-font/css/bpmn.css">

    <!-- THE NEXT LINE KILLS THE PERFORMANCE -->
    <script src="https://cdn.jsdelivr.net/gh/alpinejs/alpine@v2.x.x/dist/alpine.min.js" defer></script>
</head>
<body>
<div>
    <div id="canvas" style="height: 100vh"></div>
</div>

<script src="https://cdnjs.cloudflare.com/ajax/libs/axios/0.19.2/axios.min.js" integrity="sha256-T/f7Sju1ZfNNfBh7skWn0idlCBcI3RwdLSS4/I7NQKQ=" crossorigin="anonymous"></script>
<script src="https://unpkg.com/bpmn-js@7.2.0/dist/bpmn-modeler.production.min.js"></script>


<script>
    modeler = new BpmnJS({
[...]

PR #1236, opened 2021-04-06
add :class.class-name=... :style.style-name=... supported

Funlang:
:class.class-name=...
:style.style-name=...

### DIFF
--- a/dist/alpine-ie11.js
+++ b/dist/alpine-ie11.js
@@ -6651,6 +6651,17 @@
         el.value = value;
       }
     } else if (attrName === 'class') {
+      // @feature :class.class-name=...
+      if (modifiers && modifiers.length > 0) {
+        var v = !!value;
+        value = {};
+        modifiers.forEach(function (className) {
+          _newArrowCheck(this, _this);
+
+          return value[className] = v;
+        }.bind(this));
+      }
+
       if (Array.isArray(value)) {
         var originalClasses = el.__x_original_classes || [];
         el.setAttribute('class', arrayUnique(originalClasses.concat(value)).join(' '));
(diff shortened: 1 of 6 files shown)


Funlang:
for performance

Funlang:
for performance

aidenybai:
How is this more performant? Is this feature necessary? Why did you feel the need to add `// added by zwd@funlang.org @2021-04-08 for performance` for each of your commits?

Funlang:
- performance for this issue, https://github.com/alpinejs/alpine/issues/566
- comments in code will be deleted
- :class.xxx.xxx will be supported for modify a little code: modifiers.forEach(....

KevinBatdorf:
Just a tip (I know you're still working on the PR), if you're making performance improvements you'll want to show off a measurement of the improvement with a way for others to interact with.

Funlang:
In this test, https://codepen.io/funlang/pen/BapwWpP, it will be completed in 3s improved from 10s+.

- The mutations have 20,000+ childList, so the loop is so slow, changed to query x-data from the root run at once, it improve immediatelly.

Funlang:
In this benchmark https://krausest.github.io/js-framework-benchmark/2021/table_chrome_89.0.4389.72.html

- AlpineJs' score is not good, so I want to do some improvement

HugoDF:
@Funlang would probably be a good idea to do each perf improvement as a separate PR

Funlang:
> @Funlang would probably be a good idea to do each perf improvement as a separate PR

Thanks that is a good idea, but I don't know how to do that.

HugoDF:
> > @Funlang would probably be a good idea to do each perf improvement as a separate PR
> 
> Thanks that is a good idea, but I don't know how to do that.

You'll need a branch for each perf improvement, then you can open a PR from each of the branches into Alpine

Funlang:
> > > @Funlang would probably be a good idea to do each perf improvement as a separate PR
> > 
> > 
> > Thanks that is a good idea, but I don't know how to do that.
> 
> You'll need a branch for each perf improvement, then you can open a PR from each of the branches into Alpine

thanks, I will study to do that

Funlang:
> > > @Funlang would probably be a good idea to do each perf improvement as a separate PR
> > 
> > 
> > Thanks that is a good idea, but I don't know how to do that.
> 
> You'll need a branch for each perf improvement, then you can open a PR from each of the branches into Alpine

Thanks so much for your help, I think I learned to work on github ;-)

PR #1244, opened 2021-04-08
Performance improvements for lots of dom

Funlang:
performance improvements for a lots of dom


Funlang:
transfered from https://github.com/alpinejs/alpine/pull/1236

calebporzio:
At first glance it seems like this has completely different behavior for when there are more than 10 DOM mutations.

Can you explain why this wouldn't cause any breakages if mutations excede 10?

PR #1265, opened 2021-04-19
Add Failing Test for Duplicate Event Handlers

chriischambers:
This PR adds a failing test to illustrate the issue with nested nodes and their event handlers being duplicated after moving a node in the document.

It would appear that the previous solution tracks in the `listenForNewElementsToInitialize` method a  `__x_node_add_count` count but this is only being tracked on the parent node.  Moving this tracking to the `walkAndSkipNestedComponents` loop and storing it against the correct node does indeed solve the issue and make my test pass but then this causes several other tests in the suite to fail.  At this point I have ran out of time to delve deeper into this.


calebporzio:
This has been fixed in V3. Thanks!

PR #1270, opened 2021-04-21
Fix x-model not selecting falsy values in select dropdown

### LINKED ISSUE #1266: Unable to x-model a disabled option within a select element
I often use disabled options in select elements as placeholders, but I can't get that to work with Alpine. I'm not sure if this is a bug or intentional.

### Example
https://jsbin.com/lezusiyare/edit?html,css,output

### Expected behaviour
The disable option (Choose an option) should be selected

### Current behaviour
The first option (Option one) is selected

### Notes
I've included an example without Alpine that demonstrates the browser native behaviour that I'd expect to be replicated in Alpine.

SimoTod:
Fixes #1266
As per specs (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select)
`Each <option> element should have a value attribute containing the data value to submit to the server when that option is selected. If no value attribute is included, the value defaults to the text contained inside the element.`

This means that options already return text when value is not defined so we don't need that `||`.

Note that `||` breaks any cases where the option value is an empty string since it resolves to false so the code incorrectly uses the option text. 

This issue wasn't noticed earlier because the empty value is usually the first option so the browser just displays it.

Bug: https://codepen.io/SimoTod/pen/rNjqXQx
Fix: https://codepen.io/SimoTod/pen/xxgyvQg

### DIFF
--- a/dist/alpine-ie11.js
+++ b/dist/alpine-ie11.js
@@ -6715,7 +6715,7 @@
     Array.from(el.options).forEach(function (option) {
       _newArrowCheck(this, _this3);
 
-      option.selected = arrayWrappedValue.includes(option.value || option.text);
+      option.selected = arrayWrappedValue.includes(option.value);
     }.bind(this));
   }
 
--- a/dist/alpine.js
+++ b/dist/alpine.js
@@ -752,7 +752,7 @@
       return value + '';
     });
     Array.from(el.options).forEach(option => {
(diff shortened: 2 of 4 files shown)


ryangjchandler (approved):

calebporzio:
Perfect, thanks @SimoTod !

PR #1276, opened 2021-04-24
Update rollup to fix const statement position

stefanr:
This PR simply updates rollup to the latest version which fixes the const statement position for `const initExpression = this.$el.getAttribute('x-init');` which is currently somehow pulled to the top of the beginning of the constructor function.

Now the `x-init` property can be properly set in a `onBeforeComponentInitializeds` callback.


calebporzio:
Thanks!

PR #1310, opened 2021-05-11
Make the tailwindcss logo go to their website instead of the image

quick007:


calebporzio:
Thanks, that makes a lot of sense.

PR #1429, opened 2021-06-11
Fix model docs

jonnitto:
- [x] Fix typo in code example `Multiple select`
- [x] Add missing example of `Dynamically populated Select Options`


calebporzio:
Thanks!

PR #1436, opened 2021-06-11
Fixing some typos

gjcourtney:


packages/docs/src/en/advanced/extending.md
```diff
@@ -227,7 +227,7 @@ Now if the directive is removed from this element or the element is removed itse
 <a name="custom-magics"></a>
 ## Custom magics
 
-Alpine allows you to register custom "magics" (properties or methods) using `Alpine.magic()`. Any magic you register will be available to all your application's Alpine code with the `$` prefix.
+Alpine allows you to register custom "magic" (properties or methods) using `Alpine.magic()`. Any magic you register will be available to all your application's Alpine code with the `$` prefix.
```
markjaquith, line 230: “magics” seems intentional here. One magic, many magics.
IMG_0200

markjaquith (changes requested):
All of these changes look good except for the “magics” change — that one seems like it should stay as-is.

PR #1446, opened 2021-06-12
Removes unnecessary el argument in evaluate docs

stephenoldham:
Little PR to remove the `el` arg in the `evaluate` docs as the element is already bound as the first arg in this instance of the evaluate method.


calebporzio:
Thanks!

PR #1451, opened 2021-06-12
Update show.md

mabouguerra:
Added correct link to transition directive page.

### DIFF
--- a/packages/docs/src/en/directives/show.md
+++ b/packages/docs/src/en/directives/show.md
@@ -26,7 +26,7 @@ When the "Toggle Dropdown" button is clicked, the dropdown will show and hide ac
 <a name="with-transitions"></a>
 ## With transitions
 
-If you want to apply smooth transitions to the `x-show` behavior, you can use it in conjunction with `x-transition`. You can learn more about that directive [here](https://alpinejs.dev/directives/transition), but here's a quick example of the same component as above, just with transitions applied.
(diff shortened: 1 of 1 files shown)


packages/docs/src/en/directives/show.md
```diff
@@ -26,7 +26,7 @@ When the "Toggle Dropdown" button is clicked, the dropdown will show and hide ac
 <a name="with-transitions"></a>
 ## With transitions
 
-If you want to apply smooth transitions to the `x-show` behavior, you can use it in conjunction with `x-transition`. You can learn more about that directive [here](), but here's a quick example of the same component as above, just with transitions applied.
+If you want to apply smooth transitions to the `x-show` behavior, you can use it in conjunction with `x-transition`. You can learn more about that directive [here](https://alpinejs.dev/directives/transition), but here's a quick example of the same component as above, just with transitions applied.
```
HugoDF, line 29: Don't the docs use paths instead of full URLs?
```suggestion
If you want to apply smooth transitions to the `x-show` behavior, you can use it in conjunction with `x-transition`. You can learn more about that directive [here](/directives/transition), but here's a quick example of the same component as above, just with transitions applied.
```

mabouguerra: Probably. What I wanted to change actually was the link on this page https://alpinejs.dev/directives/show which leads to the same page instead of https://alpinejs.dev/directives/transition. Maybe this is not the right place to do so.

calebporzio:
Thanks!

PR #1477, opened 2021-06-14
feature: $root magic variable

ryangjchandler:
This pull request adds a new `$root` magic variable.

If this isn't wanted in core, I can extract into a package under the collective org but since `$el` no longer returns the root element, this could be quite useful to a lot of people (me included).

The userland workaround is using `x-ref="someRef"` and then `$refs.someRef` but this is quite tedious, especially when upgrading from 2.x to 3.x. Adding a `$root` variable essentially makes the upgrade path a find & replace from `$el` to `$root`.


calebporzio:
Yeah, I flip flopped on $root quite a bit during development. I think I'd like to leave it out as it's purpose is more for a smooth upgrade path, than actual development in my experience.

I think I'll add something like this to the docs for folks who want it:
```html
<script>
    document.addEventListener('alpine:initializing', () => {
        Alpine.magic('root', el => {
            let closestRootEl = (node) => {
                if (node.hasAttribute('x-data')) return node

                return closestRootEl(node.parentNode)
            }

            return closestRootEl(el)
        })
    })
</script>
```

ryangjchandler:
@calebporzio No worries - we'll add it to the toolkit as well so users won't have to implement it themselves on each project (bit of a pain point).

PR #1494, opened 2021-06-15
fix wrong link

hendrasob:


ryangjchandler:
I don't think this change is right. The links being used there are for the docs site, not the MD file paths.

E.g. on this page https://alpinejs.dev/alpine-101, the "Read more about `x-data`" is pointing the correct URL for the docs site.

hendrasob:
Yes, I know that.  However, I think it's better to use it because we are using git.  Or better yet link the original link from the doc site, example `https://alpinejs.dev/directives/data` instead of `/directives/data`

ryangjchandler:
I don't think so - the docs aren't designed for reading inside of GitHub, you should be using the website directly.

calebporzio:
Yeah, it's fine the way it is IMO - thanks thought!

hendrasob:
Ohh okay thanks for review

calebporzio:
Yep!

PR #1498, opened 2021-06-15
fix: binding undefined nested value

ryangjchandler:
This pull request fixes an issue where non-existent / undefined nested properties in an object would bind the literal string `"undefined"` instead of an empty string, which was the old behaviour in 2.x.

This behaviour was already implemented as part of the `x-model` binding logic, but was missing from direct `x-bind:` expressions.

### DIFF
--- a/packages/alpinejs/src/directives/x-bind.js
+++ b/packages/alpinejs/src/directives/x-bind.js
@@ -13,6 +13,9 @@ directive('bind', (el, { value, modifiers, expression, original }, { effect }) =
     let evaluate = evaluateLater(el, expression)
 
     effect(() => evaluate(result => {
+        // If nested object key is undefined, set the default value to empty string.
+        if (result === undefined && expression.match(/\./)) result = ''
+
         mutateDom(() => bind(el, value, result, modifiers))
     }))
 })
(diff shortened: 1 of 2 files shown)


packages/alpinejs/src/directives/x-bind.js
```diff
@@ -13,6 +13,9 @@ directive('bind', (el, { value, modifiers, expression, original }, { effect }) =
     let evaluate = evaluateLater(el, expression)
 
     effect(() => evaluate(result => {
+        // If nested object key is undefined, set the default value to empty string.
+        if (result === undefined && expression.match(/\./)) result = ''
```
calebporzio, line 17: the only thing that's curious about this is the .match

Consider if you are binding to a nested property that doesn't exist, but the object is referenced from somewhere else, like a function or something - then it wouldn't get picked up here.

Do you remember why we had the expression.match in the first place? Are there times you want undefined to actually result in "undefined"?

ryangjchandler: I don't remember the reasoning behind the `.match` call tbh.. it was probably because we only wanted to support this behaviour for objects, not any non-existent piece of data (which would fail now anyway).

ryangjchandler: I think leaving the `match` in there is fine as that's the same behaviour that `x-model` uses.

calebporzio:
Thanks!

PR #1536, opened 2021-06-17
Init directives for x-if contents after it's created.

Niklan:
This is a change related to this bug: https://github.com/alpinejs/alpine/discussions/1535

I'm not sure if I fixed it correctly, but it solves the problem.

* **Alpine v2:** https://codepen.io/Niklan/pen/yLMZRoM
* **Alpine v3 before:** https://codepen.io/Niklan/pen/XWMOxKa
* **Alpine v3 after:** https://codepen.io/Niklan/pen/vYxbQZL

### DIFF
--- a/packages/alpinejs/src/directives/x-if.js
+++ b/packages/alpinejs/src/directives/x-if.js
@@ -1,7 +1,8 @@
 import { evaluateLater } from '../evaluator'
-import { setStyles } from '../utils/styles'
 import { directive } from '../directives'
-import { once } from '../utils/once'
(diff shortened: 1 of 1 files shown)


Niklan:
I'm also found that in some cases (can't reproduce it in codepen, and my example is too big) my fix with `initTree()` causes duplicates for `x-for`. I've added proper DOM mutation as it done in `x-for` and it also helps.

calebporzio:
I think this is a good addition. Can you add tests for this?

PR #1556, opened 2021-06-18
Fix x-bind:class directive using object syntax removes classes that are in a true condition when the class is also present in a false condition

JacobPozaic:
A discussion about this issue can be found here: https://github.com/alpinejs/alpine/discussions/1525

Description:
Remove classes under false condition before adding classes under true condition to prevent classes from a true condition being removed if a false condition has the same class.

Changes:
- Made a suggested change from @KevinBatdorf
- added unit test


calebporzio:
This is perfect, THANK YOU!

PR #1560, opened 2021-06-18
x-for referenceError undefined when spaces are present in parenthesis

JacobPozaic:
A discussion about this issue can be found here: https://github.com/alpinejs/alpine/discussions/1520

Description:
Fix an issue with x-for where parsing the item name out of parenthesis would cause an error if spaces where present around the item name and index was not used.

Example:
-> This fails:
    ```<template x-for="( item ) in items">```
-> This succeeds:
    ```<template x-for="(item) in items">```
    
Changes:
- Fixed an issue with trimming whitespace around the item name in x-for
- Added unit test


calebporzio:
Thanks!

PR #1569, opened 2021-06-20
Cleanup nested removed element's directives

calebporzio:


mokhosh:
@calebporzio can you please consider reverting this or optimizing it? It's forcing me to go back to v2.
https://github.com/alpinejs/alpine/discussions/2329

PR #1571, opened 2021-06-20
Fix playground sample page

MuzafferDede:
Corrected playground page's missing parts to make it work out of the box.


calebporzio:
Thanks - I like adding a sample component (I'll probably go back to no head or body) - can you remove the package lock file from this PR?

MuzafferDede:
Alright, removed head and body as well. 
FYI, I still have to change source paths of the scripts because they do not work if you run index.html without a server. I think you are using valet for server environment which might not be the case for everyone.

calebporzio:
Ok great, THANKS!

PR #1580, opened 2021-06-21
feature: support args with Alpine.data()

ryangjchandler:
This pull request adds support for passing arguments through to `Alpine.data` providers.

Without too much thought, the simplest approach is the one that I took. 

Everything before the first `(` character in the expression will be used to get the data provider. If that exists and there is some expression after the first `(`, that will be evaluated inside of an array, allowing us to spread the data into the function call.

i.e. `x-data="foo('bar')"`, in this scenario `foo` will be used as the data-provider and the `'bar'` will actually be evaluated as `['bar']` so that we can then spread the result of the evaluation into the data provider's function.

Here's a few bullet points of the flow:

1. Get everything before the first occurence of `(` in `expression`. If it doesn't return anything (meaning `(` isn't present), then default to `expression`.
2. If `dataProviderName` is not empty and the `dataProviderName` is not equal to the original `expression` then we must have some arguments or at least `()` present in the `expression`.
3. Get the raw argument string by taking everything after the first `(` and then removing the last character, which should be a `)` char.
4. Evaluate this string inside of an array, i.e. `[${stringHere}]`. This will return an array into `args` which we can then `...` spread into the data provider callback.

**Question**: is there anything that needs to be done for CSP-safety here?

### DIFF
--- a/packages/alpinejs/src/directives/x-data.js
+++ b/packages/alpinejs/src/directives/x-data.js
@@ -13,14 +13,31 @@ addRootSelector(() => `[${prefix('data')}]`)
 directive('data', skipDuringClone((el, { expression }, { cleanup }) => {
     expression = expression === '' ? '{}' : expression
 
-    let dataProvider = getNamedDataProvider(expression)
+    // This returns everything up until the first `(` in the expression,
+    // returning the name of the data provider in scenarios where arguments
+    // are being passed to the callback.
+    const dataProviderName = expression.substr(0, expression.indexOf('(')) || expression
+
+    let dataProvider = getNamedDataProvider(dataProviderName)
 
     let data = {}
 
     if (dataProvider) {
         let magics = injectMagics({}, el)
+        let args = [];
 
-        data = dataProvider.bind(magics)()
+        // This adds support for passing arguments to `Alpine.data()`
+        // callbacks by evaluating the parts inside of `()` as an `[]`
+        // expression and then spreading that out to the data provider.
+        if (dataProviderName !== '' && expression !== dataProviderName) {
+            // Get everything after the first `(` in expression and remove the final character,
+            // which should be the closing `)`.
+            const argsString = expression.substring(expression.indexOf('(') + 1).slice(0, -1)
+
+            args = evaluate(el, `[${argsString}]`)
+        }
+
+        data = dataProvider.bind(magics)(...args)
     } else {
         data = evaluate(el, expression)
     }
--- a/tests/cypress/integration/custom-data.spec.js
+++ b/tests/cypress/integration/custom-data.spec.js
@@ -17,6 +17,23 @@ test('can register custom data providers',
     ({ get }) => get('span').should(haveText('bar'))
 )
 
+test('can pass arguments to data provider',
+    html`
+        <script>
+            document.addEventListener('alpine:init', () => {
+                Alpine.data('test', (value) => ({
+                    foo: value
+                }))
+            })
+        </script>
+
+        <div x-data="test('bar')">
+            <span x-text="foo"></span>
+        </div>
+    `,
+    ({ get }) => get('span').should(haveText('bar'))
+)
+
 test('init functions inside custom datas are called automatically',
     html`
         <script>


KevinBatdorf:
For completeness, might need to cover `() => test(bar)`? This might cover more:

```js
"() => test(param)".replace(/\(([^\)]+)\)/, '').replace('()', '').replace('=>', '').trim()
"test"

"test(param)".replace(/\(([^\)]+)\)/, '').replace('()', '').replace('=>', '').trim()
"test"

"test()".replace(/\(([^\)]+)\)/, '').replace('()', '').replace('=>', '').trim()
"test"

"test".replace(/\(([^\)]+)\)/, '').replace('()', '').replace('=>', '').trim()
"test"
```

ryangjchandler:
@KevinBatdorf I don't think we need to cover that do we? This is purely for `Alpine.data()` so we only need to support `test(param)` or `test`.

KevinBatdorf:
Might not have to, but people may expect it to work, and it works in this currently:
```js
<div x-data="() => test('why')">
```

ryangjchandler:
It's not a supported feature at the moment though, so I don't think we need to do anything.

KevinBatdorf:
The `replace` with regex seems pretty slow too compared to substring, so probably better not to support it

ryangjchandler:
Yeah, we'll see what @calebporzio says because he might have a better idea of the expected outcome.

KevinBatdorf:
I've also seen multiple people expecting it to work like that (and some not realizing they can inline it ie. `x-data="test(param)"` like in v2).

Maybe for CSP just add a note in the readme to not use that parameter if they need compliance (the dev might not realize it)?

calebporzio:
Interesting, I'm camping until Thursday, so I'll take a deeper look when I'm back.

I wonder if the best thing though is to make the data providers available as functions in the x data scope. So no parsing is needed and it's more flexible. We will have to do something for csp, but thats fine

ryangjchandler:
> Interesting, I'm camping until Thursday, so I'll take a deeper look when I'm back.
> 
> I wonder if the best thing though is to make the data providers available as functions in the x data scope. So no parsing is needed and it's more flexible. We will have to do something for csp, but thats fine

Yeah, I had thought about taking this approach as well. Effectively taking `datas` and making those functions available as pseudo-named functions inside of the `evaluate()` scope using the `extras` param.

I can take a look at this too so you can take a look when you're back from camping (enjoy btw, well deserved!).

allmarkedup:
Just a +1 for this - I've run into this a number of times already in playing with v3 (which is really lovely, btw!) - probably because I was expecting `Alpine.data` to be a drop-in replacement for the 'old style' of defining a function on the window (which was the impression I got from @calebporzio's "Future of Alpine' video).

FWIW, I'm also missing the ability to compose a single `x-data` object from multiple sources, which was possible using functions:

```html
<div x-data="{...someBehaviour(), ...someOtherBehaviour()}"></div>
```

but is not possible with `Alpine.data`, as far as I can tell.

Obviously the old way still works, but if `Alpine.data` is being generally suggested as a nicer API for `x-data` sharing the composable aspect would be great to get back somehow.

ryangjchandler:
I'm away on holiday this weekend but I can revisit our earlier discussion next week. The "bind data providers to the execution context approach" should work well enough.

danddanddand:
> FWIW, I'm also missing the ability to compose a single `x-data` object from multiple sources, which was possible using functions:
> 
> ```
> <div x-data="{...someBehaviour(), ...someOtherBehaviour()}"></div>
> ```
> 
> but is not possible with `Alpine.data`, as far as I can tell.

I don't use Alpine.data, but for the most part wouldn't
```
<div x-data="someBehaviour">
    <div x-data="someOtherBehaviour">
    ... 
    </div>
</div>
```
be pretty much same result? 

It's not super elegant and obviously it's creating two component roots which might not be ideal, but wouldn't this work as replacement if you were set on using Alpine.data with current API.  

We've always done spread the behaviour like your example which has worked well and one of the reasons we're sticking with functions in our code base. Stuff like this for example, 
```
import { getViewportWidth, screens } from "../utils/screens/screenUtils";

export interface BaseScreenProps extends BaseAlpineProps {
    isSm: boolean;
    isMd: boolean;
    isLg: boolean;
    isXl: boolean;
    resizeHandler: () => void;
    initScreens: () => void;
}

export const baseScreens = (): BaseScreenProps => ( {
    isSm: false,
    isMd: false,
    isLg: false,
    isXl: false,
    resizeHandler: function () {
        this.isSm = getViewportWidth() >= screens.sm;
        this.isMd = getViewportWidth() >= screens.md;
        this.isLg = getViewportWidth() >= screens.lg;
        this.isXl = getViewportWidth() >= screens.xl;
    },
    initScreens: function () {
        this.resizeHandler();
    },
} );
```
is spread into 10 components when I find usages in our codebase:
Screen Shot 2021-06-24 at 9 47 39 PM.

In V3 I suspect this could just be a 
<div x-data="baseScreens()">
  <div x-data="somethingUsingScreens()">
  </div>
</div>

Again, I don't love it, but I think it would be "a" way to do it since children components have access to parent.

PR #1637, opened 2021-06-25
x-transition in x-bind if you use functions

shiro1702:
x-transition in x-bind if you use functions don't works

https://github.com/alpinejs/alpine/discussions/1633

https://github.com/alpinejs/alpine/discussions/1458


calebporzio:
Please provide a test with this PR - Thanks!

PR #1643, opened 2021-06-26
Add warning for use of standalone x-init directive

arcanemachine:
- Warn that using standalone x-init creates a new scope and any $refs
  created outside this scope will not be accessible inside the new scope


HugoDF:
Is there a way to fix this instead maybe?

calebporzio:
Mmm, yeah it makes more sense to stack refs like data right?

calebporzio:
Thanks for this. I've implemented the fix so this warning is no longer necessary: https://github.com/alpinejs/alpine/pull/1661

PR #1645, opened 2021-06-26
RFC: Unscope x-init when nested in another component

KevinBatdorf:
It's not the cleanest solution since `rootSelectors()` is called once during boot up (where it applies its own filtering, see `outNestedComponents`), then later whenever using `closestRoot()`. During the latter, `rootSelectors` has more context available to it that the `addRootSelector` callback can then use (passing in `el`).

So essentially, the type signature changes depending on context. I set the default to `undefined` on `rootSelectors` as a hint to anyone later who uses it that el could be undefined (and that they should check for it as I did in `x-init`).

fixes #1643 

I thought about separating `rootSelectors()` into two things, like `rootSelectors()` and `rootSelectorsMaybe()` but that felt equally quirky. 

Thoughts anyone?

### DIFF
--- a/packages/alpinejs/src/directives/x-init.js
+++ b/packages/alpinejs/src/directives/x-init.js
@@ -1,8 +1,13 @@
+import { addRootSelector, closestRoot } from "../lifecycle";
 import { directive, prefix } from "../directives";
-import { addRootSelector } from "../lifecycle";
 import { skipDuringClone } from "../clone";
 import { evaluate } from "../evaluator";
 
-addRootSelector(() => `[${prefix('init')}]`)
+const isNestedInsideAComponent = (el) =>
+    el && el.hasAttribute(prefix('init'))
+    && closestRoot(el.parentNode || closestRoot(el))
+
+// If an element has init, but is a child of data then do not add it as a root selector
+addRootSelector((el) => isNestedInsideAComponent(el) ? null : `[${prefix('init')}]`)
 
 directive('init', skipDuringClone((el, { expression }) => evaluate(el, expression, {}, false)))
--- a/packages/alpinejs/src/lifecycle.js
+++ b/packages/alpinejs/src/lifecycle.js
@@ -33,14 +33,14 @@ export function start() {
 
 let rootSelectorCallbacks = []
 
-export function rootSelectors() {
-    return rootSelectorCallbacks.map(fn => fn())
+export function rootSelectors(el = undefined) {
+    return rootSelectorCallbacks.map(fn => fn(el))
 }
 
 export function addRootSelector(selectorCallback) { rootSelectorCallbacks.push(selectorCallback) }
 
 export function closestRoot(el) {
-    if (rootSelectors().some(selector => el.matches(selector))) return el
+    if (rootSelectors(el).some(selector => el.matches(selector))) return el
 
     if (! el.parentElement) return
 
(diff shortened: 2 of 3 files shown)


SimoTod:
Wouldn't be simpler to allow child component to access the refs of the parent? This should be relatively simple to implement in x-ref. I usually don't like when a single instruction has different behaviours depending on the context because bugs are harder to debug. Just my personal opinion though

danddanddand:
I pulled your branch to play with it and for some added this test but it doesn't pass and I don't get it? Quark with way test suite is implemented or issue with x-text or issue with implementation?

```
test('x-init does not create scope',
    [html`
        <div x-init="$store.test.updateFoo()">
            <span x-text="$store.test.foo">initial</span>
        </div>
    `,
        `
        Alpine.store('test', { foo: 'bar', updateFoo() { this.foo = 'blah' } })
    `],
    ({ get }) => {
        get('span').should(haveText('inital'))
    }
)
```

KevinBatdorf:
> I usually don't like when a single instruction has different behaviours depending on the context because bugs are harder to debug.. Just my personal opinion though

I agree, but currently it's already not a pure function, as `x-init` doesn't run through the init process (it's removed because of no parent root actually), but later when booted will act as a root because it remains in the array in those cases.

 It's added to the roots array when it shouldn't be. So this solves that (indirectly). Having it in the array could cause issues later when using `closestRoot()` as it will return `true` for those cases, I think (probably worth adding a test for that too actually)

> I pulled your branch to play with it and for some added this test but it doesn't pass and I don't get it?
I didn't test it with `$store` I'll take a look how that rune.

I'll also check if it's easier going through `$refs`

KevinBatdorf:
For example:

```html
<div x-data>
    <div x-init="console.log()">
      <button @click="console.log(Alpine.closestRoot($el))">press</button>    
    </div>
</div>
```

This will return the `x-init` element instead of `x-data`. I'll push this test now

KevinBatdorf:
```html
<div x-init="$store.test.updateFoo()">
  <span x-text="$store.test.foo">initial</span>
</div>
```

This actually feel like an anti-pattern to me. I would expect it to fail so if it's working now I would consider it a bug. 

I see standalone `x-init` being used to run one off commands like "set this datepicker up" or "log this value when inserted" but not to do everything a component does. ie. the `x-text` needs to be in `x-data`

KevinBatdorf:
Or maybe I misunderstood. You're also suggesting it should break, right? Hmm. I'll check some more

danddanddand:
Ya I added it expecting it to pass on your branch and was very confused why it wasn't since x-text not in component should be ignored

KevinBatdorf:
I think this might be intended actually. The function name is `addRootSelector()` so maybe it's intended to act like a root element when it's _not_ nested inside other components. 

---

But then maybe it should also always be scoped, and just have access to `$refs` like @SimoTod said. 

I guess the question is, should `x-init` or more specifically, anything registered with `addRootSelector` have new scope and behave as `x-data` does? Including return for `closestRoot()`? Regardless of being nested or not?

Should there be a separate `addTopLevelOnlyRootSelector()`?


I'll leave this PR as is and see what Caleb thinks.

I'll check out just adding access to `$refs` in a separate PR when I have some time.

calebporzio:
Thanks for all the input on this. I've implemented it and merged it here: https://github.com/alpinejs/alpine/pull/1661

PR #1655, opened 2021-06-27
pass in proper param to querySelectorAll

danddanddand:
Very small thing but rootSelectors() returns string array (which works but is technically invalid) so add comma for proper usage (https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelectorAll#parameters)

- only matters because of d.ts file I'm building for work (maybe future PR)
- covered under existing test coverage (will add test if requested)

### DIFF
--- a/packages/alpinejs/src/lifecycle.js
+++ b/packages/alpinejs/src/lifecycle.js
@@ -22,7 +22,7 @@ export function start() {
 
     let outNestedComponents = el => ! closestRoot(el.parentNode || closestRoot(el))
 
-    Array.from(document.querySelectorAll(rootSelectors()))
+    Array.from(document.querySelectorAll(rootSelectors().join(",")))
         .filter(outNestedComponents)
         .forEach(el => {
             initTree(el)


KevinBatdorf:
The signature is actually to provide a DOMString, which is auto cast to a string [according to the spec](https://heycam.github.io/webidl/#es-DOMString). 

It's probably worth seeing which one is faster and let that decide.

Is there a way to make TS ok with that I wonder?

danddanddand:
I can @tsignore that line but Intellij and ts were both complaining and Intellij complains even with no TS in project (fresh clone of main) just from it's internal definitions which in my experience are extremely accurate.  

> Screen Shot 2021-06-27 at 6 22 11 PM

I suspect JS engine is probably doing conversion under the hood anyway so I'd be surprised if there are performance consequences.

KevinBatdorf:
It might be faster if the engine is casting it, but it's probably not a noticeable difference regardless. 

Typically in the past it's been more of a "unless there's significant improvement in speed, if it's not broke, don't fix it". 
but if n% of contributors have to deal with annoying messages that might be enough to merge it in. I guess just see what Caleb thinks. Adding `join()` may make more sense for clarity anyway.

danddanddand:
Once it's in TypeScript it's higher severity error
`Screen Shot 2021-06-27 at 6 33 53 PM`

Honestly, because project uses esbuild and it supports ts out of the box, it's incredible how few issues like this there are. It shows how great of a JS programmer Caleb is. I know there isn't much code in alpine package, but if you convert all the files from .js to .ts and npm run build it compiles fine, all the tests pass. Once you turn on noImplicitAny and started typing everything small stuff like this are probably better to just fix

danddanddand:
All the tests pass and ts shuts up with the explicit String( ) cast as well, so either is fine with me if you or Caleb prefer that

danddanddand:
For what it's worth, join is faster but no human would ever notice it.

```
test( "speed test 2", () => {
    const startTime = new Date().getTime();
    for( let i = 0; i < 5000000; i++ ) {
        ['blah', 'blah2'].join( ',' );
    }
    const endTime = new Date().getTime();
    console.log( endTime - startTime );
} );

test( "speed test", () => {
    const startTime = new Date().getTime();
    for( let i = 0; i < 5000000; i++ ) {
        String( ['blah', 'blah2'] );
    }
    const endTime = new Date().getTime();
    console.log( endTime - startTime );
} );
```

> Screen Shot 2021-06-27 at 7 16 24 PM

KevinBatdorf:
Yeah so I guess if there's a change, then it should be with `join()`. The engine casting would likely be faster though (then again, I don't know)

I personally don't have a preference. Also, I'm just here the same as you so nothing I say holds any more or less weight than anyone else.

danddanddand:
You made me curious so here's codepen demonstrating there is no performance difference (takes ~45s to run on M1 in Chrome)
https://codepen.io/danddanddand/pen/XWRrQvB

calebporzio:
@danddanddand - sorry, I don't consider the "rootSelectors" method public API (don't want to be locked into not making breaking changes on a method like that)

Also, I used to join, but switched to array at some point - the MDN docs don't explicitly say it's invalid.

Going to close this - thanks for the contribution though!

danddanddand:
Ok, no problem, I felt dumb making a PR from something so trivial and understand your perspective

KevinBatdorf:
>  I felt dumb making a PR from something so trivial 

I can't speak for Caleb, of course, but I personally don't think you should feel this way at all. Not in the slightest.

PR #1660, opened 2021-06-28
Allow passing runtime parameters to Alpine.data() objects

calebporzio:
This is the current story for using Alpine.data objects in Alpine:
```html
<div x-data="test">...</div>
```
```js
Alpine.data('test', () => {
    return {
        foo: 'baz',
    }
})
```

In addition to this syntax, this PR would allow you to call that provider as a function and pass parameters in like so:
```html
<div x-data="test('baz')">...</div>
```
```js
Alpine.data('test', (arg) => {
    return {
        foo: arg,
    }
})
```

Because `test` is made available as a plain function, you can spread multiple together to compose objects like was popular in Alpine V2:

```html
<div x-data="{ ...test('baz'), ...test2('bob') }">...</div>
```


ryangjchandler:
Lol, was about to open a PR for this.

PR #1665, opened 2021-06-28
fix error where data was being hoisted to global scope introduced in 3.2.0

JacobPozaic:
Fix related to: https://github.com/alpinejs/alpine/discussions/1664


calebporzio:
Thanks for this!

PR #1676, opened 2021-06-29
Fix function syntax in dropdown example

fitimvata:


calebporzio:
This has already been fixed in a previous PR - thanks though!

PR #1679, opened 2021-06-29
feature: support registering multiple stores at once

ryangjchandler:
This pull request adds support for registering multiple stores at once via an object. Thought this was worth a shot at getting merged. 

Here's how it will affect your code:

**Before**:
```js
Alpine.store('settings', {
    darkMode: false,
})

Alpine.store('modals', {
    active: null,
})
```

**After**
```js
Alpine.store({
    settings: {
        darkMode: false,
    },
    modals: {
        active: null,
    }
})
```

This is somewhat similar to ["state modules"](https://vuex.vuejs.org/guide/modules.html) in VueX, where you register each module inside of a single object and provide a key. 

> If this has merge potential, I'll also update the docs.

### DIFF
--- a/packages/alpinejs/src/store.js
+++ b/packages/alpinejs/src/store.js
@@ -6,6 +6,14 @@ let isReactive = false
 export function store(name, value) {
     if (! isReactive) { stores = reactive(stores); isReactive = true; }
 
+    if (typeof name === 'object' && value === undefined) {
+        Object.entries(name).forEach(([key, value]) => {
+            store(key, value)
+        })
+
+        return
+    }
+
     if (value === undefined) {
         return stores[name]
     }
--- a/tests/cypress/integration/store.spec.js
+++ b/tests/cypress/integration/store.spec.js
@@ -17,6 +17,29 @@ test('can register and use a global store',
     }
 )
 
+test('can register multiple stores at once',
+    [html`
+        <div x-data>
+            <span id="foo" x-text="$store.foo.bar"></span>
+            <span id="boo" x-text="$store.boo.car"></span>
+        </div>
+    `,
+    `
+        Alpine.store({
+            foo: {
+                bar: 'baz'
+            },
+            boo: {
+                car: 'caz'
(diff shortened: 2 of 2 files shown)


calebporzio:
I see. Definitely not an intimidating implementation, I just don't want to expand the API unless there is a significant demand for this type of thing. Is there proof of a high demand along with a tangible problem this solves?

ryangjchandler:
> I see. Definitely not an intimidating implementation, I just don't want to expand the API unless there is a significant demand for this type of thing. Is there proof of a high demand along with a tangible problem this solves?

I've seen a couple of people ask about it (somebody asked today on the Discord) & somebody at my work actually tried it so I thought I'd try to get a PR in.

carlmjohnson:
If the stores are related, they could both be on the same top level store. If they're not related, I don't think it's good to combine their declarations.

ryangjchandler:
> If the stores are related, they could both be on the same top level store. If they're not related, I don't think it's good to combine their declarations. Eg.
> 
> ```js
> Alpine.store("mysite", {
>     settings: {
>         darkMode: false,
>     },
>     modals: {
>         active: null,
>     }
> })
> ```

Sure, they can be grouped together. It's a personal preference thing of course, grouping them in a single call removes extra calls to `Alpine.store` from your own code.

carlmjohnson:
My position is that it's bad to have "smart" framework functions that poke at the type of the value they get to do different things, unless the poking is a _big_ usability advantage for most users. This doesn't cross that threshold for me, and it adds complications like what happens if you do `Alpine.store(Symbol('x'), ...)`? I guess it should be smart enough to understand that a symbol is like a string, but then what about `Alpine.store(new Map())`? Is a Map like an object? Should it unpack it? How smart should it be? I favor keeping it simple and avoiding all these questions unless it really saves a lot of coding elsewhere.

calebporzio:
Going to close this until there is more expressed demand. Definitely open to it in the future - thanks!

PR #1697, opened 2021-06-30
fix x-html with alpine code as a sibling of xfor or xif

tvdr:
This pull request fixes the bug discovered in #1688 

The problem generally is that when an x-for or an x-if preceds an x-html directive with alpine content in the same scope, then the for/if directives pause observing mutations and as walking through the elements, initialize the x-html.  
As the mutationobserver is paused, the alpine content inserted from the x-html wont "render".

Running the innerHTML assignment in a nextTick seems to solve the problem without (as far as my knowledge goes) raising any othen issues

### DIFF
--- a/packages/docs/src/en/globals/alpine-data.md
+++ b/packages/docs/src/en/globals/alpine-data.md
@@ -48,7 +48,7 @@ Alpine.start()
 This assumes you have a file called `dropdown.js` with the following contents:
 
 ```js
-export default function () => ({
+export default () => ({
     open: false,
 
     toggle() {


tvdr:
Below is an example more in line with the original use case in #1688 
Not sure if these should also be test cases..

```html
<div x-data="dataprovider">

        <button @click="items.push(Math.round(Math.random() * 100))">Add new item</button>

        <h2>Items shown with x-for</h2>
        <template x-for="(item,index) in items">
            <span x-text="item"></span>
            <br>
        </template>

        <hr>

        <h2>Items listed with x-html, inserting spans with alpine code</h2>
        <span x-html="generateHtml">

        </span>

        <hr>

        <h2>Items as input x-models, inserted with x-html</h2>
        <div x-html="generateInputs">

        </div>
</div>

    <script>
        function dataprovider(){
            return {
                items:[1,2,3,4],
                generateHtml(){
                    let htmltext = '';

                    let element_tpl = '<span  x-text="%item%"></span>'
                    this.items.map((item,key)=>{
                        htmltext += element_tpl.replace('%item%','items['+key+']').replace('%key%',key);
                    })

                    return htmltext;
                },
                generateInputs(){
                    let htmltext = '';

                    let element_tpl = '<input x-model="%item%" />'
                    this.items.map((item,key)=>{
                        htmltext += element_tpl.replace('%item%','items['+key+']');
                        htmltext += '<br>';
                    })

                    return htmltext;
                }
            }
        }
    </script>
    <style>
        span{
            margin-right: 5px;
        }
    </style>
```

tvdr:
Sorry had some trouble cleaning up the code and presenting some more examples. I'll try in an other PR soon, with more care

PR #1702, opened 2021-07-01
Add doctype, encoding, and title to HTML example

jpaquim:
The heart emoji in the example was not rendering correctly in the browser, I assumed due to missing UTF-8 encoding on the document. I added the encoding meta tag and the HTML5 doctype for predictable browser rendering, as well as a document title for good measure.

I considered adding a couple more things such as a `lang` attribute, and the usual viewport-related meta tags, but I wanted to first get some feedback on whether it makes sense, or to keep it deliberately minimal.

Keep up with the great work!

### DIFF
--- a/packages/docs/src/en/alpine-101.md
+++ b/packages/docs/src/en/alpine-101.md
@@ -10,8 +10,11 @@ Create a blank HTML file somewhere on you computer with a name like: `i-love-alp
 Using a text editor, fill the file with these contents:
 
 ```html
+<!DOCTYPE html>
 <html>
 <head>
+    <meta charset="UTF-8" />
+    <title>I ❤️ Alpine</title>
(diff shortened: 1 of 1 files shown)


calebporzio:
Thanks @jpaquim - what browser were you using?

Yeah, I'd like to keep things as minimal as possible, so I'd like to avoid this addition even. However, if a common modern browser requires that this exists to render an emoji, we should add it.

Thanks!

jpaquim:
Hey @calebporzio I was testing using Chrome on macOS, so a fairly standard setup. I'm all for keeping it minimal, but this does seems like a minimal setup for compliance with the modern HTML standard, with the `<!DOCTYPE>` minimizing the risk of the browser rendering in quirks mode, and the UTF-8 encoding to correctly render the emoji.

I don't mind removing the title, although it's still considered to be a required element in the document head. What do you think?

PR #1721, opened 2021-07-05
add no-start option for CDN version of Alpine

danddanddand:
Add no-start option to for people who do not want to start Alpine from CDN version automatically. 

Used like `<script src="cdn.js" defer no-start></script>`

Test using: 
```html
<html>
<head>
    <script defer no-start src="cdn.js"></script>
</head>
<body>
<div x-data="{a:1}">
    <span x-text="a"></span>
</div>
</body>
</html>
<script>
    setTimeout(() => {
        Alpine.start()
    }, 5000);
</script>
```

https://github.com/alpinejs/alpine/discussions/1710

https://github.com/alpinejs/alpine/discussions/1705

This strategy is borrowed directly from [petite-vue](https://github.com/vuejs/petite-vue/blob/main/src/index.ts). Just doing inverse logic. 

@ryangjchandler suggested the name `lazy` instead and it's open to discussion what it should be called if added?

### DIFF
--- a/packages/alpinejs/builds/cdn.js
+++ b/packages/alpinejs/builds/cdn.js
@@ -2,6 +2,9 @@ import Alpine from './../src/index'
 
 window.Alpine = Alpine
 
-queueMicrotask(() => {
-    Alpine.start()
-})
+let s
+if (!((s = document.currentScript) && s.hasAttribute('no-start'))) {
+    queueMicrotask(() => {
+        Alpine.start()
+    })
+}


pomartel:
Looks great! And I guess it would also work if the Alpine lib was bundled with other scripts such as this:
```html
<script defer no-start src="/scripts/vendor.js"></script>
```

calebporzio:
@danddanddand - this is interesting.

"lazy" sounds like the best name, but I agree that it implies it will be auto-started when it is first used.

"no-start" seems a little inconsistent. I think "no-init" would be the best name.

This feature still feels a little odd to me being the inverse of what's intuitive.

I'm going to close this PR for now. If there is more demand for such a feature, please reopen this PR - I'm open to it, just want to keep things minimal until necessary. Thanks!

PR #1753, opened 2021-07-08
Update data.md

fgb4101:
Fix the `x-show` variable in the second example in the `Single-element components` section.


calebporzio:
Thanks!

PR #1786, opened 2021-07-13
Bug - Add support for x-data on html tag

SimoTod:
- `closestRoot` uses `el.matches` which requires a DOM Element
- `outNestedComponents` was using `parentNode` that doesn't always return a DOM Element (e.g. html.parentNode. returns document which is a Dom Node). `parentElement` instead will return null if the parent node is. not a DOM Element.
- ⚠️ I had to change `closestRoot(el.parentNode || closestRoot(el))` to `closestRoot(el.parentNode)` but I'm not sure about your original intentions so it might not be suitable. Reason: `closestRoot(el)` always returns the current element if `el` has an `x-data` attribute, so does `closestRoot(closestRoot(el))`. This means that `outNestedComponents` would filter out all the components when `x-data` is on the html tag. All test pass but I thought it was important to highlight this point since there could be a reason why the original code had `|| closestRoot(el)`. Thanks

### DIFF
--- a/packages/alpinejs/src/lifecycle.js
+++ b/packages/alpinejs/src/lifecycle.js
@@ -20,8 +20,7 @@ export function start() {
         directives(el, attrs).forEach(handle => handle())
     })
 
-    let outNestedComponents = el => ! closestRoot(el.parentNode || closestRoot(el))
-
+    let outNestedComponents = el => ! closestRoot(el.parentElement)
     Array.from(document.querySelectorAll(allSelectors()))
         .filter(outNestedComponents)
         .forEach(el => {
@@ -46,6 +45,8 @@ export function addRootSelector(selectorCallback) { rootSelectorCallbacks.push(s
 export function addInitSelector(selectorCallback) { initSelectorCallbacks.push(selectorCallback) }
 
 export function closestRoot(el) {
+    if (!el) return
+
     if (rootSelectors().some(selector => el.matches(selector))) return el
 
     if (! el.parentElement) return
--- a/packages/alpinejs/src/lifecycle.js
+++ b/packages/alpinejs/src/lifecycle.js
@@ -20,7 +20,8 @@ export function start() {
         directives(el, attrs).forEach(handle => handle())
     })
 
-    let outNestedComponents = el => ! closestRoot(el.parentElement)
+    let outNestedComponents = el => ! closestRoot(el.parentNode || closestRoot(el))
+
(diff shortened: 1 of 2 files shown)


danddanddand:
Should the test for this do something to ensure the x-data added is evaluated/used in the assert portion? Something like this makes more sense to me?

```
test('x-data works on the html tag v2',
    [html`
        <div>
            <span x-text="n+2"></span>
        </div>
    `,
        `
        document.querySelector('html').setAttribute('x-data', '{n:3}')
    `],
    ({ get }) => {
        get('span').should(haveText('5'))
    }
)
```

That way x-text is relying on x-data on html and test is using the data to for assertion. Maybe alpine was bombing with x-data on html before and that existing test is good enough. Just my thought when glancing at it.

SimoTod:
> Should the test for this do something to ensure the x-data added is evaluated/used in the assert portion? Something like this makes more sense to me?
> 
> ```
> test('x-data works on the html tag v2',
>     [html`
>         <div>
>             <span x-text="n+2"></span>
>         </div>
>     `,
>         `
>         document.querySelector('html').setAttribute('x-data', '{n:3}')
>     `],
>     ({ get }) => {
>         get('span').should(haveText('5'))
>     }
> )
> ```
> 
> That way x-text is relying on x-data on html and test is using the data to for assertion. Maybe alpine was bombing with x-data on html before and that existing test is good enough. Just my thought when glancing at it.

@danddanddand The issue is not on the evaluation side, we can safely use a string constant to make sure a simple directive is processed (before the fix it would file, see commit https://github.com/SimoTod/alpine/commit/bdedf33289a540e01948e1219af6680e2c4db848).

I usually prefer to keep my tests as simple as possible and in scope with the change introduced but Caleb can expand it later if needed for a different bug.

calebporzio:
Thanks @SimoTod ! At first glance, it looks this test would pass even without the PR right? Because `x-data` is on the <div> as well?

SimoTod:
It would fail because it's a nested component and it doesn't autostart

SimoTod:
I've reverted the code to the broken version without the additional x-data (I hadn't realised I committed it) to show the failing scenario again (https://github.com/alpinejs/alpine/pull/1786/commits/cdc02dc61e2b5eb2274878c0ec7ae0c2bb72ba0a) and then reapplied the fix.

calebporzio:
Thanks!

PR #1793, opened 2021-07-15
Add .nvmrc

markjaquith:
I tried to run the `npm run build` script, and it failed because of the version of Node I was using.

This ensures that people using NVM can run the build script regardless of what their default version of node is. I chose 15.9.0 because it's the latest version of version 15 which is what is used in the GitHub Actions workflow.


calebporzio:
Cool! Didn't know this existed - there's no way to put this in package.json right? I'd rather do that to avoid the extra file if that's possible.

ryangjchandler:
@calebporzio I believe you can do this using the `engines` key in the `package.json` file:

```json
"engines": {
    "node": ">=15.9.0"
}
```

calebporzio:
Oh cool - so the actual constraint is that workspaces are only supported in NPM v7, which I think requires Node 15 - so ideally we put an NPM version constraint in package.json.

However, NPM and Node versions are maybe tied in some way as far as compatibility goes, so maybe the Node constraint is enough?

ryangjchandler:
You can actually do an NPM constraint as well as a Node one.

markjaquith:
>  I believe you can do this using the `engines` key in the `package.json` file

Unfortunately this will only give a warning, and only if you're installing this package as a dependency in another package.

I just tried it, and it didn't warn me on `npm i`, and `npm run build` blew up. ☹️ 

I know it's an extra file, but `.nvmrc` makes it trivial. With that file present, when I `cd` into the Alpine git repo directory, my command line switches (including downloading and installing if necessary) to the correct node version. Pretty great developer experience.

SimoTod:
@markjaquith doesn't it switch only if you do `nvm use`?

thasmo:
@SimoTod, I guess @markjaquith has a setup where it would run `nvm use` automatically when changing directories; there are plugins/scripts for that.

A `.nvmrc` file does not force to use a specific Node version, nor does it run anything automatically by default - it's "just" a configuration file for `nvm`, which is a tool for managing/running multiple `Node` versions on a single machine. People used to working with `nvm` will see the `.nvmrc` file and can rely on it to switch to the desired/compatible `Node` version.

I'm used to `nvm` because we support multiple projects using different `Node` versions, but I can understand that adding yet another configuration file may not be desirable.

That said, I too think adding a `.nvmrc` file would be nice for those using `nvm`. 👍

markjaquith:
@thasmo is correct.

The [NVM installation instructions](https://github.com/nvm-sh/nvm#deeper-shell-integration) include example scripts that handle shell integration. Highly recommend if you bounce around between projects with different supported Node versions!

KevinBatdorf:
If you NVM you might want to check out FNM. In my experience it's much faster when opening a terminal.

https://github.com/Schniz/fnm

It also supports .nvmrc files. 


--

My opinion on whether it should be an added file. I think the visual noise cost of adding a file takes away more than the benefit the file provides. That's just me though. 

I wonder why NVM doesn't support adding a rule to package.json. that seems like an oversight.

calebporzio:
Thanks for the input everyone - I'm going with @KevinBatdorf's take on this and closing for those reasons - thanks!

PR #1802, opened 2021-07-16
feature: support esc shorthand

ryangjchandler:
This pull request adds support for the `.esc` modifier, acting as a shorthand for the `.escape` key modifier.

### DIFF
--- a/packages/alpinejs/src/utils/on.js
+++ b/packages/alpinejs/src/utils/on.js
@@ -168,8 +168,12 @@ function keyToModifier(key) {
         case ' ':
         case 'Spacebar':
             return 'space'
-        case 'esc':
-            return 'escape'
+        case 'Escape':
+            return 'esc'
+        case 'Meta':
+            return 'cmd'
+        case 'Control':
+            return 'ctrl'
         default:
             return key && kebabCase(key)
     }
--- a/packages/alpinejs/src/utils/on.js
+++ b/packages/alpinejs/src/utils/on.js
@@ -168,6 +168,8 @@ function keyToModifier(key) {
         case ' ':
         case 'Spacebar':
             return 'space'
+        case 'esc':
+            return 'escape'
         default:
             return key && kebabCase(key)
     }


calebporzio:
Reasoning behind this? Demand for this?

ryangjchandler:
#1791 changed the documentation that originally said `.esc`, so I'm guessing you thought it was that instead of `.escape` at some point. 

It's also shorter, never a bad thing. Most keyboards actually say Esc as opposed to Escape. The overhead is basically zero, so I can't see any reason for not merging.

markjaquith:
I don't know if this sways anything, but in addition to `.esc` being wrong in the documentation, I just realized two other things are wrong:

1. `.ctrl` is wrong, as the key is `Control`, not `Ctrl`. `@keydown.ctrl` never works — it has to be `@keydown.control`.
2. `.cmd` is wrong because that key is called `Meta` on macOS. `@keydown.cmd` never works — it has to be `@keydown.meta` (with obvious macOS/Windows implications).

Gonna hold off on PRs for these as this needs some deeper discussion about how it should work.

My gut feel is that `.cmd` should work, `.esc` should work, and `.ctrl` should work.

ryangjchandler:
@markjaquith Agreed. Those are definitely needed in my opinion. 

If @calebporzio agrees, I can update this PR to include those as well.

calebporzio:
Yeah, let's do it. Thanks

calebporzio:
Checking in on this @ryangjchandler - care to add those other shorthands?

ryangjchandler:
@calebporzio Sorry, been a busy few weeks. Will sort it out now.

markjaquith:
@ryangjchandler

I used this for testing (only have access to macOS right now):

```html
<!DOCTYPE html>
<html>
	<head>
		<script src="./packages/alpinejs/dist/cdn.js" defer></script>
	</head>

	<body>
		<div x-data="{events: []}">
			<input placeholder="esc" type="text" @keydown.esc.prevent="events.push('esc')" /><br />
			<input placeholder="escape" type="text" @keydown.escape.prevent="events.push('escape')" /><br />
			<input placeholder="ctrl" type="text" @keydown.ctrl.prevent="events.push('ctrl')" /><br />
			<input placeholder="alt" type="text" @keydown.alt.prevent="events.push('alt')" /><br />
			<input placeholder="meta" type="text" @keydown.meta.prevent="events.push('meta')" /><br />
			<input placeholder="cmd" type="text" @keydown.cmd.prevent="events.push('cmd')" /><br />
			<input placeholder="shift" type="text" @keydown.shift.prevent="events.push('shift')" /><br />
			<input placeholder="space" type="text" @keydown.space.prevent="events.push('space')" /><br />

            <h4>Events</h4>
            <ul>
                <template x-for="event in events">
                    <li x-text="event"></li>
                </template>
            </ul>
		</div>
	</body>
</html>
```

Works:

- esc
- ctrl
- alt
- cmd
- shift
- space

Does not work:

- escape
- meta

ryangjchandler:
@markjaquith Yeah, so this is a bit of a catch with how the `keyToModifier` function is actually used. What it does is take the value of `event.key` and converts it to a valid modifier. I wonder if we should actually be doing a `keyToModifier(e.key).includes()` check and then returning an array instead of a single string. @calebporzio What do you think?

calebporzio:
@ryangjchandler - yeah I'm down with that - I don't have strong opinions about that function

ryangjchandler:
Awesome. I'm away from my laptop today but will sort that out tomorrow.

calebporzio:
Any update here?

ryangjchandler:
Not had time to get back to this. Anybody who wants to do this, feel free.

calebporzio:
Lol, just fixed and merged myself - ty

PR #1808, opened 2021-07-18
update arrow key docs

broothie:
AFAICT, the `.up`-like arrow key shorthands mentioned in the [keyboard events](https://github.com/alpinejs/alpine/blob/919d5717263fc2ccfb6de79a8200fd8021bfc755/packages/docs/src/en/directives/on.md#keyboard-events) don't work. I found that they _did_ work when I used the kebab-case form from [`KeyboardEvent.key`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values#navigation_keys), i.e. `.arrow-up`.

### DIFF
--- a/tests/cypress/integration/directives/x-on.spec.js
+++ b/tests/cypress/integration/directives/x-on.spec.js
@@ -263,6 +263,26 @@ test('keydown modifiers',
     }
 )
 
+test('keydown arrow modifiers',
+    html`
+        <div x-data="{ count: 0 }">
+            <input type="text"
+                x-on:keydown.up="count++"
+                x-on:keydown.arrow-down="count++"
+            >
+
+            <span x-text="count"></span>
+        </div>
+    `,
+    ({ get }) => {
+        get('span').should(haveText('0'))
+        get('input').type('{upArrow}')
+        get('span').should(haveText('0'))
+        get('input').type('{downArrow}')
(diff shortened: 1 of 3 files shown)


broothie:
Interesting -- according to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values#navigation_keys) (at the end of the **Navigation Keys** section):

> [1] Internet Explorer, Edge (16 and earlier), and Firefox (36 and earlier) use "Left", "Right", "Up", and "Down" instead of "ArrowLeft", "ArrowRight", "ArrowUp", and "ArrowDown".

tests/cypress/integration/directives/x-on.spec.js
```diff
@@ -263,6 +263,26 @@ test('keydown modifiers',
     }
 )
 
+test('keydown arrow modifiers',
+    html`
+        <div x-data="{ count: 0 }">
+            <input type="text"
+                x-on:keydown.up="count++"
+                x-on:keydown.arrow-down="count++"
```
broothie, line 271: `x-on:keydown.up` doesn't work
`x-on:keydown.arrow-down` works

calebporzio:
Thanks however, I think this more a bug with Alpine. I just fixed Alpine to support `up`, etc... will be available in the next release

PR #1816, opened 2021-07-21
feature: add alias to $persist

pedroborges:
This pull request adds support for $persist aliases, just like in the $queryString plugin.

```html
<div x-data="{ form: $persist({}).as('form-0Jp3d') }"></div>
```

This useful for cases where you have multiple data components on the same page. With alias you can give each localStore entry a unique key. I recently needed it for a form component.


calebporzio:
Thanks for this - I actually just added this myself before I saw this PR - good instincts!

PR #1817, opened 2021-07-21
feature: support other data types for $persist

pedroborges:
This pull request adds support for other data types to $persist. The current implementation doesn't preserve the type nor supports `array` or `object`.


calebporzio:
Thanks for this - I actually just implemented this myself before I saw this - good instincts!

PR #1819, opened 2021-07-21
Fix x html with alpine content when sibling of xfor or xif

tvdr:
This pull request fixes the bug discovered in #1688 

The problem generally is that when one wants to insert alpine code with x-html but it is preceded by x-for or x-if , then the for/if directives pause observing mutations and as walking through the elements, initialize the x-html.
As the mutationobserver is paused, the alpine content inserted from the x-html wont "render".

Wrapping the el.innerHTML assignment in a nextTick solves the issue, as mutationobserver is turned back on.

Codepen showing non working examples : [Alpine3 x-html with alpine data render error](https://codepen.io/tvdr/pen/eYWvpbm) 

Same code with the fixed directive : [Alpine3 x-html with alpine data render fix](https://codepen.io/tvdr/pen/bGWrvep)

### DIFF
--- a/packages/alpinejs/src/directives/x-html.js
+++ b/packages/alpinejs/src/directives/x-html.js
@@ -1,13 +1,14 @@
-import { evaluateLater } from '../evaluator'
-import { directive } from '../directives'
-import { mutateDom } from '../mutation'
+import {directive} from '../directives'
+import {nextTick} from "../nextTick";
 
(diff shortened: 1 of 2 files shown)


calebporzio:
Thanks for submitting this. $nextTick works, but it's a hack around an underlying bug. I fixed the root problem in this PR: #1901

Issue #1805, opened 2021-07-16
x-for runs before an x-if that is wrapping it


AdamSCMM:
### Discussed in https://github.com/alpinejs/alpine/discussions/1730

<div type='discussions-op-text'>

<sup>Originally posted by **AdamSCMM** July  6, 2021</sup>
I have code where an x-if wraps an x-for to stop the x-for from running if the variables are not correct. In v2 this code worked correctly, in v3 the x-for throws an error: `can't access property "length", items is undefined` even though the x-if should be stopping it beforehand.

I've tried a few different ways of writing the if statement, but nothing seems to make a difference

The code looks like:

```
<template x-if="currentDay in days">
or
<template x-if="days[currentDay]">
    <template x-for="(_numbers, _index) in days[currentDay]" :key="_index">
```

You can see a working example at the codepen below. Click 5 or 7 and then null or clear. It breaks when it should not. 5 to 7 works fine

https://codepen.io/acxz/pen/mdmeJRo?editors=1010

It was already suggested to me to add a `|| []` at the end of the for to account for the collection being undefined, but that shouldn't be necessary. This situation should not be occurring like this</div>

wlhrtr:
I am having a similar issue if `x-if` and `x-for` are on the same `<template>` tag.

Worked fine in v2.

Hat to wrap the `x-for` in another `<template>` with the `x-if` on it.

```
<template x-if="visible">
    <template x-for="day in days">
        ...
    </template>
</template>
```

PR #1827, opened 2021-07-22
fix to x-for breaking with error when iterated item is undefined

tvdr:
possible fix for #1805 

By checking if item is not undefined, the loop function does not throw a breaking error, but properly cleans up inserted dom elements.

I think this may be the expected behaviour, even when not wrapped in x-if.  
Tests are commited for both cases.

non working codepen from the issue: [https://codepen.io/acxz/pen/mdmeJRo?editors=1010](https://codepen.io/acxz/pen/mdmeJRo?editors=1010)

same example with after fix [https://codepen.io/tvdr/pen/bGWojwd](https://codepen.io/tvdr/pen/bGWojwd)


calebporzio:
Thanks for submitting this, however I fixed it on master using a different implementation. Thanks again!

PR #1829, opened 2021-07-22
Remove leading blank line in intersect plugin

markjaquith:
A few more like this and the plugin will be *negative* 311 bytes. 😝


calebporzio:
Lol, thanks

PR #1841, opened 2021-07-24
Bug ($persist) Multiple components

SimoTod:
This PR fixes an anomaly which we can see on the documentation page: https://alpinejs.dev/plugins/persist.

When there are 2 components on the same page using the same property, the last one is setting the value in local storage as expected but when reloading the page, only the first component is initialised correctly whilst the second component uses the default of the first one. I appreciate this is not a real use case, the 2 components won't communicate with each other out of the box after the page is loaded so they would behave weirdly but I think they should be consistent.

To replicate
- go to https://alpinejs.dev/plugins/persist#how-it-works
- scroll down to the example (note that above there is another example using the same component)
- increment the counter
- reload the page
- note that the component in https://alpinejs.dev/plugins/persist#magic-persist behaves correctly but the component in https://alpinejs.dev/plugins/persist#how-it-works doesn't 

I've also added a couple of tests for booleans and objects.

Discussed in #1834

### DIFF
--- a/packages/persist/src/index.js
+++ b/packages/persist/src/index.js
@@ -10,7 +10,7 @@ export default function (Alpine) {
                 ? storageGet(lookup)
                 : initialValue
 
-            setter(initialValue)
+            setter(initial)
 
             Alpine.effect(() => {
                 let value = getter()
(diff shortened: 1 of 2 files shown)


calebporzio:
This is great - thanks @SimoTod

dsturm:
Thank you @SimoTod! 🎉

PR #1854, opened 2021-07-26
fix: set custom properties in style correctly

pohlt:
This fixes the issue mentioned here: https://github.com/alpinejs/alpine/discussions/1807

@KevinBatdorf Thanks for the initial bugfix. Could you please have a look?

### DIFF
--- a/packages/alpinejs/src/utils/styles.js
+++ b/packages/alpinejs/src/utils/styles.js
@@ -13,7 +13,7 @@ function setStylesFromObject(el, value) {
     Object.entries(value).forEach(([key, value]) => {
         previousStyles[key] = el.style[key]
 
-        el.style[key] = value
+        el.style.setProperty(key, value)
     })
 
     setTimeout(() => {
--- a/tests/cypress/integration/directives/x-bind-style.spec.js
+++ b/tests/cypress/integration/directives/x-bind-style.spec.js
@@ -21,3 +21,25 @@ test('style attribute object bindings are merged with existing styles',
         get('span').should(haveAttribute('style', 'display: block; color: red;'))
     }
 )
+
+test('CSS custom properties are set',
+    html`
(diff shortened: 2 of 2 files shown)


KevinBatdorf:
You probably want to test the surrounding functionality too, specifically this line.

```js
previousStyles[key] = el.style[key]
```
https://github.com/alpinejs/alpine/pull/1854/files#diff-0cbf36cdc048c0c051e4a96236cc2dd86fb1028070bc58f5ab620256ec4d60ecR14

Does it preserve the custom property? Does it need to be converted to `.getPropertyValue` maybe (I haven't looked, so maybe not)?

pohlt:
I added another test. Does that cover the case you had in mind?

pohlt:
Works in my fork: https://github.com/pohlt/alpine/runs/3162656086

KevinBatdorf:
Yeah I think that was worth adding. But wait and see what Caleb says.

pohlt:
@KevinBatdorf You don't happen to have the rights to start the test workflow?

KevinBatdorf:
What do you mean? To run the tests? They should run when you commit. You can remove you change so the tests fail, the commit the change again so they pass.

Is that what you mean?

I don't have any repo/org privledges though. I just try to help where I can.

pohlt:
I get this message as a last  entry in this PR
image
...so I was wondering if you could give your approval, but you don't seem to have the required rights. Thanks anyway.

KevinBatdorf:
Ah dang. I haven't seen that before either.

calebporzio:
Interesting re: needing approval to run. It appears this is a new default for GitHub. I changed the setting to be more lenient, although there is no option to simply allow all users to run actions:
image

I imagine this is a result of the recent Actions abuse they've been facing.

You should be good to run actions now

calebporzio:
Anyhow, thanks for this!

PR #1874, opened 2021-07-29
Bug: init() function doesn't have access to the parent scope

SimoTod:
Bug description: The init function that gets called automatically doesn't have access to the parent scope while x-init does.

Mentioned in #1674 and #1711

Relevant changes:
- changed simple function call to `evaluate` so the function gets full access to all Alpine magic.
- removed .only from 2 tests in custom-data.spec.js since the other tests were not running
- added test for the specific use case which fails without the fix (See https://github.com/alpinejs/alpine/runs/3189986654?check_suite_focus=true#step:6:72)
- added minimal test for `destroy` as that line of code was updated as well

Note:
I didn't run many tests for `destroy` since we might need a follow-up. In the current implementation, we call `undo` first and then `destroy` which means that the scope is removed before we call `destroy`. I believe they should be swapped but I'm not sure if it's been done on purpose so I didn't touch it in this PR.

### DIFF
--- a/packages/alpinejs/src/directives/x-data.js
+++ b/packages/alpinejs/src/directives/x-data.js
@@ -29,11 +29,11 @@ directive('data', skipDuringClone((el, { expression }, { cleanup }) => {
 
     let undo = addScopeToNode(el, reactiveData)
 
-    if (reactiveData['init']) reactiveData['init']()
+    reactiveData['init'] && evaluate(el, reactiveData['init'])
 
     cleanup(() => {
         undo()
 
-        reactiveData['destroy'] && reactiveData['destroy']()
+        reactiveData['destroy'] && evaluate(el, reactiveData['destroy'])
     })
 }))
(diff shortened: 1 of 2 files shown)


calebporzio:
Hey Simone, interesting - I wouldn't intuit that I have access to parent scope from anywhere inside a "data component" object.

I suppose this IS something available to you in any other method that is called from an element though yes?

pomartel:
@calebporzio Yes, the parent scope is available from other methods. You can even rename data component `init` function to `initialize` for example and then add `x-init='initialize'` to your markup and the parent scope will then be available. That's how I work around the bug for now while still using data components.

pomartel:
By the way, the bug is not only there with `Alpine.data()` objects but regular components as well. I made a codepen that illustrates it (and the workaround): https://codepen.io/pomartel/pen/dyWjGpv

SimoTod:
Yeah, basically if you use x-init, you can access the parent scope since the directives uses evaluate while the auto init calls the function directly so it doesn't. If it's the expected behavior, feel free to close it.

calebporzio:
Ok, good points. I'm merging this - thanks!

piotrze:
@SimoTod Thanks! It was missing piece for my nested components.

PR #1879, opened 2021-07-31
Update store.md

ktpss95112:
When using alpinejs, I could not found some functionalities in the documentation.
After reading the source code, I found what I needed.
Therefore, I updated the documentation, hoping to help someone else.


calebporzio:
Thanks!

PR #1880, opened 2021-07-31
Debounce plugin

tanthammar:
Plugin that adds a magic $debounce directive

### DIFF
--- a/packages/debounce/builds/cdn.js
+++ b/packages/debounce/builds/cdn.js
@@ -1,0 +1,5 @@
+import debounce from '../src/index.js'
+
+document.addEventListener('alpine:init', () => {
+    window.Alpine.plugin(debounce)
+})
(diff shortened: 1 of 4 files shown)


tanthammar:
Example

```html
<div x-data="{
    foo(text) {
        $debounce(() => console.info(text), 500)
    }
}">
    <button @click="foo('Hello')"></button>
</div>
```

SimoTod:
It doesn't seem to work for me, every time alpine evaluates $debounce, it returns a new callback and timeoutId is always undefined so each click triggers the function. It's probably worth adding some tests to make sure it works now and in the future.

Also, it's worth asking Caleb if he's willing to accept it into the core first.

tanthammar:
Thank you for the feedback. 

This, and the throttle PR, are more like feature requests for something that I would like in the core. 

I am hoping that someone with better knowledge of Alpine can improve it so Caleb can consider to pull it in. 

Or, if there is a better way to get these directives.

calebporzio:
Thanks Tina, going to hold off on this for now though - thanks for contributing!

PR #1881, opened 2021-07-31
Throttle plugin

tanthammar:
Plugin that adds a magic $throttle directive

This one is tricky, because I don't know how to separate the throttling state for each component.

Hoping that someone knows how to fix that :)


tanthammar:
Example

```html
<div x-data="{
    foo(text) {
        $throttle(() => console.info(text), 500)
    }
}">
    <button @click="foo('Hello')"></button>
</div>
```

calebporzio:
Thanks Tina, going to hold off on this magic for now though - thanks!

PR #1904, opened 2021-08-03
feature: improved logging / error handling ( alternate strategy )

JacobPozaic:
This is an alternative implementation off the previous PR ( https://github.com/alpinejs/alpine/pull/1607 ).  One or the other should be chosen, but not both.

Incorporates suggestions from @HugoDF regarding dispatching an event.

We found the dev experience of this version is better as it doesn't cause false down-stream errors. ( other PR causes additional errors when async evaluators error and try to rethrow )

### DIFF
--- a/packages/alpinejs/src/utils/error.js
+++ b/packages/alpinejs/src/utils/error.js
@@ -9,9 +9,7 @@ export function tryCatch(el, expression, callback, ...args) {
 }
 
 export function handleError(error, el, expression = undefined) {
-    Object.assign( error, { el, expression } )
-
     console.warn(`Alpine Expression Error: ${error.message}\n\n${ expression ? 'Expression: \"' + expression + '\"\n\n' : '' }`, el)
 
-    dispatch( document, "alpine:error", { error } )
+    dispatch( document, "alpine:error", { error, el, expression } )
 }
--- a/packages/alpinejs/src/utils/error.js
+++ b/packages/alpinejs/src/utils/error.js
@@ -9,7 +9,11 @@ export function tryCatch(el, expression, callback, ...args) {
 }
 
 export function handleError(error, el, expression = undefined) {
+    Object.assign( error, { el, expression } )
+
     console.warn(`Alpine Expression Error: ${error.message}\n\n${ expression ? 'Expression: \"' + expression + '\"\n\n' : '' }`, el)
 
     dispatch( document, "alpine:error", { error, el, expression } )
+
+    setTimeout( () => { throw error }, 0 )
 }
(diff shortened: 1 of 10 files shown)


packages/alpinejs/src/utils/error.js
```diff
@@ -0,0 +1,17 @@
...
+        handleError( e, el, expression )
+    }
+}
+
+export function handleError(error, el, expression = undefined) {
+    Object.assign( error, { el, expression } )
+
+    console.warn(`Alpine Expression Error: ${error.message}\n\n${ expression ? 'Expression: \"' + expression + '\"\n\n' : '' }`, el)
+
+    dispatch( document, "alpine:error", { error } )
```
HugoDF, line 16: We probably want to pass the element and expression in here (if they're set)

danddanddand: Okay, we can add that. They were getting assigned to the error but I agree

danddanddand: We updated it @HugoDF to be off error object and on main detail object if you want to look again

JacobPozaic:
For the two failing tests they seem to have legitimate errors as far as I can tell, that surface after these changes.

For example in store.spec increment() is not defined, the test passed before because the event listener in the alpine-init code handles the click where the @click directive would not have worked.

danddanddand:
Any update with getting this merged? I've been using it for development and it's extremely useful.

danddanddand:
I've also found some production benefits to this branch. For example, we had google charts api fail when loaded by alpine (just due to CND network) and it would cause all of alpine to lock up but on this branch unrelated components continue to function. 

Also had some unescaped code causing a component to fail that was locking up alpine and causing listing dropdowns and all kinds of stuff not to work, but on this branch that component fails and is logged and alpine continues to work. 

So there have been a few unexpected things that this branch solves for us that is especially important in prod.

Pronian:
I'm also eagerly awaiting for this to be merged

danddanddand:
I fixed the store.spec.js and customs-directives.spec.js tests that were failing on this branch once error handling was implemented.

If there are any questions or feedback that could get this merged I'd really appreciate it. We use this branch in production because it is more fault tolerant (prevents failing component from easily locking up Alpine which we experienced in our app a few times). This has proven to have production and development benefits for our team. 

The actual error handling strategy (event and re-throw) is isolated to one spot so if @HugoDF or @calebporzio prefer one strategy instead of both I'm indifferent to that part of it. We were just trying to make it easy for dev tools to support.

calebporzio:
Thanks for wanting to tackle this. However this PR is stale and the conversation has many different points in it.

If you still want this, please re-open with a thorough summary. Thanks again!

PR #1906, opened 2021-08-03
Add .dot modifier

quasipickle:
This modifier works similarly to the `.camel` modifier.  If it is provided, the event name is converted from dash format to dot format (such as used by Bootstrap).

ex:

```
@shown-bs-modal.dot.window = "alert('Caught')"
```

Will cause an event listener to be attached to the window, listening for the `shown.bs.modal` event.


quasipickle:
Discussion: https://github.com/alpinejs/alpine/discussions/1885

calebporzio:
Thanks for talking this over in the discussion first and thanks for the contribution - perfect PR!

PR #1910, opened 2021-08-05
feat(x-bind, x-html, x-text) add ssr modifier

MartinPL:
This is idea how to connect server side rendering with Alpine, not sure if it is the best way. Lets open dissicuse with this PR :)

### DIFF
--- a/packages/alpinejs/src/directives/x-bind.js
+++ b/packages/alpinejs/src/directives/x-bind.js
@@ -12,7 +12,11 @@ directive('bind', (el, { value, modifiers, expression, original }, { effect }) =
 
     let evaluate = evaluateLater(el, expression)
 
+    let skippedFirstTime;
+
     effect(() => evaluate(result => {
+        if (modifiers.includes('ssr') && !skippedFirstTime) return skippedFirstTime = true;
+
         // If nested object key is undefined, set the default value to empty string.
         if (result === undefined && expression.match(/\./)) result = ''
 
--- a/packages/alpinejs/src/directives/x-html.js
+++ b/packages/alpinejs/src/directives/x-html.js
@@ -2,11 +2,13 @@ import { evaluateLater } from '../evaluator'
 import { directive } from '../directives'
 import { mutateDom } from '../mutation'
 
-directive('html', (el, { expression }, { effect, evaluateLater }) => {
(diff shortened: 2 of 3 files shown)


calebporzio:
Interesting, please explain more about how one would use this? In what instance is it bad to overwrite what comes from the server the first time? Thanks!

MartinPL:
This is extremely helpful when we have already fully rendered website (in my case by PHP / Twig) and we want to Alpine kicks in on user interaction.
So we don't need to make some not handy workarounds like delay initialization or make some painful backend data transformation to re-render what is already on site. We just skip first Alpine render.

calebporzio:
Yeah, I understand that broadly. Can you provide a concrete example of when this is helpful? Ideally with code? Thanks.

It's unlikely that I'll add this to the core API at this moment, but I really want to understand the use case because I think this could be a valuable feature as a plugin or in core at some point.

Thanks!

skttl:
So would
```
<div x-data="{ something: '' }">
    <div x-bind="something">the value</div>
</div>
```

make Alpine initialise `something` as `the value`?

That would be super awesome!

ryangjchandler:
@skttl No, I believe the idea is that the `.ssr` modifier would skip the initial render for those directives. E.g.:

So in your example, if you set `something` to `foo` and had the `.ssr` modifier, your `<div>` would still have `the value` as it's `textContent`, until `something` is changed.

MartinPL:
Two examples from me. I wanted a second example to be a list with filters, but PokeAPI don't have that functionality :man_shrugging: 

Random pokemon: https://codepen.io/MartinPLr/pen/yLXYZjE
Pokemon list: https://codepen.io/MartinPLr/pen/GREpzdj

calebporzio:
I guess the reason I'm confused, is you can always seed the HTML with content, and then let Alpine take over and it will keep the change.

Of course this assumes that the Alpine data matches what was seeded.

If the data doesn't match and this is the reason for .ssr - that makes sense.

However, the overhead to add something like this that only applies to those specific situations where you want to seed the HTML and the Alpine data is different doesn't seem worth it to me at least at first glance.

I'm going to close this. Re-open with more reasoning if you disagree. Thanks!

PR #1911, opened 2021-08-05
Fix throttle on key events

sterat:
Add a test for the case when throttle is used with key events, and fix.

The handler was never called when throttle is used on key events (keyup and keydown).

This is because `isListeningForASpecificKeyThatHasntBeenPressed()` filters out a list of known event types and assumes that any remaining modifiers are keys to wait for. Unfortunately it omits throttle from the list of known events, so it waits for an invalid key to be pressed.

This fix moves all parsing of the modifiers to the top of `on()` so the repeat parsing is avoided.

### DIFF
--- a/packages/alpinejs/src/utils/on.js
+++ b/packages/alpinejs/src/utils/on.js
@@ -1,5 +1,5 @@
 
-export default function on (el, event, modifiers, callback) {
+export default function on (el, event, strModifiers, callback) {
     let listenerTarget = el
 
     let handler = e => callback(e)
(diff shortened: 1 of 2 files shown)


SimoTod:
@sterat It's usually a good idea to keep fixes and refactoring separated so it's easier for the maintainer to review. In general, if a PR for an OOS project does 2 things, there should be 2 PRs.
The throttle fix 
```
if (keyModifiers.includes('throttle')) {
        let debounceIndex = keyModifiers.indexOf('throttle')
        keyModifiers.splice(debounceIndex, isNumeric((keyModifiers[debounceIndex+1] || 'invalid-wait').split('ms')[0]) ? 2 : 1)
    }
 ```
would probably get merged sooner than the refactoring.

calebporzio:
Exactly what @SimoTod said. Thanks @SimoTod - please separate this out into 2 PRs

calebporzio:
Closing due to inactivity - please re-open if comments are addressed

PR #1928, opened 2021-08-10
feature: Dispatch event based x-intersect

MuzafferDede:
This PR 
- It adds `@enter`, `@leave`, `@changed` events on elements which uses `x-intersect` directive, without any breaking changes.
- It implements new `x-intersect` tests and improves the old ones.
- It was very cheap to achieve this functionality which had high demands. 

### `directive`
```html
<div x-intersect="shown = true">...</div>
```
### `enter`
```html
<div x-intersect @enter="shown = true">...</div>
```
### `leave`
```html
<div x-intersect @leave="shown = false">...</div>
```
### `changed`
```html
<div x-intersect @changed="shown = true">...</div>
```

It also passes dispatch event so we could get the event as

```html
<div x-intersect @enter="event => console.log(event)">...</div>
```

> Update => Renamed event name `change` to `changed` to prevent issues on `input` elements

### DIFF
--- a/packages/alpinejs/src/directives.js
+++ b/packages/alpinejs/src/directives.js
@@ -1,6 +1,7 @@
 import { onAttributeRemoved, onElRemoved } from './mutation'
 import { evaluate, evaluateLater } from './evaluator'
 import { elementBoundEffect } from './reactivity'
+import { dispatch } from './utils/dispatch'
 import Alpine from './alpine'
 
 let prefixAsString = 'x-'
@@ -76,6 +77,7 @@ export function getDirectiveHandler(el, directive) {
         Alpine,
         effect,
         cleanup,
+        dispatch,
         evaluateLater: evaluateLater.bind(evaluateLater, el),
         evaluate: evaluate.bind(evaluate, el),
     }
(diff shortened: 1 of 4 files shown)


MuzafferDede:
I also manage to create another solution using `x-intersect:[value]` as

```html
    <div x-data="{ count: 0 }">
        <span x-text="count"></span>

        <div 
            x-intersect="console.log('default enter only')" 
            x-intersect:enter="console.log('enter only')" 
            x-intersect:leave="console.log('leave only')" 
            style="margin-top: 100vh;" id="1">hi
        </div>
    </div>
```
which does not involves any custom events. It is more "Alpine" like solution, depending on @calebporzio decision, i can push new PR with this solution.

KevinBatdorf:
I think your second iteration makes more sense as it matches more closely with x-transition.

MuzafferDede:
> I think your second iteration makes more sense as it matches more closely with x-transition.

Agreed. Most likely second will be the best fit. It's even cheaper.

RJFares:
yea, second approach more intuitive :+1:

MuzafferDede:
Since the second approach is more likely going to be merged, i am closing this one and will open the modifier based one.

PR #1943, opened 2021-08-12
Fix in docs about x-if directive

koenwoortman:
I was doing some testing and noticed that `x-show` sets the inline style: `style="display: none;"`


calebporzio:
Good catch, thanks!

PR #1965, opened 2021-08-18
fixed #1959 #1962 x-trap crawl bug

MuzafferDede:
This PR fixes

- #1959
- #1962
- Improves tests

This would break in current `x-trap`
```html
    <button @click="open = true">Open Dialogue</button>

    <div>
        <div x-trap="open" x-show="open">
            <input type="text" placeholder="Some input...">
            <input type="text" placeholder="Some other input...">

            <button @click="open = false">Close Dialogue</button>
        </div>
        <!-- No other children -->
    </div>
</div>
```
Bug reason is that `crawlSiblingsUp` had a wrong same node check on el and sibling. So when `x-trap` node have a parent node, focus would not trap.

Tests were not good enough to check the case, so tests are now improved.

### DIFF
--- a/packages/trap/src/index.js
+++ b/packages/trap/src/index.js
@@ -42,11 +42,9 @@ function crawlSiblingsUp(el, callback) {
     if (el.isSameNode(document.body)) return
 
     Array.from(el.parentNode.children).forEach(sibling => {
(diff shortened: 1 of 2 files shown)


calebporzio:
Great, thank you!

PR #1969, opened 2021-08-18
fixed safari autofill bug

MuzafferDede:
This PR 
- fixes #1909
- fixes #1862
- fixes #1810
- fixes #1501

Fix was very simple as Safari triggers `autofill` event as CustomEvent instead of `input` event, even Alpine handles customEvent but it did not handle the assigned value properly. So i just fixed that part.

I don't have an iOS device to check but so far i could only check desktop safari and its seems fixed. It would be nice to have some feedbacks if anyone can try this.

### DIFF
--- a/index.html
+++ b/index.html
@@ -15,12 +15,4 @@
             Content...
         </span>
     </div>
-
-<div x-data="{username: '', password: ''}" >
-    <input x-model="username" type="text" name="username" id="username">
-    <input x-model="password" type="password" name="password" id="password">
-    <div x-text="username">test</div>
-    <div x-text="password">test</div>
-    <button @click="username = null, password=null">TEst</button>
-</div>
 </html>
--- a/index.html
+++ b/index.html
@@ -15,4 +15,12 @@
             Content...
         </span>
     </div>
+
+<div x-data="{username: '', password: ''}" >
+    <input x-model="username" type="text" name="username" id="username">
+    <input x-model="password" type="password" name="password" id="password">
+    <div x-text="username">test</div>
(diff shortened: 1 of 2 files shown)


ralphjsmit:
Hey MuzafferDede, thank you so much for providing this. I'd love to try it, but I'm unsure how to do so. I'm using it from a script tag, is that überhaupt possible to try out a PR?

MuzafferDede:
Oh, good point, you can use this [codepen](https://codepen.io/muzafferdede/pen/KKmjRdb) to check iOS/macOS Desktop safari. Thanks

MuzafferDede:
@ralphjsmit Good point, i just created this [codepen](https://codepen.io/muzafferdede/pen/KKmjRdb) for those who wants to try and give some feedback. Cheers

ralphjsmit:
Works like a charm, perfect! How did you found out about this?

I'm absolutely not a JS developer, PHP is more my thing, but just wondering how you came on the idea that this could be it.

tombien:
I take your code of Alpine from codepen and put it into my app. Works perfect. Tested Safari, Chrome, Firefox on MacOS BigSur and Safari iPhone. Everything works as expected. Thanks!

SimoTod:
Nice finding. To be fair ie11 is not even supported by Alpine anymore so that initial if case could also be removed. (more a thought for Caleb to consider, the fix and new tests are great).

calebporzio:
Thanks @MuzafferDede!

calebporzio:
And also thanks everyone else for testing this and verifying this

PR #1973, opened 2021-08-19
Add instructions to use $persist with Alpine.data

SimoTod:
Feel free to reword it to make it clearer.
A few people are trying to use `$persist` with `Alpine.data` but all examples on the docs use arrow functions.
Alpine tries to bind the custom context [here](https://github.com/alpinejs/alpine/blob/main/packages/alpinejs/src/datas.js#L13) but in javascript is not possible to re-bind `this` for arrow functions.

This applies every time we try to use `this` to define a property value in Alpine.data but currently `$persist` is the only magic that needs it. Using `this` inside a function defined in x-data works as expected since at that point the object has been "Alpinefied" by the custom evaluate function.


calebporzio:
Perfect, thanks for this @SimoTod !

Issue #1983, opened 2021-08-25
x-trap error when using entangle


manogi:
### Discussed in https://github.com/alpinejs/alpine/discussions/1982

<div type='discussions-op-text'>

<sup>Originally posted by **manogi** August 25, 2021</sup>
Since a few days (I think after this fix https://github.com/alpinejs/alpine/pull/1965) I get an error message when using x-trap and entangle together. It worked before.

I am using the version 3.2.4 for both alpinejs and x-trap.

This should be enough to reproduce the error ($show is a PHP variable coming from Livewire):

``` html
<div 
    class="
        fixed top-0 w-full h-full overflow-visible
        transition-left duration-300 ease-in-out
        @if($show)
            left-0
        @else
            left-full
        @endif
    "
    x-data="{ open: @entangle('show') }"
>
    <div x-trap="open">
        <p>content removed for brevity</p>
    </div>
</div>
```

This is the error in the console:

```
Uncaught (in promise) TypeError: Cannot read property 'children' of null
    at crawlSiblingsUp (app.js?id=d09d6ad179ad7772b322:552)
    at app.js?id=d09d6ad179ad7772b322:555
    at Array.forEach (<anonymous>)
    at crawlSiblingsUp (app.js?id=d09d6ad179ad7772b322:552)
    at app.js?id=d09d6ad179ad7772b322:555
    at Array.forEach (<anonymous>)
    at crawlSiblingsUp (app.js?id=d09d6ad179ad7772b322:552)
    at app.js?id=d09d6ad179ad7772b322:534
    at runIfTypeOfFunction (app.js?id=d09d6ad179ad7772b322:2231)
    at app.js?id=d09d6ad179ad7772b322:2214

```

Also see this discussion: https://github.com/alpinejs/alpine/discussions/1959</div>

PR #1986, opened 2021-08-25
Fixed shadow dom parent issue

MuzafferDede:
This PR fixes a bug caused by #1965 fix when using it with Livewire `@entangle ` .

Mentioned:
- #1983 
- #1982
- #1959 

Livewire `@entangle ` creates a bug on the Alpine components where parent node of the `x-trap` element is become a `document-fragment`. Just added `el.parentNode ` check on el to fix this.

### DIFF
--- a/packages/trap/src/index.js
+++ b/packages/trap/src/index.js
@@ -39,7 +39,7 @@ export default function (Alpine) {
 }
 
 function crawlSiblingsUp(el, callback) {
-    if (el.isSameNode(document.body)) return
+    if (el.isSameNode(document.body) || ! el.parentNode) return
 
     Array.from(el.parentNode.children).forEach(sibling => {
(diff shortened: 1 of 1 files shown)


calebporzio:
Is it possible to add a test for this? Thanks @MuzafferDede!

calebporzio:
Ok, I'll merge this, my only concern is that we reintroduce this bug at some point. Thanks for putting time on this @MuzafferDede!

manogi:
I will let you know if it is reintroduced LOL.
Seriously - this is such a great feature, I'm using it like this, sooooo clean:

```
    x-data="{ 
        open: @entangle('show'),
    }"
    x-trap="open"
```

MuzafferDede:
@calebporzio I will keep an eye on this and will find a way to test. Also folks who found reintroduce of this bug, can mention me and i will look into it.

PR #1993, opened 2021-08-27
Fix empty x-init

MuzafferDede:
This PR
- adds a check on `x-init` expression to skip `evaluate` if it's empty. 

This will allow server side conditional `x-init` expressions rendering such as:
```html
<div x-init="@if ($initialExpression) .... @endif">...</div>
```
Mentioned: 
- #1991


calebporzio:
Thanks!

PR #1994, opened 2021-08-27
Bug with $persist.as

SimoTod:
When using $persist(...).as, the alias is currently applying to all the next $persist calls rather than just the current one.
For example https://codepen.io/SimoTod/pen/XWgmJqZ
After changing the first value and refreshing the page, the update is applied to the second one despite having a different name. If you inspect localStorage, you can see that only one key is saved (the one using the alias).

This PR moves the local variable to the right scope so it doesn't conflict with other $persist calls 

Failing test case: https://github.com/SimoTod/alpine/runs/3447390431#step:6:958

### DIFF
--- a/packages/persist/src/index.js
+++ b/packages/persist/src/index.js
@@ -1,8 +1,8 @@
 
 export default function (Alpine) {
-    let alias
-
     Alpine.magic('persist', (el, { interceptor }) => {
+        let alias
+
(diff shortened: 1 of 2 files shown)


calebporzio:
Thanks @SimoTod!

And yeah, I'll comment on that discussion about my thoughts on a sessionStorage feature.

PR #1995, opened 2021-08-29
Expose stopObservingMutations

dsazup:
Hello, this PR introduces a possible fix for a problem with Turbo mentioned here https://github.com/alpinejs/alpine/discussions/1534. Currently there is a problem when livewire's `@entangle` is being used on Turbo visit. For example this bug can be seen with a fresh install of jetstream and Turbo added. Anywhere where jetstream opens a modal (or anywhere else entangle is used) you should see a console error and the modals would not work. 

Essentially what happens is alpine is being run before livewire on turbo visit because of mutation observer. One of the possible solutions is to expose `stopObservingMutations`. This way we can run it like so:

First, we stop observing mutations:

```js
document.addEventListener('turbo:before-render', (event) => {
    Alpine.stopObservingMutations()
})
```

Then we simply restart livewire and start alpine again. 
```js
document.addEventListener('turbo:render', (event) => {
    window.Livewire.restart()
    Alpine.start()
})
```
This makes `@entangle` work perfectly with Turbo. I have also tried exposing `startObservingMutations` and thought that maybe in `turbo:render` event it would be enough to start it again, but it didn't work as I expected and a full restart is needed, so I have not exposed that method in this PR.

Looking for some feedback, maybe there is a better way to do it - I am not sure, but I haven't found another way myself besides this.


bep:
I have tested this patch with Turbo in a standalone app (not Livewire) and I think it will enable fixes for "most similar issues".

That said, I have a test site up and running on https://agitated-kilby-da786e.netlify.app/ using the latest Alpine with this patch applied.

The best "Turbo setup" I have come up with is this:

```js
document.addEventListener('turbo:before-cache', () => {
	Alpine.stopObservingMutations();
	document.body.querySelectorAll('[data-turbo-permanent]').forEach((el) => {
		el._x_ignore = true;
	});
});

document.addEventListener('turbo:render', () => {
	if (document.documentElement.hasAttribute('data-turbo-preview')) {
		return;
	}
	Alpine.start();
});
```

Ideally I would remove the `_x_ignore` attributes in `turbo:render`, but that leads to lots of double-rendering of `x-for` elements. I don't understand the Alpine-internals good enough to see a fix for that.

Note that in the link above, `data-turbo-permanent` is applied to the tab switcher and both dropdowns.

/cc @SimoTod

khrome83:
Would love to see this included in the next version. Does this needed added to the docs to be accepted since this exposes a new `Alpine.` property?

calebporzio:
My hesitation with this addition is this:

If we expose a way to stop observing mutations, we should have a way to re-observe them and provide mutation "flushing" behavior.

Also, I'm nervous that people will over-use this API resulting in memory leaks where Alpine doesn't have the change to tear down it's DOM additions and listeners because it won't observe them.

khrome83:
That is a fair point. You could leave it undocumented. Would be no different than pre module libraries where people hack the internals and self harm.

That being said the reason for this is to make Alpine play nice with Turbo Drive. Which worked great with v2 and TurboLinks. I am unclear if the Alpine v3 or the Turbo Drive change (or both) causes this relationship to sour.

Not sure if you have any advice to solve that problem?

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

bep:
My 50 cents on this: 

1. To fix this, you would at least need a way to stop/pause the observer.
2. Then it would be good to eventually have some API to "re-observe them and provide mutation "flushing" behavior."

Assuming 2) is harder, I think exporting `stopObservingMutations` could be a good, undocumented, stop-gap workaround, which would remove the need for all of these poorly maintained Alpine-forks floating around.

calebporzio:
Ok, I've done thinking and work on this.

I added two new methods to Alpine. Here's the new code to make Livewire work with Alpine now:

```html
@livewireScripts
<script src="https://cdn.jsdelivr.net/gh/livewire/turbolinks@v0.1.x/dist/livewire-turbolinks.js" data-turbolinks-eval="false" data-turbo-eval="false"></script>
<script data-turbo-eval="false">
    document.addEventListener("turbo:before-cache", () => {
        Alpine.deferMutations()
    });

    document.addEventListener("turbo:load", () => {
        Alpine.flushAndStopDeferringMutations()
    })

    document.addEventListener('turbo:before-render', () => {
        let permanents = document.querySelectorAll('[data-turbo-permanent]')
        let undos = Array.from(permanents).map(el => {
            el._x_ignore = true
            return () => {
                delete el._x_ignore
            }
        })
        document.addEventListener('turbo:render', function handler() {
            while(undos.length) undos.shift()()
            document.removeEventListener('turbo:render', handler)
        })
    })
</script>
```

I did a full writeup on this problem here: https://github.com/livewire/livewire/pull/3151#issuecomment-926020270

Reply there if you have thoughts.

Closing this - thanks!

PR #2000, opened 2021-08-31
Add period and equal  directives

snellingio:
Built a simple calculator demo last night. Ran into two obvious key issues:

`@keydown.window..=` is not valid Alpine syntax 
`@keydown.window.==` is not valid HTML 

This PR solves both issues by introducing:
1. `@keydown.window.period=`
2. `@keydown.window.equal=`


calebporzio:
Thanks! Can you add documentation for this in this PR? Thanks!

snellingio:
Added @calebporzio

calebporzio:
Thanks!

PR #2006, opened 2021-09-01
Persist to custom storage

SimoTod:
This PR adds a to modifier to allow users to persist their data in an alternative storage.

### DIFF
--- a/packages/docs/src/en/plugins/persist.md
+++ b/packages/docs/src/en/plugins/persist.md
@@ -142,11 +142,11 @@ Here's a view of Chrome Devtools to see for yourself:
 
 By default, data is saved to localStorage, it does not have an expiration time and it's kept even when the page is closed.
 
-Consider the scenario where you want to clear the data once the user close the tab. In this case you can persist data to sessionStorage using the `.to` modifier like so:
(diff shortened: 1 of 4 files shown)


packages/docs/src/en/plugins/persist.md
```diff
@@ -137,6 +137,51 @@ Here's a view of Chrome Devtools to see for yourself:
...
+            }
+            return null;
+        },
+        setItem(key, value) {
+            document.cookie = key+' = '+encodeURIComponent(value)
+        }
+    }
+</script>
+
+<div x-data="{ count: $persist(0).to(sessionStorage') }">
```
MuzafferDede, line 178: this should be `<div x-data="{ count: $persist(0).to(cookieStorage') }">`

SimoTod: Thanks, bloody cut and paste :)

calebporzio:
Thanks @SimoTod - love this.

I changed `.to` to `.using` - do you agree this is a good change?

I spent some time thinking about the implications of `.in`, `.to`, `.into`, and `.using` and believe `.using` is the most logical, reads the best, and provides the possibility for other plugins using this convention where something like `.to` and `.in` doesn't work well.

Now that I'm verbalizing, I'm just going to merge this and commit. Hopefully you like it lol! Thanks again a bunch for this.

PR #2011, opened 2021-09-02
Add $root magic property

calebporzio:


ryangjchandler:
#1477 Glad to see I wasn't the only one that need this!

PR #2016, opened 2021-09-03
Additional test for x-trap (livewire issue)

SimoTod:
This PR just contains a test for the fix to the x-trap bug affecting livewire (https://github.com/alpinejs/alpine/pull/1986).
The bug was likely caused by cloneNode starting Alpine on an element temporary detached from the DOM so it contains a function to simulate the same behaviour. As the code stops on the Alpine.clone call, it will never insert the p element.
The PR was originally applied to v3.2.4 to ensure the test was failing (see https://github.com/SimoTod/alpine/actions/runs/1199403219) cc @MuzafferDede

### DIFF
--- a/tests/cypress/integration/plugins/trap.spec.js
+++ b/tests/cypress/integration/plugins/trap.spec.js
@@ -28,3 +28,31 @@ test('can trap focus',
         get('#2').should(haveFocus())
     },
 )
+
+test('works with clone',
+    [html`
+        <script>
+            function triggerClone() {
+                var original = document.getElementById('foo')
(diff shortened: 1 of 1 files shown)


SimoTod:
just realised that the test doesn't pass with the main branch too which is weird. closing for now

MuzafferDede:
@SimoTod Great, I was able to fix your test and make it work. Should i push to your branch to make sure all good or will you create new one?

SimoTod:
You can raise a PR, I was on the train and I was trying to write a test quickly but then I had to get off so I just cancelled it and i can't have a look right now.

PR #2027, opened 2021-09-07
Improved Alpine error resiliency and logging

JacobPozaic:
- Improve Alpine resiliency by preventing Alpine from locking up when an error is reached ( which would stop other valid components from running initTree )
- Errors log the cause element and expression making debugging easier (same as V2 did)
- Errors that are thrown are always assigned the element and expression (same as V2 did)
- Add Cypress configuration to test utils to assert the element and expression are on any error/log coming from Alpine and test range of errors
- Fix store test that had reference error in it
- Fix custom directives test that had reference errors in it

# Resiliency example
Without these changes: https://codepen.io/danddanddand/pen/QWgGeZB
With these changes: https://codepen.io/danddanddand/pen/eYRBqLo


Pronian:
Hello! Does anyone know what needs to happen in order to have this merged? I really want this feature.

tests/cypress/integration/store.spec.js
```diff
@@ -70,7 +70,7 @@ test('store\'s "this" context is reactive for init function',
     [html`
         <div x-data>
         <span x-text="$store.test.count"></span>
-        <button @click="$store.test.increment()" id="button">increment</button>
+        <button id="button">increment</button>
```
calebporzio, line 73: What's the purpose for removing the @click here?

tests/cypress/integration/custom-directives.spec.js
```diff
@@ -26,7 +26,6 @@ test('directives are auto cleaned up',
     `,
     `
         Alpine.directive('foo', (el, {}, { effect, cleanup, evaluateLater }) => {
-            let evaluate = evaluateLater('foo')
```
calebporzio, line 29: What's the purpose of removing these lines?

PR #2030, opened 2021-09-07
add ShadowRoot protection for older browsers

danddanddand:
Add check to ensure ShadowRoot exists so Alpine V3 loads for older browsers (Chrome/Safari/Firefox when polyfills are present)


calebporzio:
Thanks!

PR #2039, opened 2021-09-09
Adds clarification to x-if page in docs

scottzirkel:
Currently the docs do not mention the fact that `<template>` tags can only contain one root element, specifically when used with `x-if`. I've added a small point of clarification that may help those wondering why only the first element in the tag is showing.

Added the following to the end of the file:

> Remember: `<template>` tags can only contain one root level element.


calebporzio:
Thanks for this @scottzirkel!

PR #2060, opened 2021-09-13
Bug - Components initialised twice

SimoTod:
When x-init is defined on an element without x-data and that element has child components (other x-data's. x-init's or both), those children are initialised twice. The reason is that x-init will initialise the component and all children but outNestedComponents only checks for parent components having x-data to skip the duplicated call.


calebporzio:
Thanks @SimoTod !!

Issue #402, opened 2020-04-21
Compatibility with jQuery select box plugins


pcohen12:
I've been trying to get Alpine to play nice with [Select2](https://select2.org/) and/or [Selectize](https://selectize.github.io/selectize.js/) (jQuery select box plugins), but in both cases, any Alpine event binding that's set on the `<select>` element doesn't seem to get triggered.

https://codepen.io/pcohen12/pen/VwvmboW

As far as I can tell, `change` events are correctly being triggered on the `<select>` element...my only guess is that Alpine is adding the event listeners first, and Select2/Selectize are clearing all existing listeners for some reason. Does someone smarter than me have an idea of what's going on?

HugoDF:
@pcohen12 That codesandbox has a couple of issues going.

First, you're probably better off initialising `.select2()` in `x-init` (which is an Alpine lifecycle operation) instead of `document.ready()`.

Second, in order to use x-init easily, you probably want to use `x-ref="select"` on your select that you'll `.select2()`.

Third, you probably should programmatically bind to select2 events using `$(select).on('select2:select')`, in the callback to that you can put the data into Alpine.js state and use it.

Working codesandbox to illustrate this https://codepen.io/hugodf/pen/RwWogqq

SimoTod:
Some clarifications about the original example, unfortunately it uses jQuery which doesn't trigger real native Events but it uses custom versions which work with jQuery handlers and on* attributes but don't work with native listeners. See https://github.com/jquery/jquery/issues/3347

For this reason you can't use the x-model directive. Your best shot is to use an implementation like the one suggested by Hugo or to switch to a different select library that doesn't use jquery.

pcohen12:
Thanks much, @HugoDF and @SimoTod !

glpzzz:
Hello! Having a similar issue with FancyTree. When selecting items, it updates the input field correctly, but this update doesn't reflects on the model. The other input that should have the next value is not updated. If you type in this other input, the original does gets updated. 

The change event triggered by `$(selector).val(newValues)` inside the `select` event of FancyTree is not modifying the value in the Alpine.js model. 

 I tried to apply the same principles explained for select2, but no success.

vikas-ukani:
@HugoDF You are awesome, 
Your solution works for me.
Great.

Issue #310, opened 2020-03-28
Communicate with Alpine component via event listener


booni3:
I am using the new-ish livewire method to dispatch a browser event:

```php
$this->dispatchBrowserEvent('showToaster', [
    'title' => 'Oops!',
    'message' => $error
]);
```

Within the HTML, I am initialising my Alpine toaster component and then initiating the event listener within the component method:

```html
<div x-data="toaster()" x-init="listen()">
  <div x-show="show">
    <p x-text="title"></p>
    <p x-text="message"></p>
  </div>
</div>

<script>
    function toaster() {
      return {
        show: false,
        title: '',
        message: '',
        trigger(title, message) {
          this.show = true;
          this.title = title;
          this.message = message;
        },
        listen() {
          this.$el.addEventListener('showToaster', (event) => {
            this.trigger(event.detail.title, event.detail.message);
          });
        }
      }
    }
</script>
```

I feel like I really just stumbled across this method by accident and would like to get some feedback on the best approach for this.

I found that listening for the event outside of the Alpine method would result in the event being fired and the Alpine (`trigger()`) method also being successfully called... but the dom was not effected.

HugoDF:
You can add a global listener using the `.window` or `.document` modifiers eg. `x-on:show:toaster.window` or `x-on:show-toaster.document`.

I believe your specific issue is that camelCased event names cause some odd behaviour so switching to a `:` or `-` separated event will make it work.

I've got a working codepen at https://codepen.io/hugodf/pen/qBdgyzN

Full code, note that `x-on:` needs to be on an element that is a child of the one that has `x-data`.

```html
<div x-data="{ showing: false }">
  <span x-on:show-toast.window="showing = !showing" x-on:show:toast.document="showing = !showing"></span>
  <h2>Demo</h2>
  <button x-on:click="$dispatch('show:toast')">Trigger <code>show:toast</code> on the window/document</button>
  <button x-on:click="$dispatch('show-toast')">Trigger <code>show-toast</code> on the window/document</button>
  <div role="alert" x-show="showing">
    <h3>Alert</h3>
    <p>Showing an alert triggered toggled using a global event</p>
  </div>
</div>
```

HugoDF:
@booni3 did the above solve your problem?

ryangjchandler:
@HugoDF wondering whether we should include a note in the docs about this too...

HugoDF:
[...]

Issue #264, opened 2020-03-11
I have jquery on my site and i want to use alpine js, is that possible?


jajamoks:
I have jquery on my site and i want to use alpine js, is that possible?

HugoDF:
Sure, there's nothing in Alpine that conflicts with jQuery.

the94air:
> Sure, there's nothing in Alpine that conflicts with jQuery.
@HugoDF I haven't seen this line of text for a long time. Normal DOM is the BEST!

PR #2063, opened 2021-09-14
Option to set custom listener manipulation (e. g. jQuery events compatibility) 

MartinMystikJonas:
This change would allow to enable compatibility with jQuery event system

```javascript
  Alpine.setListenerManipulators(
    (target, event, handler, options, modifiers) => { $(target).on(event, handler) },
    (target, event, handler, options) => { $(target).off(event, handler) }
  )
})
```

### DIFF
--- a/packages/alpinejs/src/alpine.js
+++ b/packages/alpinejs/src/alpine.js
@@ -5,6 +5,7 @@ import { start, addRootSelector, closestRoot, initTree } from './lifecycle'
 import { interceptor } from './interceptor'
 import { debounce } from './utils/debounce'
 import { throttle } from './utils/throttle'
+import { setListenerManipulators } from './utils/on'
 import { mutateDom } from './mutation'
 import { nextTick } from './nextTick'
 import { plugin } from './plugin'
@@ -26,6 +27,7 @@ let Alpine = {
     evaluateLater,
     setEvaluator,
     closestRoot,
+    setListenerManipulators,
     // Warning: interceptor is not public API and is subject to change without major release.
     interceptor,
     mutateDom,
--- a/packages/alpinejs/src/utils/on.js
+++ b/packages/alpinejs/src/utils/on.js
@@ -1,6 +1,22 @@
 import { debounce } from './debounce'
 import { throttle } from './throttle'
 
+let theListenerAdderFunction = nativeListenerAdder
+let theListenerRemoverFunction = nativeListenerRemover
+
+export function setListenerManipulators(listenerAdder, listenerRemover) {
+    theListenerAdderFunction = listenerAdder
+    theListenerRemoverFunction = listenerRemover
+}
+
+export function nativeListenerAdder(target, event, handler, options, modifiers) {
+    target.addEventListener(event, handler, options)
+}
+
+export function nativeListenerRemover(target, event, handler, options) {
+    target.removeEventListener(event, handler, options)
+}
+
 export default function on (el, event, modifiers, callback) {
     let listenerTarget = el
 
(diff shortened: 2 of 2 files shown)


philwolstenholme:
I think this might be better off as a third-party plugin, then people who need it can use it, and people that don't need it won't need to increase Alpine's complexity and size unnecessarily.

https://alpinejs.dev/advanced/extending
https://alpinejs.dev/advanced/extending#writing-and-sharing-plugins

MartinMystikJonas:
Example usage/test that uses this feature to make Alpine compatible with jQuery events:

```html
        <script src="https://code.jquery.com/jquery-3.6.0.slim.min.js"></script>
        <script src="//unpkg.com/alpinejs" defer></script>
        <script>
                Alpine.setListenerManipulators(
                    (target, event, handler, options, modifiers) => { $(target).on(event, handler) },
                    (target, event, handler, options) => { $(target).off(event, handler) }
                )
                $("span").trigger("click")
            })
        </script>

        <div x-data="{val: 'old'}">
            <span @click="val='new'" x-text="val"></span>
        </div>
```

calebporzio:
Thanks for contributing this. In order to keep Alpine's API surface area slim I think I'm going to reject this addition. Unless there is a large enough demand from the community - if that is the case, please re-open with references to those conversations.

MartinMystikJonas:
I fully understand need to keep API as slim as possible. But I believe that many people would like to use Alpine alongside jQuery and existing jQuery plugins. It is mostly possible but event system incompatibilities are unexpected and not easy to solve.

There are examples of some questions about Alpine and jQuery compatibility:
https://github.com/alpinejs/alpine/issues/264
https://stackoverflow.com/questions/66592950/can-i-write-both-alpine-js-and-jquery-in-same-project-will-it-cause-any-documen

For some jQuery plugins event systems incompatibility can be solved by manually adding listeners and proper reactions thet pass changes to Apline so x-model can by simulated at least partially this way. There is good example how it can be done for select2 https://codewithhugo.com/alpinejs-jquery-plugin-integration-select2/

But some people struggle to make these work because they do not know about need to made these manual event listeners to allow Apline to react to events triggered by jQuery plugins: https://codepen.io/pcohen12/pen/VwvmboW

And for many jQuery plugins this is not easily possible without creating custom fork that triggers native events instead of jQuery events.

People trying use Alpine and jQuery events was mentioned few times:
https://github.com/alpinejs/alpine/issues/310#issuecomment-617696162
https://github.com/alpinejs/alpine/discussions/1131
https://github.com/alpinejs/alpine/discussions/694#discussioncomment-48375
https://github.com/alpinejs/alpine/issues/402#issuecomment-617380302

I also discussed my problem with @SimoTod here: https://github.com/alpinejs/alpine/discussions/2062 where he suggested some workarounds but I think some universal solution would be nice to have in Alpine itself instead of hacking it for every event in every jQuery plugin used.

I hope this would be enough for you to reconsider this PR, but I understand that it might be just small part of Apline users who needs to make it fully jQuery and jQuery plugins compatible so you would reject this to prevent bloating Apline API.

MartinMystikJonas:
@calebporzio Just another idea: How about detecting existence of window.jQuery and automatically switching to jQuery event listeners if jQuery is available?

dee-ene-gee:
@MartinMystikJonas My suggestion would be to create a custom directive to handle jQuery events separately, like so:
[https://jsfiddle.net/805w19vj/1/](https://jsfiddle.net/805w19vj/1/)

Overriding the default `x-on` directive with a custom directive named `on` also seems to work, including the shorthand `@` syntax, though this would require writing your own modifier functionality:
[https://jsfiddle.net/9283twqa/1/](https://jsfiddle.net/9283twqa/1/)

I copied the structure of the default [x-on.js](https://github.com/alpinejs/alpine/blob/main/packages/alpinejs/src/directives/x-on.js) directive.

MartinMystikJonas:
@dee-ene-gee I was thinking about this first but there is significant amout of logic hidden in Alpine ```on``` implementation that would need to be reimplemented: https://github.com/alpinejs/alpine/blob/f5d2857a88c7dcf48e981e16619a8d6c3265ebcd/packages/alpinejs/src/utils/on.js

f-liva:
Thank you @dee-ene-gee your solution was perfect!

Maybe it's possible to extend the native x-model directive and supporto also this with jQuery?

ekwoka:
I don't think it's going to happen.

It's like asking for IE9 support.

PR #2064, opened 2021-09-14
Updated docs - order matters w/ Alpine.data

austenc:
When upgrading a project to v3, I got tripped up by this. Betting other folks will skip straight to the upgrade guide like I did and not realize that order matters with how v3 is initialized on the page. This is the only thing I've encountered so far! Perhaps some background about why this is deprecated would also be useful? 

Is it deprecated because it's better to have these data functions scoped within Alpine, or do you expect the old (global) functions to no longer work in a future version?

Thanks! ✌️


calebporzio:
Thanks Austen!!

PR #2072, opened 2021-09-15
refactored x-trap to use focus-trap plugin

MuzafferDede:
This PR

- replaces `wicg-inert` with `focus-trap` plugin to introduce new approach to prevent focus bouncing to address bar when focus is trapped. 

This has been mentioned 
- #2057 ,discord and a tweet replay of mine. 

P.S.
I think this approach is cheaper and performs better since it does not need to crawl the dom to add or remove attributes.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7,10 +7,6 @@
             "workspaces": [
                 "packages/*"
             ],
-            "dependencies": {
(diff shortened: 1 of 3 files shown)


calebporzio:
Thanks @MuzafferDede !

PR #2081, opened 2021-09-17
Add missing commas

xiCO2k:
Was checking the source code and found this 3 missing commas.

Hope that helps.


calebporzio:
Thanks!

PR #2084, opened 2021-09-17
Fix click.away regression

### LINKED ISSUE #2080: problem with x-show on this release 3.3.4
### Discussed in https://github.com/alpinejs/alpine/discussions/2079

<sup>Originally posted by **ravenshill** September 16, 2021</sup>
x-show is not working on 3.3.4, is ok on 3.3.3 . It shows up on very simple code.  see some snippets from my code below. Have not included the full source here.  I noticed this update included some code changes for x-show.   

 <button @click="isOpen=!isOpen; console.log('Popular clicked');  console.log(isOpen);" type="button" 
...
<div x-show="isOpen" x-cloak @click.outside="isOpen = false"
..</div>

SimoTod:
Quite a few users mentioned that 3.4.4 broke their app when using x-show + click.away.

Annoyingly, the code did pass the test (https://github.com/alpinejs/alpine/blob/main/tests/cypress/integration/directives/x-on.spec.js#L353) but the exact same example loaded on a page is broken (https://codepen.io/SimoTod/pen/OJgQmba) 🤷 

I checked the reason for the fix (navigating to different tabs) and setTimeout won't pause like requestAnimationFrame so it should be fine plus is consistent with what we do in x-show (https://github.com/alpinejs/alpine/blob/main/packages/alpinejs/src/directives/x-show.js).

Closes #2080

### DIFF
--- a/packages/alpinejs/src/directives/x-transition.js
+++ b/packages/alpinejs/src/directives/x-transition.js
@@ -125,7 +125,10 @@ function registerTransitionObject(el, setFunction, defaultValue = {}) {
 }
 
 window.Element.prototype._x_toggleAndCascadeWithTransitions = function (el, value, show, hide) {
-    let clickAwayCompatibleShow = show
+    // We are wrapping this function in a setTimeout here to prevent
+    // a race condition from happening where elements that have a
+    // @click.away always view themselves as shown on the page.
+    let clickAwayCompatibleShow = () => setTimeout(show)
 
     if (value) {
         el._x_transition


calebporzio:
Thanks @SimoTod - I felt nervous about removing that but all of the tests AND my manual tests passed. I'll try to add a test fo this case. Thanks!

calebporzio:
Tagged 3.3.5 with this fix and added a manual test for it. Thanks again!

aaronjpitts:
@SimoTod FYI this has introduced a bug: https://github.com/alpinejs/alpine/discussions/2094

SimoTod:
@aaronjpitts was it working in 3.3.3?
3.3.4 was completely breaking click.away

aaronjpitts:
@SimoTod yes, here it is working with 3.3.3: https://codepen.io/aaronjpitts/pen/WNOzBgW

SimoTod:
Pretty annoying. All 3 versions have bugs for specific use case. So the problem here is that the nested x-show using a transaction doesn't like the fix (it works okay if a transaction is also defined on the parent x-show. I'll have a look at some point today. In the meantime, you can use a previous version. Thanks for letting me know.

Tim-Wils:
@aaronjpitts @SimoTod 
Maybe this could be of help.
Hacked away Aarons 3.3.5 example into a Fiddle and wrapped the "if/else" in the walker function into https://github.com/alpinejs/alpine/blob/main/packages/alpinejs/src/directives/x-transition.js#L134  

https://jsfiddle.net/geouh8mr/ (line 1739 in the jsiddle)

Works for me in Firefox and Chrome, but don't have any testing-time left for it to further test it for performance.

SimoTod:
> @aaronjpitts @SimoTod
> Maybe this could be of help.
> It appears the child-elements should also be aware of which function should be called.
> 
> Hacked away Aarons 3.3.5 example into a Fiddle and wrapped the "if/else" in the walker function into https://github.com/alpinejs/alpine/blob/main/packages/alpinejs/src/directives/x-transition.js#L134
> 
> https://jsfiddle.net/geouh8mr/ (line 1739 in the jsiddle)
> 
> Works for me in Firefox and Chrome, but don't have any testing-time left for it to further test it for performance.

Thanks, @Tim-Wils 
Unfortunately, I think  it would still not work correctly with click away in that case (I've just updated your example to show the issue https://jsfiddle.net/389qv7yn/)

I think we need to either use setTimeout on the transactionIn call or use something with higher priority in the event loop for clickAwayCompatibleShow

PR #2096, opened 2021-09-20
Bug - Another regression 

### LINKED ISSUE #2095: 3.3.5 breaking x-show enter transitions
Hey Caleb,

3.3.5 is breaking x-show enter transitions, here is a working example of an offcanvas menu working in 3.3.4: https://codepen.io/aaronjpitts/pen/xxrWNXO

In 3.3.5 you will see it's now not using the enter transitions: https://codepen.io/aaronjpitts/pen/WNOzBjY

Cheers,
Aaron

SimoTod:
It seems that the fix for click away broke x-transition when they are defined in nested x-show.
setTimeout has a lower priority so the microtask for the transition start before the parent div gets showed and before the property change happens when the div is still hidden, the browser won't show any fancy effect. :(

I tried 'let clickAwayCompatibleShow = () => Promise.resolve().then(show)' and 'let clickAwayCompatibleShow = () => queueMicrotask(show)'. Both work correctly with the nested transition but do not work with clickAway

I tried
```javascript
    let clickAwayCompatibleShow = () => setTimeout(show)
    let nestedTransitionCompatibleShow = () => setTimeout(() => el._x_transition.in(show))

    if (value) {
        el._x_transition
            ? nestedTransitionCompatibleShow()
            : clickAwayCompatibleShow()

        return
    }
```
This would fix the issue and would be compatible with click.away but it would introduce a delay so I suspect it could break some other edge case for some users relying on the transition to start immediately.

The final version reverts it to the original implementation (requestAnimationFrame) which seemed solid and it uses setTimeout only if the tab is not visible (at that point transitions are not important anyway and we just care that everything keeps running).

Bit of an annoying one, sorry.

closes #2095

### DIFF
--- a/packages/alpinejs/src/directives/x-transition.js
+++ b/packages/alpinejs/src/directives/x-transition.js
@@ -125,10 +125,15 @@ function registerTransitionObject(el, setFunction, defaultValue = {}) {
 }
 
 window.Element.prototype._x_toggleAndCascadeWithTransitions = function (el, value, show, hide) {
-    // We are wrapping this function in a setTimeout here to prevent
+    // We are running this function after one tick to prevent
     // a race condition from happening where elements that have a
     // @click.away always view themselves as shown on the page.
-    let clickAwayCompatibleShow = () => setTimeout(show)
+    // If the tab is active, we prioritise requestAnimationFrame which plays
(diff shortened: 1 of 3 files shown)


aaronjpitts:
Yes, thank you again also @SimoTod

PR #2147, opened 2021-09-29
Fix leftover "Trap" from "Collapse" docs

pxlrbt:


calebporzio:
Thanks, I believe this has already been takin care of though. Thanks again!

PR #2165, opened 2021-10-01
Fix x-model for csp build

domtra:
There are a couple of issues with the CSP build. For one, it is not available on npm.

This PR adds a failing test for the model directive with regards to reactivity. 

The included fix replaces an evaluated string that is created in the directive with a plain function. This does not break any other test although it might not cover all the use cases the initial implementation covered.

If the CSP build gains more momentum it might be a good idea to add more specific tests for it. One question then is, how to organise these tests. Within the existing specs or in separate CSP specific ones. I included the new test in the existing spec for simplicity.

Is this something you want to consider to pursue, @calebporzio?

### DIFF
--- a/packages/alpinejs/src/directives/x-model.js
+++ b/packages/alpinejs/src/directives/x-model.js
@@ -6,8 +6,6 @@ import on from '../utils/on'
 
 directive('model', (el, { modifiers, expression }, { effect, cleanup }) => {
     let evaluate = evaluateLater(el, expression)
-    let assignmentExpression = `${expression} = rightSideOfExpression($event, ${expression})`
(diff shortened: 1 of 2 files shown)


MichaelMackus:
I agree we should add some tests to the CSP build. Perhaps this is something I could help with. Any eta on merging this @calebporzio - it would be good to be able to use x-model for CSP builds. We should at least try to be feature complete with the existing non-CSP directives IMO.

joshhanley:
Thanks for the PR! `x-model` is something that we agree should be supported in the CSP build. But we're going to close this for now as it seems to be missing the `$event` scope and the CSP build isn't actually finished yet, hence why it's not on NPM and doesn't have any tests, and as suggested above requires the other directives to be reviewed too.

PR #2171, opened 2021-10-04
Updated incorrect apostrophes in dispatch.md

Houfin64:
changed a who's (possesive) to whos and an it's (possesive) to its.


calebporzio:
Thanks!

PR #2191, opened 2021-10-07
Create global Alpine.$persist to enable persisting in Alpine.store

MalcolmKnott:
Until now it has been impossible to persist values in an `Alpine.store`

This PR extracts the `$persist` magic to a globally accessible `Alpine.$persist` to enable use cases such as:

```javascript
Alpine.store('darkMode', {
    on: Alpine.$persist(true)
})
```

We've included a test, as well as an index.html that makes it easy to play with the feature. Feel free to remove the index.html before merge. :)

@danielcoulbourne says hi.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,27 +1,117 @@
-<html>
-    <script src="./packages/intersect/dist/cdn.js" defer></script>
-    <script src="./packages/morph/dist/cdn.js" defer></script>
-    <script src="./packages/history/dist/cdn.js"></script>
-    <script src="./packages/persist/dist/cdn.js"></script>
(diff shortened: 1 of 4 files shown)


DanielCoulbourne:
This right here is a grade A pull request! Only a FOOL would neglect this stunning contribution 💯

SimoTod:
Great addition! if you want to reset the index.html you can do `git checkout master -- index.html`.
It would be good to update the docs as well, it's in package/docs/src/en/plugins/persist.md

calebporzio:
Thanks @MalcolmKnott! This will be much appreciated by the community! Great work!

PR #2213, opened 2021-10-11
x-collapse: only set overflow:hidden when strictly needed

### LINKED ISSUE #2208: x-collapse overflow hidden
### Discussed in https://github.com/alpinejs/alpine/discussions/2207

<sup>Originally posted by **reefki** October 10, 2021</sup>
When using `x-collapse` I found that it adds `overflow: hidden;` style to the element even when it on collapsed state. I know that it's really necessary to get a smooth transition but it makes a little mess with my dropdown inside that being chopped due to the `overflow: hidden;`.

I think it's better to use `overflow: initial;` when the element not expanded and change it back to `overflow: hidden;` when starting the transition and when the element collapsed.

I don't know if that's possible but let see what @calebporzio think.

Thanks</div>

georgeboot:
Currently, `x-collapse` will always set `overflow: hidden` to the element. This can give unwanted side effects.

This PR addresses possible issues by only setting `overflow: hidden` when it's strictly needed and defaulting to default value otherwise.

I tried to add a full integration test (check overflow:hidden, click button, see it's still there during transition, wait for transition to complete, check it's now gone) but I couldn't get that to work. Someone with more experience in Cypress might be willing to lend a hand here?

Fixes #2207 
Fixes #2208

### DIFF
--- a/tests/cypress/integration/plugins/collapse.spec.js
+++ b/tests/cypress/integration/plugins/collapse.spec.js
@@ -44,6 +44,6 @@ test('@click.away with x-collapse and borders (prevent race condition)',
     ({ get }) => {
         get('h1').should(haveComputedStyle('height', '0px'))
         get('button').click()
-        get('h1').should(haveAttribute('style', 'height: 0px; overflow: hidden;'))
+        get('h1').should(haveAttribute('style', 'height: auto;'))
     }
 )
--- a/tests/cypress/integration/plugins/collapse.spec.js
+++ b/tests/cypress/integration/plugins/collapse.spec.js
@@ -28,7 +28,7 @@ test('@click.away with x-collapse (prevent race condition)',
     ({ get }) => {
         get('h1').should(haveComputedStyle('height', '0px'))
         get('button').click()
-        get('h1').should(haveAttribute('style', 'height: 0px; overflow: hidden;'))
+        get('h1').should(haveAttribute('style', 'height: auto;'))
     }
 )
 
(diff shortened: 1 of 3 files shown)


tests/cypress/integration/plugins/collapse.spec.js
```diff
@@ -9,8 +9,9 @@ test('can collapse and expand element',
     `],
     ({ get }, reload) => {
         get('h1').should(haveComputedStyle('height', '0px'))
+        get('h1').should(haveAttribute('style', 'height: 0px; overflow: hidden;'))
         get('button').click()
-        get('h1').should(haveAttribute('style', 'overflow: hidden; height: auto;'))
+        get('h1').should(haveAttribute('style', 'height: auto;'))
```
SimoTod, line 14: I don't think it's correct. `overflow: hidden;` cannot be removed until the animation is complete

SimoTod: right, so that's the final state, gotcha. I think cypress is too slow at checking the state since the whole animation happens very quickly.

FYI, Alpine.transition allows a callback to run before the animation start and one to run after the animation.
It will probably be the same for cypress though but it may make more sense to set and unset the overflow property in there (el.style.overflow = 'hidden' and el.style.removeProperty('overflow'), you don't need to use Alpine.setStyle if it's a single property)
```js
Alpine.transition(el, Alpine.setStyles, {
   ...
}, () => console.log('before'), () => console.log('after'))
```
This is not official feedback, of course.

calebporzio: Yeah, testing things like this with Cypress is a nightmare, not the end of the world if we can't. Ideally we have a manual test in the manual_transitions_tests.html file (I think that's what it's called)

I need to have a deeper look at this before I merge though. Will look sometime this week.

georgeboot: @calebporzio had any time yet? :)

PR #2224, opened 2021-10-14
Fix evaluator bug when expression starts with let or const

danddanddand:
Fix bug where expressions that start with let/const were not evaluating properly. 

This is sample that wasn't evaluating properly:
```
<ul x-data="{ letters: ['a','b','c'] }">
  <template x-for="letter in letters">
    <li x-text="letter"></li>
  </template>
</ul>
```

Reported here https://github.com/alpinejs/alpine/discussions/2223


calebporzio:
Oh good catch - thanks!

PR #2231, opened 2021-10-16
Add $cookie

caneco:
Here's my suggestion of a `$cookie` plugin using some Alpine magic ✨

I know that we could change the `$persist` storage to handle cookies… but I think this it's more direct way.

The codebase it's heavily inspired from the best practices in [js-cookie](https://github.com/js-cookie/js-cookie), and with that it allows the user to set expiration, paths and domains for the cookies.

Let me know what _**y'all**_ think…

### DIFF
--- a/tests/cypress/utils.js
+++ b/tests/cypress/utils.js
@@ -58,6 +58,8 @@ function injectHtmlAndBootAlpine(cy, templateAndPotentiallyScripts, callback, pa
 
         el.evalScripts(scripts)
 
+        console.info(scripts)
+
         cy.get('[alpine-is-ready]', { timeout: 5000 }).should('be.visible');
 
         // We can't just simply reload a page from a test, because we need to
--- a/README.md
+++ b/README.md
@@ -30,6 +30,7 @@ Here's a brief look at each package in this repo:
 Package | Description
 --- | ---
 [alpinejs](packages/alpinejs) | The main Alpine repo with all of Alpine's core
+[cookie](packages/cookie) | Manage browser Cookies across page loads
 [csp](packages/csp) | A repo to provide a "CSP safe" build of Alpine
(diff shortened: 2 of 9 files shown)


caneco:
I've tried to make more tests for the plugin… but I could make cypress work locally.

localhost_60709____

Any idea of what I'm missing here?

tests/cypress/utils.js
```diff
@@ -58,6 +58,8 @@ function injectHtmlAndBootAlpine(cy, templateAndPotentiallyScripts, callback, pa
 
         el.evalScripts(scripts)
 
+        console.info(scripts)
```
SimoTod, line 61: Debug print

caneco: Whoops… I was trying to figure it out what was happening.

Thanks for the head's up!

calebporzio:
Yeah, I'm with you @caneco, I don't like the string API myself. Definitely would prefer something concrete and JS runtimey like `cookieStorage` (but not that lol).

So the question is, like you're saying @SimoTod, what's the best API.

`$persist(val).using(cookieStorage)`

Is the most consistent, but of course, unlike `sessionStorage` and `localStorage`, it's not available globally in JS.

So do we add `cookieStorage` to the window? Seems iffy introducing that globally.

Is `$cookieStorage` better? Is it worth adding that magic variable to all of Alpine when someone is using this plugin? I also don't think that's ideal.

`$persist(...).using(Alpine.cookieStorage)` seems like the best middleground, but the `Alpine.` of course isn't the absolute best thing.

I do think it might be the best we have though. Oppinions?

joshhanley:
Thanks for the PR! We've decided to hold off on this for now, until a better API can be thought up, as user can already do this in their own app by using the custom `cookieStorage` code provided in the docs https://alpinejs.dev/plugins/persist#custom-storage

PR #2262, opened 2021-10-23
Add documentation note that x-on events are case insensitive.

iUnknwn:
X-on cannot listen to events with upper case characters in their names. Added a line in the documentation to mark this and to reference why.

Info taken from the answer to this question: https://github.com/alpinejs/alpine/discussions/2215#discussioncomment-1469125


calebporzio:
Perfect. Thank you!

PR #2294, opened 2021-11-01
Make the upgrade from v2 easier

Jubeki:
AlpineJS has `$root` magic implemented in the core. Therefor it is not necessary to define a custom one.

This makes the upgrade vom v2 even smoother.

### DIFF
--- a/packages/docs/src/en/upgrade-guide.md
+++ b/packages/docs/src/en/upgrade-guide.md
@@ -55,6 +55,8 @@ Upgrading from Alpine V2 to V3 should be fairly painless. In many cases, NOTHING
 </div>
 ```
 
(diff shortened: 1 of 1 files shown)


packages/docs/src/en/upgrade-guide.md
```diff
@@ -50,30 +50,15 @@ Upgrading from Alpine V2 to V3 should be fairly painless. In many cases, NOTHING
...
-
-                return closestRootEl(node.parentNode)
-            }
-
-            return closestRootEl(el)
-        })
-    })
-</script>
-```
+For a smoother upgrade experience, you can replace all instances of `$el` with a custom magic called `$root`.
```
Jubeki, line 58:
```suggestion
```

this could be removed completely if you want to.

calebporzio:
Good catch, thanks!

PR #2309, opened 2021-11-03
Add .inert and .noscroll x-trap modifiers

calebporzio:


packages/docs/src/en/plugins/trap.md
```diff
@@ -178,3 +178,76 @@ Here is nesting in action:
...
+### .inert
+
+When building things like dialogs/modals, it's recommended to hide all the other elements on the page from screenreaders when trapping focus.
+
+By adding `.inert` to `x-trap`, when focus is trapped, all other elements on the page will receive `aria-hidden="true"` attributes, and when focus trapping is disabled, those attributes will also be removed.
+
+```alpine
+<!-- When `open` is `false`: -->
+<body>
+    <div x-trap="open" ...>
```
jonnitto, line 195: Shouldn't that be `x-trap.inert="open"`?

namjul:
In the past this used to use the `inert` attribute https://github.com/alpinejs/alpine/pull/2072/files#diff-a8ed2e650c6326125d195100753ccf415d05ffe859da586ddd57660ee0e711ffL16-L28 and now "only" `aria-hidden`.

Would be curious about the justification for that direction. My current understanding based on https://web.dev/articles/inert?hl=de is that `inert` is a better fit?

PR #2310, opened 2021-11-03
[Docs] Fix inert example in x-trap

Jubeki:
In the example you forgot to add the `.inert` modifiers to `x-trap`.
I also added an `x-data` attribute to the body element. (This is just my opinion and I can also remove it.


Jubeki:
The Tests seems to fail in `plugins/trap.spec.js` which has nothing todo with the docs.

calebporzio:
Perfect, thanks @Jubeki!

PR #2321, opened 2021-11-05
Fix x-trap test

SimoTod:
The test is failing on master because it assumes that the scrollbar width is 0 but it's not the case if the test runs on a real browser.


calebporzio:
Thanks @SimoTod - good catch!

PR #2327, opened 2021-11-06
Name lifecycle event sections after event name

Xenonym:
I found it confusing that the lifecycle event sections were named `Alpine.event` instead of just being named after the event (`alpine:event`). This PR names the sections after the event they document instead.


calebporzio:
I hear you, thanks!

PR #2330, opened 2021-11-06
Docs: Add missing .inert modifier

jonnitto:


jonnitto:
Just saw that there is already a PR: #2310

PR #2353, opened 2021-11-10
Use the hidden attribute & `display:none` to hide x-collapse content from the accessibility tree

philwolstenholme:
Big fan of Alpine! Thanks for all the time that has been put into it :) 

I was checking out [x-collapse](https://alpinejs.dev/plugins/collapse) and noticed that visually collapsed content is not hidden/removed from the accessibility tree, so keyboard users can tab to focusable elements that should be hidden, and the content may be announced to screenreader users.

At the moment we just hide content with `overflow: hidden; height: 0px;` but this isn't enough to hide content from assistive tech.

One way to hide the content safely would be to add a `hidden` attribute after the transition has ended and the content has been collapsed (or add it on load), and remove the attribute before un-collapsing the content.

The `hidden` attribute approach is particularly nice as it will play nicely with Tailwind's https://tailwindcss.com/docs/space utility. The utility won't add space around content that has been hidden with `hidden`:

```css
.space-y-1>:not([hidden])~:not([hidden]) {
    --tw-space-y-reverse: 0;
    margin-top: calc(.25rem * calc(1 - var(--tw-space-y-reverse)));
    margin-bottom: calc(.25rem * var(--tw-space-y-reverse))
}
```

`hidden` is essentially the same as `style="display:none"` except by using the `hidden` attribute we get that extra compatibility with Tailwind's space utilities 👍 

The browser user-agent stylesheet manages the hiding for us via a rule like:

```css
[hidden] { display: none; }
``` 

A downside of that is that author CSS can very easily override the `[hidden]` rule by setting a `display` property on the element being collapsed/uncollapsed, but that can be worked around by the author adding something like:

```css
[hidden] { display: none !important; }
```

This PR uses the `hidden` attribute and updates the Cypress tests for `x-collapse` to test for the presence or lack of the attribute.

### DIFF
--- a/packages/collapse/src/index.js
+++ b/packages/collapse/src/index.js
@@ -25,6 +25,8 @@ export default function (Alpine) {
 
         el._x_transition = {
             in(before = () => {}, after = () => {}) {
+                el.hidden = false;
+
                 let current = el.getBoundingClientRect().height
 
                 Alpine.setStyles(el, {
@@ -44,7 +46,6 @@ export default function (Alpine) {
                     start: { height: current+'px' },
                     end: { height: full+'px' },
                 }, () => {
-                    el.hidden = false
                     el._x_isShown = true
                 }, () => {})
             },
@@ -61,12 +62,13 @@ export default function (Alpine) {
 
                     // check if element is fully collapsed
                     if (el.style.height == `${floor}px`) {
-                        Alpine.nextTick(() => Alpine.setStyles(el, {
-                            overflow: 'hidden'
-                        }))
+                        Alpine.nextTick(() => {
+                            Alpine.setStyles(el, {
+                                overflow: 'hidden'
+                            })
+                            el.hidden = true;
+                        })
                     }
-
-                    el.hidden = true
                 })
             },
         }
--- a/packages/collapse/src/index.js
+++ b/packages/collapse/src/index.js
@@ -52,9 +52,7 @@ export default function (Alpine) {
                     during: transitionStyles,
                     start: { height: current+'px' },
                     end: { height: full+'px' },
-                }, () => {
-                    el._x_isShown = true
-                }, () => {})
+                }, () => el._x_isShown = true, () => {})
             },
 
             out(before = () => {}, after = () => {}) {
--- a/packages/collapse/src/index.js
+++ b/packages/collapse/src/index.js
@@ -6,6 +6,7 @@ export default function (Alpine) {
         if (! el._x_isShown) el.style.height = `${floor}px`
         if (! el._x_isShown) el.style.removeProperty('display')
         if (! el._x_isShown) el.style.overflow = 'hidden'
+        if (! el._x_isShown) el.hidden = true
 
         // Override the setStyles function with one that won't
         // revert updates to the height style.
@@ -42,7 +43,10 @@ export default function (Alpine) {
                     during: transitionStyles,
                     start: { height: current+'px' },
                     end: { height: full+'px' },
-                }, () => el._x_isShown = true, () => {})
+                }, () => {
+                    el.hidden = false
+                    el._x_isShown = true
+                }, () => {})
             },
 
             out(before = () => {}, after = () => {}) {
@@ -61,6 +65,8 @@ export default function (Alpine) {
                             overflow: 'hidden'
                         }))
                     }
+
+                    el.hidden = true
                 })
             },
         }
--- a/packages/collapse/src/index.js
+++ b/packages/collapse/src/index.js
@@ -5,8 +5,14 @@ export default function (Alpine) {
 
         if (! el._x_isShown) el.style.height = `${floor}px`
         if (! el._x_isShown) el.style.removeProperty('display')
-        if (! el._x_isShown) el.style.overflow = 'hidden'
+        // We use the hidden attribute for the benefit of Tailwind
+        // users as the .space utility will ignore [hidden] elements.
         if (! el._x_isShown) el.hidden = true
+        // We use display:none as the hidden attribute has very low
+        // CSS specificity and could be accidentally overriden by a
+        // user.
+        if (! el._x_isShown) el.style.display = 'none'
+        if (! el._x_isShown) el.style.overflow = 'hidden'
 
         // Override the setStyles function with one that won't
         // revert updates to the height style.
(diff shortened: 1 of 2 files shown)


philwolstenholme:
I'm now wondering if it might be worth adding the `hidden` attribute _and_ an inline `display:none;` rule. The `hidden` attribute would help out Tailwind `space` utility class users, and the inline `display:none;` rule would be useful for preventing accessibility bugs where someone has set `display: flex;` (for example) on the element being toggled… It seems like duplication of effort but I also think including both approaches makes sense… 🤔

SimoTod:
despite `hidden` looking very cool at first, it's unfortunately really weak and any styling rule would override it, as you said. x-show uses `display: none` so it may be a good idea to keep it consistent. 
Probably you can ask directly on the PR so Caleb can give you a feedback when he reviews.

packages/collapse/src/index.js
```diff
@@ -6,6 +6,7 @@ export default function (Alpine) {
         if (! el._x_isShown) el.style.height = `${floor}px`
         if (! el._x_isShown) el.style.removeProperty('display')
         if (! el._x_isShown) el.style.overflow = 'hidden'
```
philwolstenholme, line 8: I think I'll add a `display: none` rule here too, for the reasons given in the PR comments and here: https://github.com/alpinejs/alpine/pull/2353#issuecomment-965709781

packages/collapse/src/index.js
```diff
@@ -5,6 +5,13 @@ export default function (Alpine) {
 
         if (! el._x_isShown) el.style.height = `${floor}px`
         if (! el._x_isShown) el.style.removeProperty('display')
+        // We use the hidden attribute for the benefit of Tailwind
+        // users as the .space utility will ignore [hidden] elements.
+        if (! el._x_isShown) el.hidden = true
+        // We use display:none as the hidden attribute has very low
+        // CSS specificity and could be accidentally overriden by a
+        // user.
+        if (! el._x_isShown) el.style.display = 'none'
```
philwolstenholme, line 14: For these changes, it'd be more consistent with `x-show` if we only set `el.style.display = 'none'`.

I've added the `el.hidden` parts to try and be helpful to Tailwind users who might be using the [space between](https://tailwindcss.com/docs/space) utilities to add spacing between collapsed elements. Without using `el.hidden`, the CSS would add extra space around collapsed items, which would be annoying. With `el.hidden`, Tailwind will ignore collapsed items. I don't know for sure, but this might be a CSS pattern used in other codebases too?

SimoTod: Instead of reading display none, you can probably remove line 7 (which removes the style set by x-show)

philwolstenholme: Ah yes, that's smarter! I did wonder about what the `el.style.removeProperty()` call was for, but I forgot that `x-collapse` works on top of `x-show`. I've just pushed a change with your suggestion.

calebporzio:
Your points regarding the focusing nested elements are valid. The tailwind spacer utilities as well.

Because you're setting `style.display = null` here, will that conflict with a utility class like "flex"? or will it just remove it and default back to display flex?

If that's the case, I feel good about it.

Thank you as always for the feedback and help @SimoTod ❤️

philwolstenholme:
Thanks for the quick response @calebporzio !

> Because you're setting `style.display = null` here, will that conflict with a utility class like "flex"? or will it just remove it and default back to display flex?

`style.display = null` will remove the `display: none` rule from the inline `style` attribute on the element, but any other CSS rules (like one coming from a `.flex` utility class) will be unaffected, and the element will go back to being styled by the CSS cascade rather than the inline style 👍 

You mentioning utilities has reminded me that if someone was using `x-collapse` with a utility that set the `display` property (e.g. `.flex`, `.inline` etc) and that utility was `!important` (e.g. via the Tailwind [important option](https://tailwindcss.com/docs/configuration#important)) then the utility would override the inline `display: none` coming from this PR, and the site would have the same accessibility bug of being able to focus collapsed content. Many users would never notice because visually it would all look fine thanks to the `height: 0; overflow: hidden` rule.

We could force our inline `display: none` to be `!important` too by using `el.style.setProperty('display', 'none', 'important');` perhaps, but that change would need to be made in `x-show` as well. I can't think of a time when I'd want to use `x-show` or `x-collapse` and _not_ have the extra certainty that my content will _definitely_ be hidden when it's meant to be, but you never know… What do you think @calebporzio and @SimoTod ?

tests/cypress/integration/plugins/collapse.spec.js
```diff
@@ -1,19 +1,23 @@
-import { haveAttribute, haveComputedStyle, html, test } from '../../utils'
+import { haveAttribute, haveComputedStyle, html, notHaveAttribute, test } from '../../utils'
 
 test('can collapse and expand element',
     [html`
         <div x-data="{ expanded: false }">
             <button @click="expanded = ! expanded">toggle</button>
-            <h1 x-show="expanded" x-collapse>contents</h1>
+            <h1 x-show="expanded" x-collapse>contents <a href="#">focusable content</a></a></h1>
```
chrisrowe, line 7:
```suggestion
            <h1 x-show="expanded" x-collapse>contents <a href="#">focusable content</a></h1>
```

chrisrowe (changes requested):
Fix double closing link in test

SimoTod:
It won't apply to collapse and i don't know if anyone uses it that way but you can force a div to be always visible on, for instance, mobile using important and have it showable on desktop. 
But as soon as you add transitions, it will start looking weird so I doubt anyone does it. 

As far as I know, nobody complained about x-show needing important so we can maybe wait.

About hidden, I'm not sure the spacing thing makes much difference because most of the time your element will be a block. Inline blocks, when disappearing, will make the rest of the line move which is probably not desirable.

philwolstenholme:
I think maybe let's leave things as they are currently in the PR. There's a risk that `!important` utilities could undo the inline `display:none;` provided this fix, but at least then the accessibility issue is coming from the user's code (where they can fix it themselves) rather than coming from Alpine where it'll be harder for many users to fix themselves.

SimoTod:
Yeah, If people start reporting it, we can add important at a later stage. Obviously the final decision is with Caleb so let's wait for his feedback

calebporzio:
Ok, thanks for talking that through for me and for weighing in @SimoTod!

Merging this ❤️

PR #2366, opened 2021-11-12
Accessibility: display none on fully collapsed

SimonPadbury:
If the element is fully collapsed, it can still be tabbed into, and a focus ring gets lost in the hidden element. Adding `display: none` again when the element is fully collapsed will enable tabbing, etc. to skip over collapsed elements.


SimoTod:
There's already a PR open for this

SimonPadbury:
Thanks, I saw that PR later. I will close this one.

PR #2369, opened 2021-11-13
use body-scroll-lock package for more sophisticated scroll disabling

danddanddand:
Make x-trap.noscroll work on iOS

Related to this https://github.com/alpinejs/alpine/discussions/2368


calebporzio:
Thanks for this addition. Two things:

1) The tests are failing because of bad package.json formatting, should be a quick fix
2) I like this, it seems necessary, my only hesitation is the weight it adds to the x-trap plugin. Looks like it adds a full 1.3KB

image

Is all of the behavior necessary? Does it warrant the extra weight? I suppose I'm OK if it does, but it of course would be ideal to not have it as the entire plugin is only 3.3kb itself you know?

SimoTod:
@danddanddand you can ignore the last failure. See #2388

danddanddand:
I don't see much room to simplify body-scroll-lock code to get iOS to work properly https://github.com/willmcpo/body-scroll-lock/blob/master/src/bodyScrollLock.js. 

Might be able to remove options it takes in and just assume people want scroll bar gap, but probably not worth it. 

Here is iOS use case in my app which I think is typical where background shouldn't scroll: 
https://user-images.githubusercontent.com/47586871/142681712-b514728f-4482-46ca-8604-aa76b6ea3aac.mov

You can use hamburger menu in Gmail on iOS to see behaviour you'd want.

calebporzio:
Thanks for posting that @danddanddand - would you mind post an alternative of what it looks like in iOS WITHOUT this PR addition? Just so I can personally witness the problem and make sure it's worth it. Thanks!

danddanddand:
Before (can scroll bg on iOS)
https://user-images.githubusercontent.com/47586871/142735462-933c433c-3381-4ed3-880b-bd29edb46f5a.mov

After (can't scroll bg on iOS)
https://user-images.githubusercontent.com/47586871/142735500-b3de716e-0f3d-439e-9be4-08d9572d6114.MP4

calebporzio:
Thanks for posting that @danddanddand. This addition makes me hesitant for both the size and maintenance reasons that @SimoTod mentioned here: https://github.com/alpinejs/alpine/discussions/2368#discussioncomment-1672924

Think I'm going to close this one for now. Thanks for contributing though!

PR #2339, opened 2021-11-08
Fix x-transition fast double click bug

georgeboot:
Fixes #2335 

Related discussion: #2334

Bug originally introduced in #2213

@Jumaniyozov please improve your mouse clicking record with this code ;-)

PR #2370, opened 2021-11-13
Test for x-collapse double click bug

SimoTod:
Just an additional test to prevent future regressions.
The bug was already addressed by #2339
Branched off 3.5.0 to confirm that the test was failing as expected (see https://github.com/SimoTod/alpine/runs/4198418042?check_suite_focus=true#step:6:927then) then merged 3.5.1 into this PR to confirm that the test passes.


calebporzio:
Wow, thanks for adding a test like this for an already present feature @SimoTod. ❤️

PR #2376, opened 2021-11-15
Fix components instantiated twice because of complex mutations

SimoTod:
This is mostly an issue with third-party libraries such as phoenix livewire which trigger complex mutations in certain scenarios, causing x-data scope to be multiplied, duplicate event listeners, x-init to run multiple times, etc.

Basically, they trigger multiple mutations containing the same components so alpine runs initTree multiple times by mistake.

This Pr leverage the existing _x_ignore functionalities to skip any redundant processing.

Failed test: https://github.com/SimoTod/alpine/runs/4211546805?check_suite_focus=true#step:6:900
After fix: https://github.com/SimoTod/alpine/runs/4211553560?check_suite_focus=true#step:6:900

### DIFF
--- a/packages/alpinejs/src/mutation.js
+++ b/packages/alpinejs/src/mutation.js
@@ -160,12 +160,20 @@ function onMutate(mutations) {
         onAttributeAddeds.forEach(i => i(el, attrs))
     })
 
+    // Mutations are undled together by the browser but sometimes
+    // for complex cases, there may be javascript code adding a wrapper
+    // and then an alpine component as a child of that wrapper in the same
+    // function and the mutation observer will receive 2 different mutations.
+    // when it comes to run them, the dom contains both changes so the child
+    // element would be processed twice as Alpine calls initTree on
+    // both mutations. We mark all node as _x_ignored and only remove the flag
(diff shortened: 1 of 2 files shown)


calebporzio:
Thanks for another deep fix @SimoTod! ❤️

SimoTod:
hi @f-liva 
do you have an example to show what you think the regression is?
It seems to work as expected from a quick check using the latest version: https://codepen.io/SimoTod/pen/qBPRMyQ
Also, the init test instantiates the component once as expected.

f-liva:
The problem was caused by a <script></script> fragment in the content of the ajax response being injected into the page. In that fragment, the dom was manipulated again, and apparently Alpine did not notice this with the new change.

I solved it by moving that <script> fragment to the Alpine component, and everything works now. 

Happy to provide further clarification if needed.

SimoTod:
@f-liva Yeah, if you can post a quick example, so I can check the behaviour before the PR and see if we can do anything, it would be great. 
I tried myself but, by default, script tags don't get processed when injected in a page for security reasons so I need your specific setup.

PR #2388, opened 2021-11-19
Update test following #2353

SimoTod:
The test was written before #2353 so it has to consider the addition of `display: none` now


calebporzio:
Thanks!

PR #2398, opened 2021-11-21
Misc - Fix evaluatorMemo memory leak

### LINKED ISSUE #2126: evaluatorMemo cache leaks memory and DOM elements
### Discussed in https://github.com/alpinejs/alpine/discussions/2125

<sup>Originally posted by **bep** September 24, 2021</sup>
* The core of this issue lives in the `evaluatorMemo` cache: https://github.com/alpinejs/alpine/blob/4f272c579d9f9e132544096f93934f73d2b35f8a/packages/alpinejs/src/evaluator.js#L46
* That cache gets written to, but never deleted from.
* When values in that cache has references to a DOM element (as in the `x-for` example below), it effectively prevents garbage collection and leaks DOM elements and memory. 

There is a example that reproduces this at https://alpine-testlab.netlify.app/leak1/ running with the latest AlpineJS -- description repeated below:

* Open the Memory profiler in Chrome
* Click the red button, all the `x-for` components are removed.
* Click the garbage icon in the profiler to force GC.
* Click on the round button to create a new profile.
* Filer by `Detached HTMLTemplateElement`

You should now see 20 detached elements. If you click on one, it points to `evaluatorMemo`.</div>

SimoTod:
Fixes #2126 

The issue is that generateEvaluatorFromString stores a reference to the latest operation into the cached entry after the function has run. This causes leaks when the element is removed from the DOM because the cached entries keep dangling reference to alpine scopes.

`func.result` is already set to `undefined` at each execution on this line https://github.com/alpinejs/alpine/blob/main/packages/alpinejs/src/evaluator.js#L86 so setting it to `undefined` after the function has run won't change any existing behaviour. All tests are passing.

Since it's not testable in cypress, I've added a manual test for it. Feel free to checkout b2a19dcc0be9ccf8e42a9c9d908acd7bf7dea0be first to have a play.

There are other leaks to fix but I believe it's safer to just fix them one by one for each new release so we can capture bug reports from the community if there are any.

cc @bep

### DIFF
--- a/packages/alpinejs/src/evaluator.js
+++ b/packages/alpinejs/src/evaluator.js
@@ -97,11 +97,18 @@ function generateEvaluatorFromString(dataStack, expression, el) {
             if (func.finished) {
                 // Return the immediate result.
                 runIfTypeOfFunction(receiver, func.result, completeScope, params, el)
+                // Once the function has run, we clear func.result so we don't create
+                // memory leaks. func is stored in the evaluatorMemo and every time
+                // it runs, it assigns the evaluated expression to result which could
+                // potentially store reference to DOM element that will be removed
+                // later on
+                func.result = undefined
             } else {
                 // If not, return the result when the promise resolves.
                 promise.then(result => {
                     runIfTypeOfFunction(receiver, result, completeScope, params, el)
                 }).catch( error => handleError( error, el, expression ) )
+                .finally( () => func.result = undefined )
             }
         }
     }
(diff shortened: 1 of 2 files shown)


calebporzio:
@SimoTod - you're the best. Thanks so much for tackling this. The memory leak issues are tough to track down. Doing them step by step like this is a smart approach. Definitely something that needs attention for Alpine to be used in Turbo-esque contexts (and all contexts too of course)Thanks ❤️

PR #2400, opened 2021-11-23
Add id magic

calebporzio:


packages/docs/src/en/magics/id.md
```diff
@@ -0,0 +1,106 @@
+---
+order: 9
+prefix: $
+title: id
+---
+
+# $id
+
+`$id` is a magic property that can be used to generate an element's ID and ensure that it is within on the page and won't conflict with other IDs of the same name.
```
waldyrious, line 9: Hey @calebporzio, can you clarify what you mean by "it is within on the page"? I am a little confused as to what that means.

PR #2406, opened 2021-11-24
Update morph documentation

SimoTod:
The current implementation relies on Alpine being available on the window object.
Whilst that line was optional for other plugins, it's a required step for this specific one when using npm


calebporzio:
Good catch, thanks @SimoTod!

PR #2421, opened 2021-11-28
Wrong order when x-for children use x-if

### LINKED ISSUE #2210: Iteration is reversed when used with x-if
### Discussed in https://github.com/alpinejs/alpine/discussions/2209

<sup>Originally posted by **Jelenkee** October 10, 2021</sup>
https://codepen.io/kuhschubser/pen/zYdYpOY</div>

SimoTod:
Fixes #2210

If the root child of `x-for` uses `x-if`, the generated nodes were rendered at reverse because the loop pointer didn't move to the correct last item.

There was also an issue when reordering those nodes because the generated elements were not moved with the linked template tag.


calebporzio:
Thanks @SimoTod!

PR #2428, opened 2021-11-30
Remove mention of being able to access the address bar in `trap.md`

philwolstenholme:
Since #2072 the address bar can intentionally not be reached, we need to update the docs though.

While I was in this file I've also replaced usages of the word `dialogue` with `dialog`. Dialogue usually refers to speech, e.g. dialogue in a film or TV show, and dialog is the technical term used for a popup window. We can see the spelling dialog being used in ARIA (https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/dialog_role) and in HTML (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog)


calebporzio:
Good call, thanks!

PR #2434, opened 2021-12-01
Remove plugin reference from teleport doc

robdekort:
`x-teleport` was still called a plugin.

### DIFF
--- a/packages/docs/src/en/directives/teleport.md
+++ b/packages/docs/src/en/directives/teleport.md
@@ -7,7 +7,7 @@ graph_image: https://alpinejs.dev/social_teleport.jpg
 
 # Teleport Plugin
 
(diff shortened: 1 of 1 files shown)


robdekort:
Can't wait to use `x-teleport` btw. Thank you!

packages/docs/src/en/directives/teleport.md
```diff
@@ -7,7 +7,7 @@ graph_image: https://alpinejs.dev/social_teleport.jpg
 
 # Teleport Plugin
```
ajitbohra, line 8: # x-teleport

calebporzio:
Ah, good catch. Thank you!

packages/docs/src/en/directives/teleport.md
robdekort: Woops another one. How did I miss that? I'll do another PR as this one is already merged.

PR #2435, opened 2021-12-02
Explain querySelector usage in x-teleport docs

valorin:
The `x-teleport` docs reference the selector as working like `document.querySelector`, but for anyone who is unfamiliar with that function, the examples only show it working on the `<body>` element.

This modifies the examples to demo the different selector types: tag, id, and class. I also added a link to the `querySelector` docs and expanded the explination to mention the CSS selector.

Note, I haven't tested my changes to the demo code as I don't have the docs running locally. If there is an easy way to do so, let me know and I'll happily test. I couldn't see anything in the repo about how that works though.

### DIFF
--- a/packages/docs/src/en/directives/teleport.md
+++ b/packages/docs/src/en/directives/teleport.md
@@ -16,7 +16,9 @@ This is useful for things like modals (especially nesting them), where it's help
 
 By attaching `x-teleport` to a `<template>` element, you are telling Alpine to "append" that element to the provided selector.
 
-> The `x-template` selector can be any string you would normally pass into something like `document.querySelector`
+> The `x-template` selector can be any string you would normally pass into something like `document.querySelector`. It will find the first element that matches, be it a tag name (`body`), class name (`.my-class`), ID (`#my-id`), or any other valid CSS selector.
+
+[→ Read more about `document.querySelector`](https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector)
 
 Here's a contrived modal example:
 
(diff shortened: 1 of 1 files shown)


packages/docs/src/en/directives/teleport.md
```diff
@@ -105,24 +109,26 @@ Alpine does this by looking for event listeners registered on `<template x-telep
...
-            
-            <div x-data="{ open: false }">
-                <button @click="open = ! open">Toggle Nested Modal</button>
-
-                <template x-teleport="body">
-                    <div x-show="open">
-                        Nested modal contents...
-                    </div>
-                </template>
+<div class="modals">
```
lupinitylabs, line 112: At this point it should also be mentioned that if there are more elements with the `modal` class exist in the DOM, only the first one will be targeted.

lupinitylabs:
The example using body should probably not be replaced? I think body will be one of the most used targets... 🤔

calebporzio:
Thanks for this @valorin, can we keep the message at the top that you PR'd (the bit about query selector and the query selector mdn link), and revert the `x-template=".modals"` and `x-template="#modal"` stuff?

I see that that makes it easier for people to understand that you can pass in any selector, however, I think the most common (and probably wisest) usage is to use `body` for those things. So I'd rather stick with that if that's cool. Thanks!

valorin:
@calebporzio Done :-)

The first use case I thought of for this had nothing to do with `body` tags or modals, but I understand your point and I think my use-case may be a bit of an abuse of this anyway. (Moving buttons outside context without needing to handle events, components, and inheritance mess.)

valorin:
Hrm, not a clue why that's failed. I didn't touched that file. 😕

calebporzio:
This is perfect. Thanks!

PR #2436, opened 2021-12-02
Docs: Remove another plugin reference

robdekort:
Missed one! :-)


ajitbohra:
We can remove the second heading
https://github.com/alpinejs/alpine/blob/80d4ee690a50ee09faf0d0ebad28e18ab68f3f58/packages/docs/src/en/directives/teleport.md?plain=1#L15

robdekort:
This has already been fixed by now.

PR #2441, opened 2021-12-03
Bug - x-collapse flickering

SimoTod:
As highlighted by #2409, there is a sporadic bug when transitioning out an element.
This seems to be more frequent with slow devices or when the CPU is busy: after collapsing an element, sometimes the full content shows briefly and then disappears again.

This has probably been introduced by removing `overflow: hidden` when an element has been expanded. When we are about to collapse the element, the element starts without that rule which gets set by Alpine in the transitioning phase.
The cleanup callback will then remove that property so, to retain it, we set it again in the complete callback.

Because the complete callback runs before the cleanup one, the code uses nextTick to introduce a small delay: this means that there is a short period when the element is fully collapsed, still visible but without `overflow: hidden` which, I think, causes the bug described above. 

This PR removes `overflow: hidden` from the reversible rules and just toggles it on and off in sensible places so we don't need to use nextTick and the property will be set to the right value before the cleanup phase.
This PR also limit the usage of Alpine.setStyle to the point where we actually need a revert callback.

### DIFF
--- a/packages/collapse/src/index.js
+++ b/packages/collapse/src/index.js
@@ -17,11 +17,10 @@ export default function (Alpine) {
         let setFunction = (el, styles) => {
             let revertFunction = Alpine.setStyles(el, styles);
 
-           return styles.height ? () => {} : revertFunction
+            return styles.height ? () => {} : revertFunction
         }
 
         let transitionStyles = {
-            overflow: 'hidden',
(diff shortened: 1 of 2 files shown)


calebporzio:
This is perfect - I've been noticing this myself - you are a god-send @SimoTod - thank you!

PR #2443, opened 2021-12-03
Add info about modifying throttle duration

arcanemachine:
Just added some info confirming that `.throttle` can use the same duration modifier as `.debounce`. This functionality was implied by throttle's similarity to 'debounce', but now it's explicitly mentioned, avoiding any confusion.


calebporzio:
This is perfect, thank you!

PR #2464, opened 2021-12-07
feature: `x-data.isolate` modifier

ryangjchandler:
This pull request introduces a new `.isolate` modifier for the `x-data` directive. When present, it will prevent the root element's data-stack inheriting all parent components.

This came from a request in the Discord server and prior to the `x-teleport` directive being available, I would have liked something similar.

If this is something that is mergeable / wanted, I'll update the docs to include information about the modifier too.

### DIFF
--- a/packages/alpinejs/src/directives/x-data.js
+++ b/packages/alpinejs/src/directives/x-data.js
@@ -10,7 +10,7 @@ import { evaluate } from '../evaluator'
 
 addRootSelector(() => `[${prefix('data')}]`)
 
-directive('data', skipDuringClone((el, { expression }, { cleanup }) => {
+directive('data', skipDuringClone((el, { expression, modifiers }, { cleanup }) => {
     expression = expression === '' ? '{}' : expression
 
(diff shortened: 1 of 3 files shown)


calebporzio:
Interesting.

The implementation doesn't look too bad, but my gut on adding a fork in the behavior on something as crucial as the data scope in Alpine worries me.

I'm having a hard time picturing bad implications though.

Maybe can you expound on when this is useful as I've never needed it myself.

ryangjchandler:
Completely forgot about this PR. I believe we can close this PR for now, unless somebody else has a reason for it. It was originally a solution to a problem somebody on Discord had, but I believe they've found another route.

lovetingyuan:
it is useful to build a component system based on alpinejs

ekwoka:
I'm not sure it would have much use in component systems, outside of just being technically more safe.

If your component is trying to access values that don't exist in it's context, that is a problem independent of whether or not an upstream component has defined it.

But technically the safety is good.

PR #2467, opened 2021-12-08
Prevent breaking when Alpine is transpiled to ES5 (fixes #2454)

loilo:
This PR prevents the Alpine attribute evaluator from breaking when it's transpiled to ES5 by a third party bundler. The main problem in this scenario is that the `AsyncFunction` constructor can not properly be retreived. (More on this in #2454.)

I hoped to be able to fix #2454 by just wrapping the generated function's call in a Promise layer. However, it turns out that the Babel-transpiled `AsyncFunction` leads to all kinds of scope-related problems and seems to not be (easily) fixable.

The consequence I draw from this is that we should avoid using a possibly transpiled `AsyncFunction` at all.

I see two ways to achieve this:

1. Prevent the `async function() {}` that retreives the `AsyncFunction` constructor from being transpiled in the first place, via `eval`.
2. Check the `AsyncFunction` constructor name after retreiving it. If it's `"Function"`, we are in an ES5-transpiled environment and should use the regular `Function` constructor as a fallback.

Each of these approaches has its own unique drawback: (1) will fail users that transpile to ES5 on purpose because it will still use an `async` function. (2) will fail users that accidentally transpile to ES5 (e.g. via higher-level tools like Laravel Mix) but are not actually targeting ES5-only browsers and expect the `await` keyword to work in their Alpine attributes.

For this PR, I've decided to go with approach (1) because I think the set of users who transpile a library (which does not support ES5 anyway) on purpose should be negligible.

### DIFF
--- a/packages/alpinejs/src/evaluator.js
+++ b/packages/alpinejs/src/evaluator.js
@@ -51,7 +51,8 @@ function generateFunctionFromString(expression, el) {
         return evaluatorMemo[expression]
     }
 
-    let AsyncFunction = Object.getPrototypeOf(async function(){}).constructor
+    // Use `eval` to protect AsyncFunction constructor from possible third-party ES5 transpilation
+    let AsyncFunction = Object.getPrototypeOf((0, eval)('(async function(){})')).constructor
 
     // Some expressions that are useful in Alpine are not valid as the right side of an expression.
     // Here we'll detect if the expression isn't valid for an assignement and wrap it in a self-
--- a/packages/alpinejs/src/evaluator.js
+++ b/packages/alpinejs/src/evaluator.js
@@ -51,8 +51,11 @@ function generateFunctionFromString(expression, el) {
         return evaluatorMemo[expression]
     }
 
-    // Use `eval` to protect AsyncFunction constructor from possible third-party ES5 transpilation
-    let AsyncFunction = Object.getPrototypeOf((0, eval)('(async function(){})')).constructor
+    // Use `new Function` to protect AsyncFunction constructor from deconstruction
+    // through possible third-party ES5 transpilation
+    let AsyncFunction = Object.getPrototypeOf(
+        new Function('return async function() {}')()
+    ).constructor
 
     // Some expressions that are useful in Alpine are not valid as the right side of an expression.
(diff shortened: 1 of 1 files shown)


loilo:
Changed code to use `new Function` instead of `eval` as per https://github.com/alpinejs/alpine/discussions/2454#discussioncomment-1779247.

Also looking for feedback on how to weigh encapsulation vs. possible performance gains if we'd move the whole `new Function` call out of the `generateFunctionFromString` function (to only be executed once)?

calebporzio:
Mmmm definitely a valid problem. I'd be hesitant on any solution that involved either `eval` or `new Function` however.

My reasoning is partially for the "unsafe-eval" messaging around the framework as mentioned in the discussion, but more concretely around performance.

Both of those functions are slow and this would double the cost of function generation in Alpine - something i'd be extremely hesitant to do.

Is there any other way to make bundlers happy and not have any runtime evaling? Thanks for looking into this

loilo:
Thanks for the feedback @calebporzio. I couldn't find another way to appease Babel, unfortunately.

Regarding the initialization cost: Since there is only one `AsyncFunction` constructor, we could hoist the code that retreives it out of the `generateFunctionFromString` function, so it only has to be called once. (In fact, I already have another commit in this branch locally doing exactly that, but wanted to wait for initial feedback on this PR.)

As far as I can see, this would solve the performance issue, right? Or would it make the code hit another JS engine slow path that I'm not aware of?

SimoTod:
Another option would be to leave it as it is and wrap that line in a try catch so it will print a sensible error message (possibly also update the docs). At the end of the day, Alpine already required ES2017 before that PR, just it wasn't erroring until you were trying to use await in a directive and,for many the real issue is that they don't understand the error

loilo:
This was an approach I considered as well. Not sure though how that would come across, because there are a lot of devs that really don't (and understandably don't want to) understand the depths of dev tooling. I think we should probably not annoy these particular users. Just my 2 cents though.

joshhanley:
Thanks for the PR! We're going to close this for now due to the use of `new Function` as mentioned by Caleb above.

PR #2515, opened 2021-12-20
Docs: `transform` utility is now redundant in Tailwind

robdekort:
I'm not sure what Tailwind version you're running on the docs so I kept them in the Verbatim section.


calebporzio:
MUCH nicer. Thanks!

PR #2524, opened 2021-12-23
Bug - Morph text nodes.

SimoTod:
patchChildren uses `firstChild` to select the first child node which works correctly with DOM elements and text nodes but then it uses `nextElementSibling` to get the next node which only returns DOM elements. As a consequence, if the children list is a mix of DOM elements and text nodes (for example, a paragraph content with `<br>` tags), any text node that is not the first child will be ignored and left in the previous status.

Flagged by #2463

Failing test: https://github.com/alpinejs/alpine/runs/4619366189?check_suite_focus=true#step:6:1106

### DIFF
--- a/packages/morph/src/morph.js
+++ b/packages/morph/src/morph.js
@@ -263,8 +263,8 @@ async function patchChildren(from, to) {
         // Patch elements
         await patch(currentFrom, currentTo)
 
-        currentTo = currentTo && dom(currentTo).next()
-        currentFrom = currentFrom && dom(currentFrom).next()
+        currentTo = currentTo && dom(currentTo).nodes().next()
+        currentFrom = currentFrom && dom(currentFrom).nodes().next()
     }
 
     // Cleanup extra froms
@@ -358,8 +358,8 @@ class DomManager {
     }
 
     traversals = {
-        'first': 'firstChild',
-        'next': 'nextSibling',
+        'first': 'firstElementChild',
+        'next': 'nextElementSibling',
         'parent': 'parentElement',
     }
 
--- a/packages/morph/src/morph.js
+++ b/packages/morph/src/morph.js
@@ -6,23 +6,23 @@ function breakpoint(message) {
     if (! debug) return
 
     message && logger(message.replace('\n', '\\n'))
-   
+
     return new Promise(resolve => resolveStep = () => resolve())
 }
 
 export async function morph(from, toHtml, options) {
     assignOptions(options)
-    
+
     let toEl = createElement(toHtml)
 
     // If there is no x-data on the element we're morphing,
     // let's seed it with the outer Alpine scope on the page.
     if (window.Alpine && ! from._x_dataStack) {
         toEl._x_dataStack = window.Alpine.closestDataStack(from)
-        
+
         toEl._x_dataStack && window.Alpine.clone(from, toEl)
     }
-    
+
     await breakpoint()
 
     patch(from, toEl)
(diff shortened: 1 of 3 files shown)


packages/morph/src/morph.js
```diff
@@ -358,16 +358,16 @@ class DomManager {
     }
 
     traversals = {
-        'first': 'firstElementChild',
-        'next': 'nextElementSibling',
+        'first': 'firstChild',
+        'next': 'nextSibling',
```
SimoTod, line 362: This one is the relevant change

packages/morph/src/morph.js
```diff
@@ -222,8 +222,8 @@ async function patchChildren(from, to) {
                 domKeyHoldovers[domKey] = currentFrom
                 currentFrom = addNodeBefore(currentTo, currentFrom)
                 domKeyHoldovers[domKey].remove()
-                currentFrom = dom(currentFrom).nodes.next()
-                currentTo = dom(currentTo).nodes.next()
+                currentFrom = dom(currentFrom).nodes().next()
+                currentTo = dom(currentTo).nodes().next()
```
SimoTod, line 226: Not relevant to this bug but I believe it's a typo as nodes should be a function call

carlmjohnson:
Totally dumb question, but I just learned about https://developer.mozilla.org/en-US/docs/Web/API/Document/createTreeWalker. Would that be useful for the traversals or is it not relevant?

SimoTod:
I don't know, i never used it. it would change the original implementation quite a lot so you need to ask Caleb. Zeal

calebporzio:
Great question @carlmjohnson!

When I cracked into this problem I started there, but there are two problems with TreeWalker:
A) Oddly enough, t's actually slower than manually crawling in almost every case. I just recently checked it's performance again, and still, slower.
B) That's enough to not use it, but more than that, to my knowledge, there is no way to "skip" a subtree with the TreeWalker and that is required functionality here.

packages/morph/src/morph.js
calebporzio: Good call, should be yeah

packages/morph/src/morph.js
calebporzio: So, I think you're right in identifying the problem, however, this might not be the place. Sorry for the not-amazingly-readable code here, but here's my intent with these traversals:

By default all the traversals should operate on full Element nodes, unless you chain on the `.nodes()` method, in which case, all the traversal methods will be replaced with Node specific ones.

So if there's a place in the morph codebase that uses Node traversal in one place and Element in another, we should fix it upstream by adding `.nodes()` or subtracting `.nodes()`

Does that make sense?

Let me know if I'm the one with the miss-understanding and as always thank so much for your very high quality work on Alpine ❤️

SimoTod: I think I've initially tried removing the nodes() method but some tests started failing because it wasn't teleporting the nodes anymore or something like that. I need to have a look again. 👍

packages/morph/src/morph.js
```diff
@@ -263,8 +263,8 @@ async function patchChildren(from, to) {
         // Patch elements
         await patch(currentFrom, currentTo)
 
-        currentTo = currentTo && dom(currentTo).next()
-        currentFrom = currentFrom && dom(currentFrom).next()
+        currentTo = currentTo && dom(currentTo).nodes().next()
+        currentFrom = currentFrom && dom(currentFrom).nodes().next()
```
SimoTod, line 267: Relevant change. When getting the next element for the next loop iteration, we want to use nodes() so we don’t skip text nodes.

packages/morph/src/morph.js
SimoTod: Gotcha. Yeah, that makes sense now. I've updated the PR to fix the issue in the way you suggested.

calebporzio:
Thanks @SimoTod!

PR #2525, opened 2021-12-24
Change wrong directive name in docs

SuPythony:
Changed `x-template` to `x-teleport` as `x-template` is not a directive and the line talks about `x-teleport`'s selector.


calebporzio:
Yep! Thanks!

PR #2528, opened 2021-12-25
Fix failing tests - account for the morph function now returning a promise

VicGUTT:
The morph tests do not account for the morph function now being asynchronous. This PR is an attempt to remedy that.
That said, this PR seemed to have revealed a false positive in the test [`non-keyed elements are replaced instead of moved`](https://github.com/alpinejs/alpine/blob/aa2f5f2530b2c80755ff911f51889f1eb79089bb/tests/jest/morph/internal.spec.js#L14). I'll try and figure out what's going on at a later date.

### DIFF
--- a/packages/morph/src/morph.js
+++ b/packages/morph/src/morph.js
@@ -57,7 +57,7 @@ function assignOptions(options = {}) {
     adding = options.adding || noop
     added = options.added || noop
     key = options.key || defaultGetKey
-    lookahead = options.lookahead || true 
+    lookahead = options.lookahead || false
     debug = options.debug || false
 }
 
@@ -190,12 +190,12 @@ async function patchChildren(from, to) {
             if (toKey && domKeyHoldovers[toKey]) {
                 let holdover = domKeyHoldovers[toKey]
 
-                dom.append(from, holdover)
+                dom(from).append(holdover)
                 currentFrom = holdover
 
                 await breakpoint('Add element (from key)')
             } else {
-                let added = addNodeTo(currentTo, from)
+                let added = addNodeTo(currentTo, from) || {}
 
                 await breakpoint('Add element: ' + added.outerHTML || added.nodeValue)
 
@@ -320,7 +320,7 @@ function addNodeTo(node, parent) {
         return clone
     }
 
-    return parent;
+    return null;
 }
 
 function addNodeBefore(node, beforeMe) {
--- a/packages/morph/src/morph.js
+++ b/packages/morph/src/morph.js
@@ -25,7 +25,7 @@ export async function morph(from, toHtml, options) {
     
     await breakpoint()
 
-    patch(from, toEl)
+    await patch(from, toEl)
 
     return from
 }
@@ -319,6 +319,8 @@ function addNodeTo(node, parent) {
 
         return clone
     }
+
+    return parent;
 }
 
 function addNodeBefore(node, beforeMe) {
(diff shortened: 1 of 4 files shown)


tests/jest/morph/internal.spec.js
```diff
@@ -1,50 +1,50 @@
...
     dom.querySelector('span').is_me = true
 
-    morph(dom, '<div><span>bar</span></div>')
+    await morph(dom, '<div><span>bar</span></div>')
 
     expect(dom.querySelector('span').is_me).toBeTruthy()
 })
 
-test('non-keyed elements are replaced instead of moved', () => {
+test('non-keyed elements are replaced instead of moved', async () => {
```
VicGUTT, line 14: TODO:

- Figure out the intent
- Figure out what went wrong where

packages/morph/src/morph.js
```diff
@@ -319,6 +319,8 @@ function addNodeTo(node, parent) {
 
         return clone
     }
+
+    return parent;
```
SimoTod, line 323: I'm not sure it should return parent sincr that is not an added node and it would make the debugger print thr wrong statement. It should probably return null or another falsy value and the caller should handle it properly.

tests/jest/morph/internal.spec.js
SimoTod: The intent is to demostrate that the first list item does not move. The test sets a custom property on the li tag then it triggers a mutation where the text content of the first list item changes and a second list item is added. Note, the new dom element looks like the first one before the mutation but it's actually a brand new node as the custom property is still set on the first item. If the node was keyed, the mutation would move the first node to position due and prepend a new node so the custom property would be on the second list item.

VicGUTT:
@SimoTod Thanks for the feedback. I have definitely not forgotten my PRs. This new year started with lot of life changing events on my end. That said, I should be back on this somewhere this week or at most early next week :v:

calebporzio:
Thanks @VicGUTT! And thanks for the feedback on this @SimoTod

packages/morph/src/morph.js
VicGUTT: Yes, agreed, good catch. I've returned `null` from the function instead of `parent` and made the following changes to [line 198](https://github.com/alpinejs/alpine/pull/2528/files/02653339f78115c2a5f731c476e2e110d6c6c998#diff-16dbe79737c2ee37a02e810f6102eac94023aa7204051573a1463b3a846f5824R198) :

```diff
- let added = addNodeTo(currentTo, from)
+ let added = addNodeTo(currentTo, from) || {}
```

The variable `added` is only used for the debugger; and I'm assuming having a message of `Add element: undefined` is tolerable.

tests/jest/morph/internal.spec.js
VicGUTT: Oh no I did get what the code was doing, my question was more like "_Why is this the expected behavior in the first place_". But I admit my phrasing wasn't the best, and thanks for the thorough answer :v:.

Now I figured this behavior is wanted when the `lookahead` option is set to `false`, otherwise the expected behavior is the opposite, meaning : "_non-keyed elements are moved instead of replaced_".

So my question now would go to @calebporzio directly because I'm confused as to what the default boolean value for the `lookahead` option should be.

The [docs](https://github.com/alpinejs/alpine/blob/main/packages/docs/src/en/plugins/morph.md?plain=1#L199) sais it should be `false` but it was changed to `true` in [#d07aeee](https://github.com/alpinejs/alpine/commit/d07aeee#diff-16dbe79737c2ee37a02e810f6102eac94023aa7204051573a1463b3a846f5824R60).

SimoTod: Yep, it's already reverted to false on the main branch. 👍

calebporzio:
Thanks!

PR #2535, opened 2021-12-27
Change interactive visualization button text color (Morph plugin docs)

SuPythony:
The button text is not visible so changing it's color to black.


calebporzio:
Thanks! Turns out this is actually caused by us changing `bg-brand` to `bg-aqua-X`

Thanks for catching this!

PR #2544, opened 2021-12-29
Bug - morph lookahead and keys

SimoTod:
- The lookahead value was defaulted to true instead of false as per docs
- lookahead was just checking the first next sibling 
- There was a typo in patchChildren when using keys


calebporzio:
Great PR, thanks for identifying those issues.

My only comment is changing the default value to `true` instead of `false`.

I believe I intentionally set it to `false`, but I forget exactly why.

Some potential reasons I may have done that:
- to keep parity with morphdom (morphdom doesn't have any lookahead)
- to save on speed if people don't want that behavior

Not exactly sure why it's not the default though as it's useful in many situations.

What do you think? If you feel good about it, I'll merge this.

Thanks a ton @SimoTod!

calebporzio:
Right right, good call. Thanks @SimoTod !

PR #2545, opened 2021-12-29
Remove .only from x-on.spec

sztadii:
I found that we are not running all the tests, because of the `test.only`.
I think it was added during development, but by mistake, it is merged to the main branch.

Before:
Screenshot 2021-12-29 at 17 59 08

After:
Screenshot 2021-12-29 at 18 02 00

PS. I was thinking to add eslint to prevent this kind of issue, but if you will agree to it I will add it in the next PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ node_modules
 scratch.md
 dist/
 .env.json
+.idea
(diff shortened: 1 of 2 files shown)


calebporzio:
Thanks! Hate it when that happens

PR #2546, opened 2021-12-30
Add resize plugin

marekalgoud:
Hello,

As I had the usecase, I propose this simple plugin inspired from the intersected plugin but for the resize Observer.
I had a debounce function to improve performance (not really sure about what I did with the "evaluateLater" function, but it works).

I hope it will help some developpers !

Thanks for your amazing work.

Bests wishes for the next year !

Marek

### DIFF
--- a/packages/docs/src/en/plugins/resize.md
+++ b/packages/docs/src/en/plugins/resize.md
@@ -1,0 +1,73 @@
+---
+order: 6
+title: Intersect
+description: An Alpine convenience wrapper for Resize Observer that allows you to easily react when an element is resized.
+
+---
+
+# Resize Plugin
+
+Alpine's Resize plugin is a convenience wrapper for [Resize Observer](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) that allows you to easily react when an element is resized.
+
+This is useful when you want to trigger js script depending of the size of an element.
+
+<a name="installation"></a>
+## Installation
+
+You can use this plugin by either including it from a `<script>` tag or installing it via NPM:
+
+### Via CDN
+
+You can include the CDN build of this plugin as a `<script>` tag, just make sure to include it BEFORE Alpine's core JS file.
+
+```alpine
+<!-- Alpine Plugins -->
+<script defer src="https://unpkg.com/@alpinejs/resize@3.x.x/dist/cdn.min.js"></script>
+
+<!-- Alpine Core -->
+<script defer src="https://unpkg.com/resize@3.x.x/dist/cdn.min.js"></script>
+```
+
+### Via NPM
+
+You can install Resize from NPM for use inside your bundle like so:
+
+```shell
+npm install @alpinejs/resize
+```
+
+Then initialize it from your bundle:
+
+```js
+import Alpine from 'alpinejs'
+import intersect from '@alpinejs/resize'
+
+Alpine.plugin(intersect)
+
+...
+```
+
+<a name="x-resize"></a>
+## x-resize
+
+The primary API for using this plugin is `x-resize`. You can add `x-resize` to any element within an Alpine component, and when that component is resized, the provided expression will execute.
+
+For example, in the following snippet, `shown` will remain `false` until the element is scrolled into view. At that point, the expression will execute and `shown` will become `true`:
+
+```alpine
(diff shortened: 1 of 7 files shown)


packages/docs/src/en/plugins/resize.md
```diff
@@ -0,0 +1,73 @@
...
+
+...
+```
+
+<a name="x-resize"></a>
+## x-resize
+
+The primary API for using this plugin is `x-resize`. You can add `x-resize` to any element within an Alpine component, and when that component is resized, the provided expression will execute.
+
+For example, in the following snippet, `shown` will remain `false` until the element is scrolled into view. At that point, the expression will execute and `shown` will become `true`:
```
SimoTod, line 55: This paragraph needs updating, I think

packages/docs/src/en/plugins/resize.md
```diff
@@ -0,0 +1,73 @@
...
+
+```alpine
+<div x-data="{ width: 0 }" x-resize="width = $el.innerWidth">
```
SimoTod, line 58: The example needs updating: innerWidth is a property of the window object and with this callback, it will always return 'undefined'. You can use clientWidth instead (https://developer.mozilla.org/en-US/docs/Web/API/Element/clientWidth)

packages/docs/src/en/plugins/resize.md
```diff
@@ -0,0 +1,73 @@
...
+
+
+
+<a name="x-resize-debounce"></a>
+### x-resize:200
+
+You can debounce the resize function execution, so the code below will be executed just one time, after 200ms
+
+```alpine
+<div x-resize:200="console.log($el.innerWidth)">...</div>
```
SimoTod, line 71: I think it would be less confusing if you use a modifier to make it consistent with other directives (https://alpinejs.dev/directives/on#debounce).

P.s. The example needs updating too: same as above

packages/resize/src/index.js
```diff
@@ -0,0 +1,24 @@
+const debounce = (callback, wait) => {
+    let timeout = null
+    return (...args) => {
+        clearTimeout(timeout)
+        timeout = setTimeout(() => {
+            callback.apply(null, args)
+        }, wait)
+    }
+}
```
SimoTod, line 9: You don't need this part, you can just use `Alpine.debounce` at line 13 to save bytes.

packages/docs/src/en/plugins/resize.md
```diff
@@ -0,0 +1,73 @@
...
+```alpine
+<div x-data="{ width: 0 }" x-resize="width = $el.innerWidth">
+    Element width <span x-html="width"></span>
+</div>
+```
+
+
+
+<a name="x-resize-debounce"></a>
+### x-resize:200
```
SimoTod, line 66: The anchor needs to match the title or it won't work.

SimoTod (commented):
A couple of typos.

Maybe a window modifier would be useful as well for cases where the element has a fixed size but needs to react when the window is resized

packages/docs/src/en/plugins/resize.md
marekalgoud: My bad, it's ok know.

packages/docs/src/en/plugins/resize.md
marekalgoud: I changed the example for getBoundingClientRect()

packages/docs/src/en/plugins/resize.md
marekalgoud: You're right, I made changes to be more consitent.

packages/resize/src/index.js
marekalgoud: Done :)

packages/docs/src/en/plugins/resize.md
marekalgoud: I've fixed it :)

marekalgoud:
Thank you Simone for your quick feedback. I made some changes. Have a nice day !

calebporzio:
Thanks for being willing to contribute! Unfortunately, in the interest of keeping Alpine's API as minimal as possible and keeping the maintenance burden to a minimum, I'm going to hold off for now on this. Thanks again!

PR #2550, opened 2021-12-31
New event specific for plugin initialization

ethan-ho:
This change is to introduce an extra event fired prior to `alpine:init` for initializing plugin, namely `alpine:plugin-init`. This is based on discussion in #2465.

Currently, plugins are initialized in `alpine:init` cycle which is same as what user would normally do for their own code, and this could result in timing issue that plugins aren't initialized yet before user accessing plugin in their code

There are ways to mitigate this issue but not ideal, e.g. enforcing code loading order: plugins → user's code → alpine library. Introducing an extra event fired before `alpine:init`.

### DIFF
--- a/packages/alpinejs/src/lifecycle.js
+++ b/packages/alpinejs/src/lifecycle.js
@@ -8,6 +8,7 @@ import { warn } from './utils/warn'
 export function start() {
     if (! document.body) warn('Unable to initialize. Trying to load Alpine before `<body>` is available. Did you forget to add `defer` in Alpine\'s `<script>` tag?')
 
+    dispatch(document, 'alpine:plugin-init')
     dispatch(document, 'alpine:init')
     dispatch(document, 'alpine:initializing')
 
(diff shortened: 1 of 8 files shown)


joshhanley:
Just putting this here as a small summary of the linked discussion and the problem it solves:

The original error was `TypeError: Alpine.$persist is not a function` when trying to use `$persist` in a custom alpine:init listener

```html
<script>
      Alpine.store('darkMode', {
        on: Alpine.$persist(false).as('darkMode_on'),
        //on: false,
        toggle() {
          this.on = !this.on
        }
      })
    })
</script>
```

joshhanley:
We're going to leave this PR open for now, as Caleb wants to review it in detail, but currently doesn't have time whilst work on Livewire V3.

ekwoka:
There are actually already 2 different initializing events that fire. I believe `alpine:init` and `alpine:initialize`.

PR #2556, opened 2022-01-03
Fix x-if sub expressions still being evaluated after x-if evals to false

kylethielk:
Hi There!

This PR is to address an issue where after an `<template x-if="expression"></template>` has evaluated to false and thus the DOM element has been removed, sub expressions are still evaluated. Example reports of this issue are: #2540 #2296 #1837 #1917

Imagine the following code:

```
<div x-data="{user: {name: 'lebowski'}}">
  <button @click="user=null">Logout</logout>
  <template x-if="user">
    <span x-text="user.name"></span>
  </template>
</div>
```

Before the suggested fix the above code would produce the following when "Logout" is clicked:

Screen Shot 2022-01-03 at 5 03 05 PM

This is because of the following pseudo timeline:

1. `user` is set to `null`
2. This causes `effect()` on the `x-text` directive to queue  an expression evaluation i.e `user.name`.
3. The `x-if="user"` expression is evaluated and the result is now false. The cleanup tasks run and removes the `<span></span>` element from the DOM.
4. The queue runs and evaluates the expression from Step 2. At this point user is null and thus `user.name` throws an exception.

My solution is to simply remove the queued effect from the queue in the `el._x_undoIf()` method call. I am however not sure how maintainers will feel about exposing a new method in `scheduler.js` though. I could not determine a better way though.

This was my first peek into the internals of Alpine so there very well may be a better way to solve this issue. I added a new Cypress test to demonstrate this bug and to also demonstrate how this solution fixes the bug.

Very open to feedback on a better solution.

### DIFF
--- a/packages/alpinejs/src/directives/x-if.js
+++ b/packages/alpinejs/src/directives/x-if.js
@@ -3,6 +3,8 @@ import { addScopeToNode } from '../scope'
 import { directive } from '../directives'
 import { initTree } from '../lifecycle'
 import { mutateDom } from '../mutation'
+import { walk } from "../utils/walk"
+import {dequeueJob} from '../scheduler'
 
 directive('if', (el, { expression }, { effect, cleanup }) => {
     let evaluate = evaluateLater(el, expression)
@@ -22,7 +24,18 @@ directive('if', (el, { expression }, { effect, cleanup }) => {
 
         el._x_currentIfEl = clone
 
-        el._x_undoIf = () => { clone.remove(); delete el._x_currentIfEl }
+        el._x_undoIf = () => {
+            walk(clone, (node) => {
+                if (!!node._x_effects) {
+                    node._x_effects.forEach((effect) => {
+                        dequeueJob(effect)
+                    })
+                }
+            })
+            clone.remove();
+            delete el._x_currentIfEl
+            
+        }
 
         return clone
     }
--- a/packages/alpinejs/src/scheduler.js
+++ b/packages/alpinejs/src/scheduler.js
@@ -10,6 +10,13 @@ function queueJob(job) {
 
     queueFlush()
 }
+export function dequeueJob(job)
+{
+    var index = queue.indexOf(job);
+    if (index !== -1) {
+        queue.splice(index, 1);
+    }
+}
 
 function queueFlush() {
     if (! flushing && ! flushPending) {
--- a/tests/cypress/integration/directives/x-if.spec.js
+++ b/tests/cypress/integration/directives/x-if.spec.js
@@ -51,3 +51,26 @@ test('x-if initializes after being added to the DOM to allow x-ref to work',
         get('li').should(haveText('bar'))
     }
 )
+
+//If x-if evaluates to false, the expectation is that no sub-expressions will be evaluated.
+test('x-if removed dom does not evaluate reactive expressions in dom tree',
+    html`
+    <div x-data="{user: {name: 'lebowski'}}">
+        <button @click="user = null">Log out</button>
+        <template x-if="user">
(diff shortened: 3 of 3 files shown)


kylethielk:
I've pushed a new commit [2dc402b](https://github.com/alpinejs/alpine/pull/2556/commits/2dc402b7515e353524c9b98eab29367e7f029722) to extend this fix for `x-for` as well thanks to comment by @SimoTod [here](https://github.com/alpinejs/alpine/discussions/2540#discussioncomment-1904735).

packages/alpinejs/src/directives/x-for.js
```diff
@@ -6,6 +6,7 @@ import { initTree } from '../lifecycle'
 import { mutateDom } from '../mutation'
 import { flushJobs } from '../scheduler'
 import { warn } from '../utils/warn'
+import {dequeueJob} from '../scheduler'
```
markjaquith, line 9: Spacing.

packages/alpinejs/src/directives/x-for.js
```diff
@@ -135,6 +136,12 @@ function loop(el, iteratorNames, evaluateItems, evaluateKey) {
         for (let i = 0; i < removes.length; i++) {
             let key = removes[i]
 
+            //Remove any queued effects that might run after the DOM node has been removed.
+            if (!!lookup[key]._x_effects) {
+                lookup[key]._x_effects.forEach((effect) => {
+                    dequeueJob(effect)
+                })
```
markjaquith, line 143: Could just call `lookup[key]._x_effects.forEach(dequeueJob)` without the arrow function.

packages/alpinejs/src/directives/x-if.js
```diff
@@ -3,6 +3,8 @@ import { addScopeToNode } from '../scope'
 import { directive } from '../directives'
 import { initTree } from '../lifecycle'
 import { mutateDom } from '../mutation'
+import { walk } from "../utils/walk"
+import {dequeueJob} from '../scheduler'
```
markjaquith, line 7: Spacing.

packages/alpinejs/src/directives/x-if.js
```diff
@@ -22,7 +24,18 @@ directive('if', (el, { expression }, { effect, cleanup }) => {
 
         el._x_currentIfEl = clone
 
-        el._x_undoIf = () => { clone.remove(); delete el._x_currentIfEl }
+        el._x_undoIf = () => {
+            walk(clone, (node) => {
+                if (!!node._x_effects) {
+                    node._x_effects.forEach((effect) => {
+                        dequeueJob(effect)
+                    })
```
markjaquith, line 32: Could just call `node._x_effects.forEach(dequeueJob)` without the arrow function.

packages/alpinejs/src/scheduler.js
```diff
@@ -10,6 +10,13 @@ function queueJob(job) {
 
     queueFlush()
 }
+export function dequeueJob(job)
+{
+    var index = queue.indexOf(job);
+    if (index !== -1) {
+        queue.splice(index, 1);
+    }
+}
```
markjaquith, line 19: Opening brace on same line, `const` instead of `var`, omit semicolons.

markjaquith (changes requested):
This is working great, and fixes a seriously annoying bug. Nicely done. 👏🏻  Ran my own tests, and this definitely fixes the issue.

All I have is a bunch of boring code formatting consistency advice.

kylethielk:
Thanks @markjaquith . Pushed recommended changes.

calebporzio:
This looks great. Thanks for contributing!

PR #1933, opened 2021-08-10
feature: Value/Modifier based intersect

MuzafferDede:
This PR

- Adds `:enter` `:leave` value directives to `x-intersect`.
- No breaking changes. `x-intersecet="[expression]"` still works as expected. 
- Improves existing `x-intersect` tests. (old ones had issues)
- It was very cheap to achieve such a high demanding behavior. 

## Directives

| Directive      | Description |
| ---            | --- |
| `:enter`       | Applied only entering phase. |
| `:leave`       | Applied only leaving phase. |

### :enter

```html
<div x-intersect:enter="shown = true">...</div>
```

### :leave

```html
<div x-intersect:leave="shown = false">...</div>
```

### both

```html
<div  x-intersect:enter="shown = true" x-intersect:leave="shown = false">...</div>
```

PR #2557, opened 2022-01-05
Fix x-intersect reliability

markjaquith:
`x-intersect` doesn't currently reliably work.

See #1833, #1871.

I can verify those experiences. Sometimes you can scroll an element into view, and the expression doesn't fire.

Next, #1933 added `x-intersect:leave` (yay) and `x-intersect:enter`, which was an implied alias (though the docs didn't acknowledge this explicitly) with one key difference: it works.

The problem is that the `IntersectionObserver` handler bails on the following three conditions:

```js
     ! entry.isIntersecting && value === 'enter'
    || entry.isIntersecting && value === 'leave'
    || entry.intersectionRatio === 0 && ! value
```

The first two are fine. The last one is no good. `intersectionRatio` is a tricky beast, and receiving a `0` doesn't mean that the element isn't visible. Since Alpine is supplying a `threshold` of `0`, for slow scrolls or small elements, the browser may send a `IntersectionObserverEntry` with an `intersectionRatio` of `0`.

It won't send another, so this is the cause of `x-intersect` being unreliable.

This PR makes `x-intersect` a true alias of `x-intersect:enter`, and ensures that it reliably works when the tracked item enters the viewport.

It no longer relies on the untrustworthy `intersectionRatio`.

I also updated the docs to make it clear that `x-intersect:enter` is an alias, for parity with `:leave` (and it also matches `x-transition`).

### DIFF
--- a/packages/docs/src/en/plugins/intersect.md
+++ b/packages/docs/src/en/plugins/intersect.md
@@ -78,16 +78,18 @@ For example, in the following snippet, `shown` will remain `false` until the ele
 <a name="x-intersect-enter"></a>
 ### x-intersect:enter
 
-You can opt to only trigger x-intersect when the element ENTERS the viewport by adding the `:enter` suffix to `x-intersect` like so:
+The `:enter` suffix is an alias of `x-intersect`, and works the same way:
 
 ```alpine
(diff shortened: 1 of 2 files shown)


calebporzio:
Thanks!

PR #2558, opened 2022-01-05
Add rootMargin support to x-intersect with .margin

markjaquith:
`x-intersect` has limited utility, because it can only fire as an element is actually entering the viewport. There are many more cases where people would like to do something like load an image or a video player or an ad not when the element is on the screen, but when it's **almost** on the screen. Or think of an infinite scroll situation where you want to start loading more content before the user hits the bottom. And there are other cases where people would like to do something before something leaves the screen, say when it's "within 50px of the edge".

`IntersectionObserver` supports an option called `rootMargin` which allows you to tweak the margins of the viewport and do all of the above.

---
✨✨ `x-intersect.margin.200px="almostVisible = true"` ✨ ✨
---
✨✨ `x-intersect:leave.margin.5%="almostGone = true"` ✨ ✨
---

This PR adds that ability to `x-intersect` via `x-intersect.margin`.

* `x-intersect.margin.200px` — expands in all directions
* `x-intersect.margin.-200px` — contracts in all directions
* `x-intersect.margin.200px.100px` — expands 200px top/bottom, 200px left/right
* `x-intersect.margin.400px.300px.200px.100px` — expands 400px top, 300px right, 200px bottom, 100px left
* `x-intersect.margin.10%.20%` — expands 10% top/bottom, and 20% left/right

Percentages and pixels are the only two units supported by `IntersectionObserver`. I went with the bare number representing pixels, because I think this is the unit most people will expect. Negative numbers are supported (used to "push" the triggering boundary inward).

I also named the modifier `.margin` instead of `.rootMargin` or `.root-margin` because it's a single word that makes intuitive sense to me, and it makes the API prettier.

The values map cleanly to the value you'd pass to `rootMargin` (e.g. `400px 300px 200px 100px`) with the only difference being "dots" instead of spaces, and the ability to omit `px`, the same way you can omit `ms` in Alpine's timing modifiers.

The multiple trailing modifiers thing is something I hadn't seen yet in Alpine, but it seems like it fits the style. Most people will be using a single margin value, but it could be nice to have full control, and the implementation code is a simple `for` loop.

Note that the Cypress test I added won't pass until #2557 is merged to fix `x-intersect`'s unreliable triggering.

fixes #2247
fixes #2394
fixes #2510


Sulaymon-dev:
Like 😍😍😍

calebporzio:
This is really well done. Thanks!

robdekort:
Wow this is fantastic!

PR #2560, opened 2022-01-06
Crashfix on try Alpine(store, null)

shatilov-diman:
Alpine.store('userId', null)

cdn.js:308 Uncaught TypeError: Cannot convert undefined or null to object
    at Function.getOwnPropertyDescriptors (<anonymous>)
    at recurse (cdn.js:308)
    at initInterceptors (cdn.js:321)
    at Object.store (cdn.js:1190)
    at <anonymous>:1:8


calebporzio:
Closing this because no test has been provided. By providing a test, we can ensure this fix is fixed forever and this bug isn't re-introduced. Please re-submit after adding a test. Thanks!

PR #2561, opened 2022-01-06
Extend store with unstore function

shatilov-diman:
I believe it will be nice to remove something from the store.
But I'm not sure if it fit to the concept.


calebporzio:
Thanks for being willing to contribute, however I don't see the value this API adds and this doesn't include tests or docs so I'm closing.

PR #2586, opened 2022-01-17
Replace ... with closing tag character

ravage84:
The ... didn't really make it clearer and did not add anything beneficial to the docs.


calebporzio:
Good call - thank you

PR #2603, opened 2022-01-22
Add noreturn modifier to x-trap

AndrewMonty:
There may be scenarios where you don't want to return the previous focus state after `x-trap` evaluates to false.

Consider a dropdown that is triggered by focusing a search input. If focus is returned to the input after closing the dropdown, it would just trigger the dropdown to open again.

A new modifier `x-trap.noreturn` would prevent this behavior.


calebporzio:
Wow, this is the perfect PR. Great use case, presented well, docs and tests included, abides by coding style. Thank you for going the extra mile ❤️

PR #2668, opened 2022-02-13
Update debounce.js

sacarvy:
Hope that this implementation, is going to help.

Although My first contribution, 🤗


calebporzio:
Thanks for this contribution! However, I don't see the value it adds and it diverges from some of the coding style of Alpine - so unfortunately I'm going to close this. But don't give up on contributions! Thanks!

PR #2682, opened 2022-02-17
Focus trap doesn't release focus when the element is removed from the dom

SimoTod:
If the focus trap is active and the element is removed from the dom, focus doesn't get released and tabbing around the page becomes impossible. This happens in Alpine if the focus is combined to x-if.
```html
    <button @click="open = true">open</button>
    <template x-if="open">
        <div x-trap="open">
            <button @click="open = false">close</button>
        </div>
    </template>
</div>
 ```

Before https://github.com/alpinejs/alpine/pull/2556/files this was coincidentally working because the effect wasn't released correctly but it doesn't work anymore in v3.9.0

### DIFF
--- a/packages/focus/src/index.js
+++ b/packages/focus/src/index.js
@@ -3,7 +3,7 @@ import { focusable, tabbable, isFocusable } from 'tabbable'
 
 export default function (Alpine) {
     let lastFocused
-    let currentFocused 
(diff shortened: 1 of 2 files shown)


calebporzio:
Perfect, thanks @SimoTod !!

PR #2701, opened 2022-02-26
Docs - fix x-teleport modal example

mjdiloreto:
Came across this inconsistency on the docs page, where the rendered example does not match the example code.

 Not _very_ confusing... but a little confusing.

Screen Shot 2022-02-26 at 10 42 52 AM


calebporzio:
Thanks!

PR #2711, opened 2022-03-01
Removed non breaking space

tschallacka:
The non breaking space gives issues in webpack where it may be transformed into
 
```javascript
l.setAttribute('style', cache ||Â '')
```

image


calebporzio:
Weird, thanks though!

PR #2712, opened 2022-03-02
Fix morhdom sibling cleanup bug

calebporzio:


tests/cypress/integration/plugins/morph.spec.js
```diff
@@ -255,3 +255,34 @@ test('can morph text nodes',
         get('h2').should(haveHtml('Foo <br> Baz'))
     },
 )
+
+test.only('can morph with added element before and siblings are different',
```
VicGUTT, line 259: @calebporzio This is probably not intentional _(referring to the `.only`)_, right ? ^^

Issue #2693, opened 2022-02-22
destroy() does not have access to the current scope.


hirasso:
### Discussed in https://github.com/alpinejs/alpine/discussions/2103

<div type='discussions-op-text'>

<sup>Originally posted by **bep** September 21, 2021</sup>
So, custom `data` components can have a `destroy()` function, which is great.

I would expect to use this for _cleaning_, e.g. helping the Garbage Collector do its job. But the `this` does not seem to be the same as the `data` `this`.</div>

Opening an issue for this, since it would open the door to a lot of great functionality to be able to access the scope from inside the `destroy()` directive.

AugustMiller:
A brilliant user in the Discord server helped me with a similar question, yesterday—hopefully this will help unblock someone else who is interested in this "cleanup" process!

The key here is to avoid the "implicit return" factory function that the docs provide, as it starves you of space to do "setup" logic, or to store “non-state” data:

```js
Alpine.data('my-component', () => ({ // Implicitly returned object, not a function body!
  init () {
    this.interval = window.setInterval(() => {
      console.log('Tick!');
    });
  },
  destroy () {
    window.clearInterval(this.interval); // Uh oh!
  },
}));

Instead, open a regular function body:

```js
Alpine.data('my-component', () => {
  // Stuff available across all methods, and to destroy():
  let interval;
  let myObject;

  // Component definition, same as the implicit return from the first example:
  return {
    init () {
      myObject = new SomeBigScaryObject();

      interval = window.setInterval(() => {
        console.log('Tick!');
      }, 1000);
    },
    destroy () {
      window.clearInterval(interval);
      myObject.someDeInitMethod();
    },
  };
});
```

🌳

AugustMiller:
Looks like this behavior was revised in #2733!

PR #2717, opened 2022-03-03
Documentation for `destroy()` and component cleanup

AugustMiller:
Hi!

This PR adds two new sections to the [`Alpine.data()` documentation page](https://alpinejs.dev/globals/alpine-data) that seek to explain these concepts:

1. The unadvertised `destroy()` component method;
2. How to track local, non-reactive data;

Both of these took some deeper exploration of the source (and issue history, discussions) to understand, and felt like they deserved clarification. In particular, the apparent preference for the terse “implicit return” syntax in the component definition syntax was a subtly distracting pattern—one that had me fully convinced that there was no strategy for holding non-reactive data that might be required to mount third-party libraries (say, a MapboxGL instance), or to memoize scalar values (like an interval ID) that need manual cleanup at the end of a component's lifecycle.

I do understand that blurring Alpine's reactivity and local scope could result in an undue support burden—but I think the documentation is already plenty explicit about object properties and methods’ relationship to the reactivity engine.

Hope this tracks, and that the writing is stylistically acceptable!

## Related Issues and Discussions:
- #2692 
- #2693 
- #2103 
- #775

### DIFF
--- a/packages/docs/src/en/globals/alpine-data.md
+++ b/packages/docs/src/en/globals/alpine-data.md
@@ -123,7 +123,7 @@ Alpine.data('blinker', () => {
 })
 ```
 
-> Note the difference in the first line of this component definition—we're opening a function body, rather than the arrow function’s [implicit return](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#function_body) feature, as in previous examples. Your component's final definition still takes the same shape, but must be explicitly returned by the factory function.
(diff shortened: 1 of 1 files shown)


SimoTod:
If #2733 gets merged, this PR will need updating

joshhanley:
Thanks for the PR! We're going to close this for now as `destroy()` is undocumented due not being as reliable as we'd like and that there should also probably be an `x-destroy` directive with comprehensive tests around all of it.

PR #2721, opened 2022-03-04
Ensure `Alpine.closestDataStack` exists

joshhanley:
When implementing the Morph plugin in Livewire, Livewire's tests were using an old version of Alpine, as such were getting the below error.

Screen Shot 2022-03-04 at 9 01 23 pm

This PR ensures `Alpine.closestDataStack` exists, before using in the Morph plugin, as it wasn't introduced until 3.5.2.

Didn't add a test for it, as wasn't sure how to run a test using an older version of Alpine in Cypress.

Hope this helps!

### DIFF
--- a/packages/morph/src/morph.js
+++ b/packages/morph/src/morph.js
@@ -23,7 +23,7 @@ export async function morph(from, toHtml, options) {
 
     // If there is no x-data on the element we're morphing,
     // let's seed it with the outer Alpine scope on the page.
-    if (window.Alpine && ! from._x_dataStack) {
+    if (window.Alpine && window.Alpine.closestDataStack && ! from._x_dataStack) {
         toEl._x_dataStack = window.Alpine.closestDataStack(from)
 
         toEl._x_dataStack && window.Alpine.clone(from, toEl)


SimoTod:
Morph was released in v3.5.2 so it's expected. In general, official plugins and Alpine should be on the same version to guarantee full compatibility. Any reason why you can't just update Alpine? This Pr will still require you to use a different plugin version, you would update the plugin instead of updating alpine.
Otherwise, people could argue that every plugin should be backwards compatible with v3.0 (mostly impossible because they use features that are introduced in later versions).

joshhanley:
@SimoTod oh yeah agreed! People should install the same plugin version as Alpine version in their own apps.

But I'm currently replacing Morphdom in Livewire's core with Morph. So while it requires the `@alpinejs/morph` package, we're only making using of the Morph function, not the plugin itself.

Due to that though, we have no control over which version of Alpine is installed in an app, so we have to make sure it runs regardless of which Alpine version a user may have installed.

I actually ran into it because Livewire's tests use Alpine V2 and an old V3 version (3.0.6) 😅

This is the draft PR, if you're interested https://github.com/livewire/livewire/pull/4758

SimoTod:
Ah, I see. In this case, given it's an internal thing, it's probably fine 😂

joshhanley:
Haha ok sweet! 😁

calebporzio:
Good calls here - thanks @SimoTod and @joshhanley

joshhanley:
No worries!

PR #2725, opened 2022-03-05
Fix Morph missing next sibling when replacing element

joshhanley:
When morphing inline elements, Morph is currently adding a space before the "!" in the test.

With this as the starting from and to elements:

image

Running morph results in this:

image

This PR adds a failing test for this use case.

It also removes the `.only` left on the previous morph testcase.

I have some ideas on why this is happening so will have a play with it.

Hope this helps!

### DIFF
--- a/packages/morph/src/morph.js
+++ b/packages/morph/src/morph.js
@@ -207,7 +207,7 @@ async function patchChildren(from, to) {
             } else {
                 let added = addNodeTo(currentTo, from) || {}
 
-                await breakpoint('Add element: ' + added.outerHTML || added.nodeValue)
+                await breakpoint('Add element: ' + (added.outerHTML || added.nodeValue))
 
                 currentTo = dom(currentTo).nodes().next()
 
--- a/packages/morph/src/morph.js
+++ b/packages/morph/src/morph.js
@@ -278,11 +278,14 @@ async function patchChildren(from, to) {
             }
         }
 
+        // Get next from sibling before patching in case the node is replaced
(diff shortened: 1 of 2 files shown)


joshhanley:
Turns out that this was due to `currentTo.nextSibling` returning `null` after the element had been patched. In this case, it was because the `<span>` was replaced with `<b>`, as such `currentTo` didn't have any reference to what the next sibling was anymore.

Have pushed up a fix to this PR.

Hope this helps!

packages/morph/src/morph.js
```diff
@@ -207,7 +207,7 @@ async function patchChildren(from, to) {
             } else {
                 let added = addNodeTo(currentTo, from) || {}
 
-                await breakpoint('Add element: ' + added.outerHTML || added.nodeValue)
+                await breakpoint('Add element: ' + (added.outerHTML || added.nodeValue))
```
joshhanley, line 210: For text nodes, the nodeValue wasn't outputting, so had to add brackets for it to output properly

tests/cypress/integration/plugins/morph.spec.js
```diff
@@ -256,7 +256,7 @@ test('can morph text nodes',
     },
 )
 
-test.only('can morph with added element before and siblings are different',
+test('can morph with added element before and siblings are different',
```
joshhanley, line 259: Removed the `.only` so all tests would run

joshhanley:
No worries!

PR #2731, opened 2022-03-05
Fix trouble combining keyup event, key name and modificator "once"

lozunoff:
Hello! This PR  fix trouble combining keyup event, key name and modificator "once". Issue: https://github.com/alpinejs/alpine/issues/2732


lozunoff:
Added test

PR #2733, opened 2022-03-05
Feature/destroy scope

SimoTod:
This is not properly a bug but a lot of people are expecting to have access to the current scope in the destroy callback, similarly to what happens in init so they can do stuff like
```html
<div x-data="{
  foo: null,
  init() {foo = setInterval(() => console.log('tick'), 1000)},
  destroy() {clearInterval(foo)}
}">
```

Note, there are workarounds such as storing data on the global object or inside the function scope when using Alpine.data but they are not obvious ones.

The feature is undocumented so I don't think it's a bug but more of a nicety. Feel free to close it if you think it's not important or there is a specific design choice behind it.

Failing test applied to the current build: https://github.com/SimoTod/alpine/runs/5433954162?check_suite_focus=true#step:6:273


calebporzio:
This makes total sense - thanks @SimoTod!

PR #2743, opened 2022-03-10
Fix x-transition documentation

agenordebriat:
I noticed an error in the `x-transition` directive documentation: the `.duration` modifier currently handles both `enter` and `leave` durations.

### DIFF
--- a/packages/docs/src/en/directives/transition.md
+++ b/packages/docs/src/en/directives/transition.md
@@ -52,9 +52,9 @@ You can configure the duration you want for a transition with the `.duration` mo
 <div ... x-transition.duration.500ms>
 ```
 
-The above `<div>` will transition for 500 milliseconds when entering, and 250 milliseconds when leaving.
+The above `<div>` will transition for 500 milliseconds when entering, and 500 milliseconds when leaving.
 
-This difference in duration generally desirable default. If you wish to customize the durations specifically for entering and leaving, you can do that like so:
+If you wish to customize the durations specifically for entering and leaving, you can do that like so:
 
 ```alpine
 <div ...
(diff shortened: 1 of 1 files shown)


calebporzio:
Thanks for the contribution, however I believe that the actual transition time is half the duration for "leave" by default, so that may not be wrong - or am I missing something?

agenordebriat:
You're absolutely right!
I've updated my PR so the documentation is as clear as possible.

packages/docs/src/en/directives/transition.md
```diff
@@ -46,15 +46,17 @@ You can override these defaults with modifiers attached to `x-transition`. Let's
...
+Initially, the duration is set to be 150 milliseconds when entering, and 75 milliseconds when leaving.
+
 You can configure the duration you want for a transition with the `.duration` modifier:
 
 ```alpine
 <div ... x-transition.duration.500ms>
 ```
 
-The above `<div>` will transition for 500 milliseconds when entering, and 250 milliseconds when leaving.
+The above `<div>` will transition for 500 milliseconds when entering, and 500 milliseconds when leaving.
```
calebporzio, line 57: Sorry, am I wrong in think that this shouldn't be changed to 500? Please clarify - thanks!

agenordebriat: I found out that as soon as you specify a duration, it takes over for both entering and leaving.

You can see what I'm talking about in [this CodePen](https://codepen.io/agenordebriat/pen/vYpXEjo).

In this example, I bumped the duration to `5000ms` instead of the `500ms` on the doc and you can clearly see it obviously takes `5000ms` to appear, but also `5000ms` to disappear and not `2500ms` as currently stated in the doc.

The halved duration only exists for the default value, when you don't specify any duration.

calebporzio: Ok great, thanks!

PR #2751, opened 2022-03-15
Swap mainfields in build script

ekwoka:
This reduces the build size of the AlpineJS module by ~20%

AlpineJS
Before: 112kb
After: 87kb

Focus Plugin:
Before: 32kb
After: 28kb

I've used a variation of this build myself on projects for a while now and have yet to see any actual issues in browser or with bundlers.

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -59,7 +59,7 @@ function bundleFile(package, file) {
                 outfile: `packages/${package}/dist/${file.replace('.js', '.esm.js')}`,
                 bundle: true,
                 platform: 'neutral',
-                mainFields: ['main', 'module'],
+                mainFields: ['module', 'main'],
             })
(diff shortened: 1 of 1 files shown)


calebporzio:
Interesting, here are my result before and after this change (there is no build size difference)
Before: 
image

After: 
image

Can you clarify what you're going for here and why it makes the bundle smaller? Of course I always love making the bundle smaller!

ekwoka:
@calebporzio 

Yes, that is because your build size report reports on the minified CDN size.

Not the esm module.

I should have clarified, but these sizes reported are the size of the output `esm.js`.

So those importing the project without using the CDN would be affected.

Naturally, the sizes shown are exaggerated from the esm not being minified, but it's still extra stuff.

PR #2766, opened 2022-03-17
X-html in x-for

SimoTod:
Currently, for newly created elements, the initTree callback is inside a mutate dom call which means that directives injected by `x-html` are not picked up by the mutation observer. It shouldn't be a big issue to move initTree outside of mutateDOm as we do for refreshScope when dealing with elements that are shuffled around.

### DIFF
--- a/packages/alpinejs/src/directives/x-for.js
+++ b/packages/alpinejs/src/directives/x-for.js
@@ -187,9 +187,9 @@ function loop(el, iteratorNames, evaluateItems, evaluateKey) {
 
             mutateDom(() => {
                 lastEl.after(clone)
-            })
 
-            initTree(clone)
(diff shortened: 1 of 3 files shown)


leobm:
ok , cool, the fix looks really simple. What are the implications of the fix? Why was the init for cloned nodes previously in mutateDom?

SimoTod:
From my tests, it doesn't seem to have any undesired side effects. I don't know if there is a reason why it was built that way but Caleb will know when it gets time to review it.

SimoTod:
Thanks for the feedback, sounds good to me. I'll update the PR.

calebporzio:
Thanks @SimoTod! I'm much more comfortable merging this. Thanks again

PR #2792, opened 2022-03-28
Introduce x-model modifier to preference text over value

byjpr:
Adds two new modifiers `.text` and `.prefer-text` for use with Select and Options tags.

### Current behaviour
If you have the following code:

```html
<select name="user[role]" x-model="select">
  <option value="role1" selected="">Admin</option>
  <option value="role2">User</option>
</select>
```

The text inside `<span x-text="select"></span>` would be `role1`.

However, 

```html
<select name="user[role]" x-model="select">
  <option selected="">Admin</option>
  <option>User</option>
</select>
```

The text inside `<span x-text="select"></span>` would be `Admin`.

#### Using .text modifier

```html
<select name="user[role]" x-model.text="select">
  <option value="role1" selected="">Admin</option>
  <option value="role2">User</option>
</select>
```

The text inside `<span x-text="select"></span>` would be `Admin`.

#### Using .prefer-text modifier
The text inside `<span x-text="select"></span>` would be `Admin` provided an inner was present. Failing that it would fall back to value.

### DIFF
--- a/packages/alpinejs/src/directives/x-model.js
+++ b/packages/alpinejs/src/directives/x-model.js
@@ -100,14 +100,10 @@ function generateAssignmentFunction(el, modifiers, expression) {
                         return safeParseNumber(rawValue)
                     })
                     : Array.from(event.target.selectedOptions).map(option => {
-                        if(modifiers.includes('text')) {
-                            return option.text
-                        } else if (modifiers.includes('prefer-text')) {
-                            return option.text || option.value
-                        } else {
-                            return option.value || option.text
-                        }
+                        return option.value || option.text
                     })
+            } else if (el.tagName.toLowerCase() === 'select' && modifiers.includes('text')) {
+                return event.target.options[event.target.selectedIndex].text || event.target.value
             } else {
                 let rawValue = event.target.value
                 return modifiers.includes('number')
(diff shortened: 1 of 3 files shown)


byjpr:
After spending some time thinking, it makes no real sense to fall back on value. If you had a `select` that had [no inner text it would be empty](https://codepen.io/byjpr/pen/JjMJxKE).

- [ ] Abandoned falling back to value
- [ ] Proposition is fully coded
- [ ] Removed the pseudocode that I put together yesterday which was correctly pointed out in the discussion to be in the wrong area.
- [ ] Docs have been created
- [ ] Test is passing

calebporzio:
The feature makes sense, I'm just wondering if it's something we want in core - is there significant demand for this?

byjpr:
> The feature makes sense, I'm just wondering if it's something we want in core - is there significant demand for this?
I don't have any research on the demand for this. Mainly put it together for me. Wanted to push it upstream :)

> This isn't a perfect solution, but you can always set the "value" to the same as the innerText of an element right?
Depends on where you're sending the content and if you have control of the server, or if the server can accept human-readable content as an input. E.G: you might have a country list that you displayed as ISO 3166 country names, but the server accepts ISO 3166-1 country codes.

PR #2801, opened 2022-03-31
Add missing x-data at x-for docs

sertxudev:
The x-for range docs are missing the empty x-data.

I was testing some functionality and the x-for wasn't working, I checked the docs and copy-pasted the example, wasn't working either.

Finally had to check the test to find out why it wasn't working.
https://github.com/alpinejs/alpine/blob/3b152ced8bd57ced34cbf5bc544942c4f804ad28/tests/cypress/integration/directives/x-for.spec.js#L393-L402

Awesome work by the way Caleb.

### DIFF
--- a/packages/docs/src/en/directives/for.md
+++ b/packages/docs/src/en/directives/for.md
@@ -77,7 +77,7 @@ You can also access the index inside a dynamic `:key` expression.
 If you need to simply loop `n` number of times, rather than iterate through an array, Alpine offers a short syntax.
 
 ```alpine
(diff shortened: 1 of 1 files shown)


calebporzio:
Yes, thanks @SimoTod. Good to know that people get tripped up on this. We opt to not include x-data in every single code snippet in the docs. I'd be open to another PR that explains this situation somewhere else better? Or maybe enough people disagree with me and in that case you can re-open this and we can merge it. Thanks for contributing!

PR #2824, opened 2022-04-11
Fix link in warning

hebbet:
The link in the warnings goes to `alpine.dev`, which doesn't resolve.

i guess it should be `alpinejs.dev`


hebbet:
was introduced with #2815

Atem18:
Can confirm, it should be alpinejs.dev

calebporzio:
sry, this has already been fixed, thanks though!

PR #2834, opened 2022-04-13
add `x-ref` binding example

browner12:
based on the current documentation I was trying to setup an `x-ref` via a binding with 

```js
trigger: {
    ['x-ref']() {
        return 'trigger';
    },
},
```

which does not work.

After some trial and error I got it to work with 

```js
trigger: {
    ['x-ref']: 'trigger',
},
```

so I figured I'd add it to the docs so no one needs to struggle like I did 😄


calebporzio:
thanks!

PR #2841, opened 2022-04-15
Add Promise to $nextTick

ekwoka:
To bring the functionality more inline with [Vue's implementation](https://v2.vuejs.org/v2/api/?redirect=true#Vue-nextTick) of `$nextTick`, the magic returns a Promise that resolves when the callback is run.

This allows pausing your own functions until after pending DOM updates without needing to handle your own scheduling.

Shouldn't break anything since nobody should be assigning the output of `$nextTick` to anything anyway.

Side Story:
I was having issues with the reactivity related to an array between two components where the last update to the array would not be reflected in the DOM (and one or two other similar things) and I dove into the Vue reactivity docs and they mentioned using `$nextTick` when there are rendering issues, and I tried it and it worked for Alpine, even without the await 🤷, but I was curious about the implementation and potential benefits of enabling a very Alpiney options for waiting for DOM updates.

### DIFF
--- a/packages/alpinejs/src/nextTick.js
+++ b/packages/alpinejs/src/nextTick.js
@@ -3,14 +3,19 @@ let tickStack = []
 
 let isHolding = false
 
-export function nextTick(callback) {
-    tickStack.push(callback)
-
-    queueMicrotask(() => {
-        isHolding || setTimeout(() => {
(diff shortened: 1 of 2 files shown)


calebporzio:
I like this, Thanks!

PR #2845, opened 2022-04-17
Added forward-slash named keyboard event

Tom-Julux:
This PR adds a named key to `packages/alpinejs/utils/on.js`.

Analog to `.` (Period) and `=` (Equal), html attributes may also not contain the `/` ((Forward-Slash) character.


calebporzio:
thanks for the PR, this actually already exists as "slash", just isn't documented, so I modified the PR. Thanks!

PR #2848, opened 2022-04-18
Update x-for documentation

csfrancis:
Ref: https://github.com/alpinejs/alpine/discussions/2830

Updates `x-for` documentation to state that keys must be unique.


browner12:
Could you say 

> It is important to specify unique keys for each `x-for` iteration...

instead?

calebporzio:
good call, thanks!

PR #2870, opened 2022-04-25
change "Alpine.store(...)" link to /globals/alpine-store

mhuxain:
currently Alpine.store(...) links to /magics/store, which is counter intuitive, hence change the like to the actual  Alpine.store(...) docs at  /globals/alpine-store


calebporzio:
Thanks!

PR #2887, opened 2022-04-27
add description of new packages

browner12:
- descriptions where taken from the documentation
- also alphabetized list
- left out `portal` because I'm not sure what it is


calebporzio:
Thanks for doing this! Good catch!

PR #2922, opened 2022-05-11
Remove space between logical not and variable for docs

jmmarco:
Hello there 👋 
When I go over the docs for Alpine.js I find expressions that make use of the logical not operator have a space between it and the corresponding variable. Like this:
```
this.open = ! this.open
```
It sounds a bit silly, but I found myself re-reading the expression to make sure I understood what I saw. Since the the [standard convention is to have no spaces](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_NOT), perhaps we could apply it here as well:
```
this.open = !this.open
```

Most the codebase files make use of this particular styling as well. It works, but it looks a bit odd.
I figured we could make it clear for the documentation examples 🙂 

If this gets approved maybe we could do the same for the codebase (separate PR). Let me know what you think.

### DIFF
--- a/packages/docs/src/en/directives/bind.md
+++ b/packages/docs/src/en/directives/bind.md
@@ -33,7 +33,7 @@ Here's a simple example of a simple dropdown toggle, but instead of using `x-sho
 
 ```alpine
 <div x-data="{ open: false }">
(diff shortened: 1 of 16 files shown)


calebporzio:
Thanks, but gonna hold off on this for now

PR #2945, opened 2022-05-20
Allow disabling x-mask with a `falsy` value

stepanjakl:
Hi,

I have a specific use case for the Mask plugin, where the user selects an international phone code and the input value gets formatted accordingly.

Screenshot 2022-05-20 at 12 22 01

Screenshot 2022-05-20 at 12 22 19

However, the user should also have an option to type the number without being forced the format (in a case, the format is not available in the list). Example:

Screenshot 2022-05-20 at 12 27 33

In the end, the example code looks like this:

```
x-mask:dynamic="selectValue ? `${options[selectValue].international} ${options[selectValue].format}` : false"

// Above translates to:
x-mask:dynamic="'cz' ? `+420 999 999 999` : false" --> +420 987 654 321

// Or:
x-mask:dynamic="null ? `` : false" --> no forced formatting
```

And that's what this PR allows for. Let me know if anything needs changing.

### DIFF
--- a/packages/mask/src/index.js
+++ b/packages/mask/src/index.js
@@ -48,6 +48,9 @@ export default function (Alpine) {
 
             let template = templateFn(input)
 
+            // If a template value is `falsy`, then don't process the input value
+            if(!template) return false
+
             // If they hit backspace, don't process input.
             if (lastInputValue.length - el.value.length === 1) {
                 return lastInputValue = el.value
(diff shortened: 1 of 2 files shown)


calebporzio:
This makes a lot of sense - would you mind adding a test for it?

calebporzio:
This is great, thanks!

PR #2946, opened 2022-05-20
Adds autofocus awareness to the focus plugin

aidan-casey:
Currently, when using `x-trap`, the first child element in the tab order is focused per the `focus-trap` package defaults. This pull request adds a more sensible default by focusing the first element that has the `autofocus` attribute. If none is found, the first child element in the tab order is focused.

I could see this being especially helpful in the context of modals and drawers.


calebporzio:
Thanks!

jannisborgers:
Just wanted to leave a heart. This saved me a ton of work and is super convenient! 👏

PR #2957, opened 2022-05-25
expose stopObservingMutations and startObservingMutations

jimsproull:
expose these two APIs globally


jimsproull:
Closing as we realized we were mutating data after render (which I guess is a no-no!). We now only render the data once and no longer need these APIs exposed.

Issue #2803, opened 2022-03-31
Inconsistent update with template x-if


cjloong:
### Discussed in https://github.com/alpinejs/alpine/discussions/2802

<div type='discussions-op-text'>

<sup>Originally posted by **cjloong** April  1, 2022</sup>
Hi there,
I have a page structured as follows:
section1 (x-if="bar")
- input x-model="foo"

section2
- part 1 (x-if="!bar")
- part 2 (appears in section 1 and 2)
- part 3 (x-if="!bar")

The code to control the above is quite complicated. We discovered an intermittent issue where section 2, part 1 sometimes does not appear. Same as part 3.

I have manage to condense the issue into a codepen as follows:
https://codepen.io/cjloong/pen/zYpzQGr

Click on the "Simulation Error" button to reproduce the error.

</div>

Foxbud:
I believe I'm experiencing a related issue with `x-show`. It almost seems like `x-show` breaks when its dependent value toggles too quickly.

In this self-contained example, clicking `Enable` under the `Unexpected Behavior` header results in both the `Enable` and `Disable` buttons being displayed despite them having mutually exclusive display conditions. Then the `Expected Behavior` header demonstrates how adding a slight delay seems to fix the issue.

```html
<html>
	<head>
		<script src="https://cdnjs.cloudflare.com/ajax/libs/alpinejs/3.10.2/cdn.min.js" defer></script>
	</head>
	<body>
		<h1>Unexpected Behavior</h1>
		<div x-data="{state: false}"
			 x-init="$watch('state', val => {
						if (val) {state = false}
					})">
			<button type="button"
					x-show="state"
					@click="state = false">
				Disable
			</button>
			<button type="button"
					x-show="!state"
					@click="state = true">
				Enable
			</button>
		</div>
		
		<h1>Expected Behavior</h1>
		<div x-data="{state: false}"
			 x-init="$watch('state', val => {
						setTimeout(() => {if (val) {state = false}}, 100)
					})">
			<button type="button"
					x-show="state"
					@click="state = false">
				Disable
			</button>
			<button type="button"
					x-show="!state"
					@click="state = true">
				Enable
			</button>
		</div>
	</body>
</html>
```
Codepen: https://codepen.io/foxbud/pen/xxYRjQv

vlad-iliescu-tallarium:
[...]

PR #2968, opened 2022-06-01
Fix x-show updating DOM inconsistently if state changes too fast

mgschoen:
This fixes a bug where multiple consecutive state changes in a short time cause x-show to update the DOM in an inconsistent way so that elements that should be hidden are visible. Demo: https://codepen.io/mgschoen/pen/XWZqprb 

This is caused by using `requestAnimationFrame` or `setTimeout` for the `show()` method, but `queueMicrotask()` for the `hide()` method, causing `hide()` calls from multiple state updates to be fired before the first `show()` call, effectively showing elements that should actually be hidden.

This bug has also been reported in https://github.com/alpinejs/alpine/issues/2803#issuecomment-1126751072


calebporzio:
This is great, thank you!

dtcooper:
This is a gnarly one - inconsistently breaks UI and worth releasing when you have a moment, @calebporzio. Ty!

PR #3002, opened 2022-06-23
Allow for using x-show with an important modifier

timkley:
Following up on [this discussion](https://github.com/alpinejs/alpine/discussions/1415).

There are cases where someone may use a class which itself sets the `display` property. If it is set with the `!important` property `x-show` stops working as expected because the class will always take precedence.

This PR adds an `.important` modifier to `x-show` which sets the inline style with `!important` itself.

I've added a test and updated the docs, please let me know if there are other changes needed :)


ryangjchandler (changes requested):

calebporzio:
This is great, thank you!

PR #3014, opened 2022-06-30
Add warning about x-if directive and livewire

hichxm:
Today, I past 4 hours of debugging about one Livewire component inside a `<template>` tag. 

The directive `x-if` don't work with any Livewire component, for this reason : The Livewire component generated by Laravel backend can't be removed and regenerated dynamically by JavaScript on the frontend.


joshhanley:
Thanks for the PR! We're going to close this due to Alpine being standalone from Livewire and could be used by others not using Livewire. As such this note should probably be in the Livewire documentation.

PR #3015, opened 2022-07-01
Allow x-modelable to be a descendant of x-model

TWithers:
With `x-modelable` being very helpful in templating, I find it hard to have both `x-model` and `x-modelable` attributes on the same element. Most of my components look something like below:

```
<div x-data="{ defaultData : 'Hello World' }"> 
    <x-fancy-input x-model="defaultData" x-data="{ options: { bold: true, onFocus: ()=>console.log('focused') } }"></x-fancy-input>
</div>

// component markup for <x-fancy-input> blade component
<div class="wrapper" {{ $attributes }}>
    <div x-data="fancyInput" x-modelable="parentModel">
        <textarea x-ref="input"></textarea>
    </div>
</div> 
<script>
Alpine.data('fancyInput', ...    
```

If the `x-modelable` directive is changed from `el._x_model` to `el.closest('[x-model]')` it will still select the current `<div>`, but if it doesn't exist, it will continue up the DOM tree until it is found. If one isn't found, it will return null, and the directive can be short-circuited at that point.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -42,6 +42,10 @@
             "resolved": "packages/intersect",
             "link": true
         },
+        "node_modules/@alpinejs/mask": {
+            "resolved": "packages/mask",
+            "link": true
+        },
         "node_modules/@alpinejs/morph": {
             "resolved": "packages/morph",
             "link": true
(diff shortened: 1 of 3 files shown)


ekwoka:
Running into a use case for this, I think it should also be smart to ensure it stops as the next parent `x-modelable` for if you have nested components with `x-modelable` on them.

Wouldn't want it to try to bind in a many to one relationship.

Issue #2996, opened 2022-06-21
$watch array output the same value between old and new.


dinyo:
maybe i am doing it wrong.
doing $watch on an array and edit the input value in loop, (console) output the same value between old and new.

code:
https://codepen.io/dinyo/pen/gOvJowE

thank you.

jokersk:
I think this is not an issue of alpine, even use that pure javascript will get the same result.
you setting the "field" in an object, and the object was pushed in an array, when you update the "field" value in this object
no-mater the new val or the oldval of the array is pointing to the same object in memory heap.

dinyo:
i think the problem in line 20 and line 23 in $watch.js
https://github.com/alpinejs/alpine/blob/main/packages/alpinejs/src/magics/%24watch.js

`oldValue = value`
because of deep watch, the value is a proxy object, then the oldvalue refer to the proxy object.

currently i am using this on my local build.
`oldValue = JSON.parse(JSON.stringify(value))`

puckowski:
Issue still seems to be present in latest main (08/23/2022 5:52 EST). See CodePen:
https://codepen.io/puckowski/pen/WNzWRQw?editors=1111

I think https://github.com/alpinejs/alpine/pull/3046 addresses the issue with additional Cypress tests added.

PR #3046, opened 2022-07-16
Add fix for issue 2996

puckowski:
* Add fix for issue 2996. Watch magic had an issue with value reference.

Should address: https://github.com/alpinejs/alpine/issues/2996

Cypress tests replicated here and all passing using the patch:
https://codepen.io/puckowski/pen/poLRwEr?editors=1111
39 assertions passing.

User reported in https://github.com/alpinejs/alpine/issues/2996 that the fix solved their issue.


puckowski:
image

Fix works for user reported Code Pen at: https://codepen.io/dinyo/pen/eYVaKgj?editors=1111
User Code Pen from: https://github.com/alpinejs/alpine/issues/2996

stepanjakl:
I am facing the same issue `this.$watch('filter.active', (value, oldValue) => { this.onFilterChange(value, oldValue) })`, where deeply nested values don't retain the `oldValue`. This PR should help.

basics09:
Thanks for creating this pull request @puckowski. 
Currently also entering a hole-of-dev-doom when using $watch.… : )

I’ve looked at the PR and when calling `JSON.parse(JSON.stringify(value))` on an array of `File` objects it will return an array of empty JS objects. This might not be what people using it are expecting.

If this PR gets accepted, the documenation should clarify the capabilites/limitation for `$watch` a bit better.

SimoTod:
Most recent browsers support structuredClone but it's quite new and you would need a polyfill for the rest (https://github.com/zloirock/core-js#structuredclone). I suppose $watch could warn the user that they should include a polyfill if structuredClone does not exist and it's watching an object (+ defaulting to your solution so it doesn't break) but just use structuredClone for recent browsers

packages/alpinejs/src/magics/$watch.js
```diff
@@ -17,10 +17,10 @@ magic('watch', (el, { evaluateLater, effect }) => (key, callback) => {
             queueMicrotask(() => {
                 callback(value, oldValue)
 
-                oldValue = value
+                oldValue = JSON.parse(JSON.stringify(value))
```
SimoTod, line 20: JSON.stringify is expensive, the code already does it at line 12 so you can just assign it to a temp variable and reuse that in here?

packages/alpinejs/src/magics/$watch.js
```diff
@@ -17,10 +17,10 @@ magic('watch', (el, { evaluateLater, effect }) => (key, callback) => {
...
             })
         } else {
-            oldValue = value
+            oldValue = JSON.parse(JSON.stringify(value))
```
SimoTod, line 23: same as above

puckowski:
I reworked my fix to use structuredClone, but it seems Cypress doesn't have access to structuredClone some some checks have failed with the latest push. Users would need to polyfill structuredClone for older browser support from core-js (https://www.npmjs.com/package/core-js) in order for this solution to work.
image

puckowski:
Will scrutinize the changes I have made later today. Had to update Cypress to get access to structuredClone, and also updated focus-trap with a minor code change in the focus index.js.

puckowski:
To summarize, I used structuredClone instead of JSON.parse(JSON.stringify()) which means that users of watch will need to polyfill structuredClone for older browser support. I log a warning if structuredClone is not defined on the window object.

I had to update Cypress from 5.5.0 to 9.7.0 in order for the tests to pass (newer version of browser with structuredClone support). 

Updating Cypress seems to have broken some tests for focus, so I updated focus-trap to 6.9.4 from 6.6.1 and made a small code change with comments to focus index.js.

@SimoTod

puckowski:
- Simple watch case working in https://codepen.io/puckowski/pen/VwXBKWZ?editors=1111
- All watch tests passing, plus some logic for watching File objects working in https://codepen.io/puckowski/pen/VwXBKWZ?editors=1111 (CodePen does not fully log File objects, add the code to a local IDE to test more thoroughly).
- Added test for watch with arrays containing objects.
- Removed unnecessary JSON.stringify call for performance.
- Added improved handling for cloning some proxied objects. Prefer structuredClone first. If that fails, shallow clone a new object and then perform a structuredClone on the shallow copy. Logic is working.

packages/alpinejs/src/magics/$watch.js
```diff
@@ -1,26 +1,87 @@
...
 
     let firstTime = true
 
     let oldValue
 
-    let effectReference = effect(() => evaluate(value => {
-        // JSON.stringify touches every single property at any level enabling deep watching
-        JSON.stringify(value)
+    const shallowClone = (obj) => {
+        if (typeof obj !== 'object') {
```
njpanderson, line 15: Hi all — just implemented this PR in my own project and got a JS error for my troubles. It was due to this line assuming that if `obj` is an object, then it can be parsed by Object.keys — This is incorrect, as `null` is also an object, but cannot be parsed:

```
Object.keys(null)
> Uncaught TypeError: can't convert null to object
``` 

I would recommend you add another check for `null`, e.g:

```
if (
    typeof obj !== 'object' ||
    obj === null
) {
```

(There may be other clauses needed here but this one got it working for me).

Also: Is it odd that it tested? Apologies, but I didn't have time to look at your test suites.

njpanderson: (Not doing this as an official review as I don't know your contrib process. Sorry if I'm stepping on toes here!)

njpanderson: Apologies guys but I've also noticed that the watcher only seems to detect the change in value. No subsequent changes are detected. I've not got a lot of time to look into this properly but thought you should know regardless. Thanks!

puckowski:
- Synced fork with latest main (08/21/22).
- Added support for null case and added two Cypress tests dealing with null cases, which are passing.
- https://codepen.io/puckowski/pen/PoRLZNO?editors=1111 demonstrates latest CDN build updates new and old values correctly

@njpanderson Let me know if this resolves your issue.

Old:
- Simple watch case working in https://codepen.io/puckowski/pen/VwXBKWZ?editors=1111
- All watch tests passing, plus some logic for watching File objects working in https://codepen.io/puckowski/pen/VwXBKWZ?editors=1111 (CodePen does not fully log File objects, add the code to a local IDE to test more thoroughly).
- Added test for watch with arrays containing objects.
- Removed unnecessary JSON.stringify call for performance.
- Added improved handling for cloning some proxied objects. Prefer structuredClone first. If that fails, shallow clone a new object and then perform a structuredClone on the shallow copy. Logic is working.

njpanderson:
I've given the latest push a few run throughs @puckowski and not seeing any further issues. Nice work!

kenset:
Apologies if this is not the right place to ask, but is there a timeline for this being merged and released?

ekwoka:
@kenset 

No. Caleb has been very busy with the Livewire rewrite and has a been a bit of an absent father to little Alpine. It may not be until after the new Livewire is released.

joshhanley:
Thanks for the PR and all the support!

It's well thought out and makes sense. There are a couple of concerns with this PR. One is the size of the implementation, it adds quite a few lines of code, which increases the package size. And the other is using an API that is less than a year old, which may cause issues for end users. 

We're going to leave it open for the moment, so it can be reviewed in detail.

joshhanley:

PR #3048, opened 2022-07-18
docs: switch unpkg to jsdelivr

ayuhito:
I think it would be appropriate for alpine.js to move away from unpkg since it is unmaintained and [full of issues](https://github.com/mjackson/unpkg/issues). On the other hand, [jsDelivr](https://www.jsdelivr.com/) is actively maintained and uses a more reliable [fallback system](https://www.jsdelivr.com/network/infographic) using multiple CDN networks.

Let me know if there are any concerns!


ekwoka:
While maybe out of scope with your change, showing how to use `esm.sh` for modules could be a nice addition as well.

joshhanley:
Not as nice url as unpkg, but makes sense, thanks!

Igloczek:
Sorry for digging this up, but, right on the https://alpinejs.dev/ there is still `<script src="//unpkg.com/alpinejs" defer></script>`.

ekwoka:
The homepage is something only @calebporzio can update.

PR #3051, opened 2022-07-19
Fix for #3040 re-evaluating x-transition with x-bind

Tim-Wils:
This pull request fixes re-evaluating expressions with x-transition binding.
See #3040 and #1633


Tim-Wils:
Fixed conflicts for 3.10.3

SimoTod:
In the example you posted on codepen _x_attributeCleanups seems to grow every time you change direction and retrigger the animation. I assume it's because you are rebinding new directives every time. I also suspect that, if you have an event listener (like @click) in your x-bind, Alpine will also duplicate events at each change.

Tim-Wils:
@SimoTod Thanks! It seems _x_attributeCleanups is indeed growing :/. That should not happen and I have to work out first how to clean these up. Do you perhaps have some ideas?

As for listeners: I just checked if it's also duplicating the clicks, but it seems like that's not happening.

SimoTod:
@Tim-Wils I'm not sure how you can clean up that list. It could include entries from other directives on the same element so it might be tricky to identify the ones you need.

Re listeners, I think they get duplicated under certain conditions, see https://codepen.io/SimoTod/pen/RwMVqyO.

I'm not completely sure that x-bind is the right place to fix the transition issue. Probably it has more to do with the fact that x-transition doesn't use effect and evaluateLater

Tim-Wils:
@SimoTod Yeah, you're probably right. I think maybe the expression in x-transition should use effect/evaluateLater instead https://github.com/alpinejs/alpine/blob/main/packages/alpinejs/src/directives/x-transition.js#L9
I'll try to look into that.

Tim-Wils:
@SimoTod I think I found a way to workaround the problem.
Would you mind checking out this one? https://codepen.io/tim-wils/pen/BardwOr

Tim-Wils:
@SimoTod I'm sorry bumping you :)

SimoTod:
Hi Tim, is there a branch where I can see the code changes?

Tim-Wils:
@SimoTod I'm sorry, not yet. This was still a build from my local dev environment. Shall I push the changes to this PR?

SimoTod:
Sounds good but no rush. I won't have much time today anyway.

Tim-Wils:
@SimoTod I've committed the changes. In short:
- x-bind.js doesn't do the effect anymore, but caches the expression in a `_x_bindExpression` property on the element
- x-transition uses the cached-property if the property exists and applies an evaluateLater on this cached expression

The most difficult thing is: A x-bind expression can return multiple attributes at once. Therefore we have to check if the "stage" (start/leave etc.) exists in the evaluated expression. Then that `expressionAtStage` can be passed to as how x-transition works.

SimoTod:
> @SimoTod I've committed the changes. In short:
> 
> * x-bind.js doesn't do the effect anymore, but caches the expression in a `_x_bindExpression` property on the element
> * x-transition uses the cached-property if the property exists and applies an evaluateLater on this cached expression
> 
> The most difficult thing is: A x-bind expression can return multiple attributes at once. Therefore we have to check if the "stage" (start/leave etc.) exists in the evaluated expression. Then that `expressionAtStage` can be passed to as how x-transition works.

It does work, I'm not totally sure but the problems are mostly in the existing x-transition directive that behaves in a different way (other directives want sting with quotes) and cannot be changed so I don't have a better idea for now. 
I guess we'll wait for Caleb to have a look and see what he thinks.

Tim-Wils:
@SimoTod Jup, my conclusions were exactly the same. x-bind in combination with x-transition is one of the most difficult processes along with cleaning up stuff.  Thanks for checking!

@calebporzio this is where you can chime in :) the PR works, but it needs your approval if it's the 'best' we can do to make this work for now?

joshhanley:
Thanks for the PR! We're currently going through and processing all the PRs but we're going to leave this one open for now so Caleb can have a look at it in more detail.

joshhanley:

ElpayetSupervan:
Hello @calebporzio @SimoTod is this change still on hold ?

PR #3054, opened 2022-07-23
Fixed throttling on keypress

Wulfheart:
Currently, the usage of a throttle modifier on a `keyup` or `keydown` event does not work. I took the ideas from @sterat's PR(#1911) and [discussion](https://github.com/alpinejs/alpine/discussions/1895) and applied the fix which was suggested by @SimoTod. 

This PR also includes a testcase which fails without the modification in the `x-on.js` file and passes with it.


joshhanley:
Looks good, thanks!

Wulfheart:
Thanks for merging!

PR #3083, opened 2022-08-04
Update events.md

HazJ:
I have added a note about using camel casing conventions for event names. There is some weird behaviour, I'm not sure if this is intended or not, but I couldn't find it documentated. Feel free to tidy up the wording and example if needed.

### DIFF
--- a/packages/docs/src/en/essentials/events.md
+++ b/packages/docs/src/en/essentials/events.md
@@ -7,31 +7,10 @@ title: Events
 
 Alpine makes it simple to listen for browser events and react to them.
 
-<a name="listening-for-simple-events"></a>
+<a name="a-note-about-naming-events"></a>
 ## A note about naming events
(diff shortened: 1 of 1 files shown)


SimoTod:
It's known. Attribute in html are case insensitive and they are always interpreted as lowercase. There is a special modifier for camel case events. See https://alpinejs.dev/directives/on#camel

HazJ:
Ah, I didn't know about this. I still think this could be improved by displaying a note on the events page. I'll update this later.

SimoTod:
No problem, be careful because you are mentioning another php library while people using Alpine may or may not be livewire users so your example won't make much sense to them.

HazJ:
I've just added a simple link for now. I saw some new markdown "Notice" banner but I can't find it right now. I'll look again later. I just think it's gonna be super useful to have a note, because I missed it and others might too.

PR #3086, opened 2022-08-05
Add $fetch

hankhank10:
This is my implementation of adding two new magics: `$fetch` and `$fetchjson`.

As discussed in [this discussion](https://github.com/alpinejs/alpine/discussions/3068#discussioncomment-3328978), I think Alpine currently lacks a straightforward way to request data from a server and populate it into an element.

This pull request takes my alpine-fetch library and implements it directly into Alpine. I recognise you might not want to merge this right away (would need to be documented first, in any case) but I've opened this for discussion.

### Making requests with Alpine today

It's quite hard work and verbose:

````
<div x-data="{
        result: null,
        async retrieveData() {
            this.result = await (await fetch('/endpoint')).text();
        }
     }"
     x-init="retrieveData()"
>
     <span x-text="result"></span>
</div>
````

(I'm aware this is a Javascript issue, not an Alpine issue)

### Making requests with $fetch

Much more intuitive as it abstracts away all of the promises and asyncs and other stuff that stump new JS devs.

Fetch some text and put it into a span:

````
<div x-data>
    <span x-text="await $fetch('/endpoint)"></span>
</div>
````

or fetch some JSON and put a particular element into a span:

````
<div x-data>
    <span x-text="await $fetchjson('https://weathermockapi.herokuapp.com/some_json', jsonItem='weather')"></span>
</div>
````

or load it into a x-data and use it wherever you want:

````
<div x-data="{ hello_world: await $fetch('https://weathermockapi.herokuapp.com/hello_world' }">
    <span x-text="hello_world"></span>
    <span x-text="hello_world"></span>
</div>
````

### Links

More examples: https://hankhank10.github.io/alpine-fetch/

### DIFF
--- a/packages/alpinejs/src/magics/$fetch.js
+++ b/packages/alpinejs/src/magics/$fetch.js
@@ -1,0 +1,51 @@
+import { magic } from "../magics";
+
+magic('fetchjson', () => {
+    return async (
+        url,
+        jsonItem = null,
+        method = "GET"
+    ) => {
+        let response = await xfetch(url = url, jsonItem = jsonItem, method = method)
+        return await response;
+    }
+})
+
+magic('fetch', () => {
+    return async (
+        url,
+        method = "GET"
+    ) => {
+        let response = await xfetch(url = url, jsonItem = null, method = method)
+        return await response;
+    }
+})
+
+
+// Actual fetch function
+async function xfetch(url, jsonItem = null, method = 'GET') {
+
+    if (jsonItem == null) {
+
+        return fetch(url, {method: method})
+            .then((response) => response.text())
+            .then((responseText) => {
+                return responseText
+            })
+            .catch((error) => {
+              console.log(error)
+            });
+
+    } else {
+
+        return fetch(url, {method: method})
+            .then((response) => response.json())
+            .then((responseJson) => {
+                return responseJson[jsonItem]
+            })
+            .catch((error) => {
+              console.log(error)
+            });
+
+    }
+}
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by ekwoka** [comment]
I'm not sure this should really be in the scope of AlpineJS, or even a core plugin.

Additionally this code is rather inefficient for the goal of keeping AlpineJS small. Quite a bit of unneeded repetition. I'm not sure why there are so many `=` in the your function calls. You already established the default behavior in one place, you don't need to keep doing it.

Good attempt, and the test seems good.

**Comment by hankhank10** [comment]
Thanks for your comments @ekwoka. The point around the overly verbose code is taken: spot the person who came to javascript from python!

**Comment by ekwoka** [comment]
> Thanks for your comments @ekwoka. The point around the overly verbose code is taken: spot the person who came to javascript from python!

Haha I was thinking it looked a bit like you expected kwargs to be a thing

**Comment by coderofsalvation** [comment]
I really like the idea of this plugin, but also agree with the comments above.
The smallest/fastest DIY way I could come up with is loading the json into a store:

```
<span x-text="$store.data.title"/>   <!-- from server        -->

<script>
          fetch('/data')
          .then( (res) => res.json() )
          .then( (data) => Alpine.store('data', data) )
        })
</script>
```

Vectrex:
The code only covers the most basic fetch there is (no custom headers, no handling of timeouts, no aborts, etc.) which can be - as already mentioned - pretty much covered by 3 or 4 lines of vanilla code. So you either blow up AlpineJS by a fair amount or end up writing your own custom fetch anyway.

joshhanley:
Thanks for being willing to contribute! Unfortunately, in the interest of keeping Alpine's API as minimal as possible and keeping the maintenance burden to a minimum, we're going to hold off for now on this. Thanks again!

PR #3091, opened 2022-08-06
Clarify cloak.md in docs

hankhank10:
The `x-cloak` page in the docs is less clear than it could be.

At present it doesn't address the main cause of "blipping": where an element is hidden by `x-show` but momentarily blips onto the page on load.

This edit clarifies that by using `x-show` as the first example, but also including the existing example (where an element "blips" onto page before its contents is set.

It also moves the "alternative" approach to a global style to its own sub-section.


joshhanley:
Thanks!

PR #3100, opened 2022-08-10
Add condition to $nextTick

MakiDizajnerica:
This pr adds condition to `$nextTick` magic method.
It can be useful in situations when for example you have accordion, and you want to scroll into view only when it gets expanded. So we are goint to use `$nextTick` to achieve that, but we don't want to scroll into view when it gets hidden.

```html
<div x-data="{
        expanded: false,
        toggle() {
            this.expanded = ! this.expanded;
            $nextTick(() => $refs.element.scrollIntoView(), this.expanded);
        },
    }">
    <button @click="toggle">Toggle Content</button>
 
    <div x-ref="element" x-show="expanded" x-collapse>
        ...
    </div>
</div>
```

So here we are passing second param to the `$nextTick` that will represent the condition.
If condition is true, the `callback` will run, otherwise it will not.

### DIFF
--- a/packages/alpinejs/src/nextTick.js
+++ b/packages/alpinejs/src/nextTick.js
@@ -3,7 +3,7 @@ let tickStack = []
 
 let isHolding = false
 
-export function nextTick(callback = () => {}) {
+export function nextTick(callback = () => {}, condition = true) {
   queueMicrotask(() => {
     isHolding || setTimeout(() => {
       releaseNextTicks()
@@ -12,7 +12,7 @@ export function nextTick(callback = () => {}) {
 
   return new Promise((res) => {
     tickStack.push(() => {
-        callback();
+        if (condition) callback();
         res();
     });
   })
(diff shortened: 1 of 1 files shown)


SimoTod:
How does it differ from `if(condition) $nextTick(...)` (or `condition && $nextTick(...)`)? 
It's also a dangerous feature to add because depending on how you word it in the documentation, people will think that the condition is evaluated on the next tick which is not what it does.

MakiDizajnerica:
Well it is a lot cleaner I think. And yes, the naming should be changed probably to something like runCallback or something like that.

ekwoka:
I agree with others, this just doesn't seem like what nextTick is supposed to be.

joshhanley:
Thanks for being willing to contribute! We're going to close this PR, as pointed out above, this could be easily misunderstood and wrapping in a conditional or using short-circuit evaluation is straight forward enough.

PR #3105, opened 2022-08-13
Enhance directive to accept order option

vimeshjs:
When creating a new directive, it is always executed at low priority. 

```javascript
const DEFAULT = 'DEFAULT'

let directiveOrder = [
    'ignore',
    'ref',
    'data',
    'id',
    'bind',
    'init',
    'for',
    'mask',
    'model',
    'modelable',
    'transition',
    'show',
    'if',
    **DEFAULT**,
    'teleport',
]
```

In some cases, it is just unacceptable. For example, when you try to create some component directive, just like Alpine headless UI does,

```javascript
export default function (Alpine) {
    Alpine.directive('dialog', (el, directive) => {
        if      (directive.value === 'overlay')     handleOverlay(el, Alpine)
        else if (directive.value === 'panel')       handlePanel(el, Alpine)
        else if (directive.value === 'title')       handleTitle(el, Alpine)
        else if (directive.value === 'description') handleDescription(el, Alpine)
        else                                        handleRoot(el, Alpine)
    })
....

```

We may define some properties in `data` as default value. We want to be able to override them at the html element. Just imagine we have a `type` property for `dialog`. There are three types : info, warn, error. 
```javascript

function handleRoot(el, Alpine) {
    Alpine.bind(el, {
        'x-data'() {
            return {
               type: 'info'
....
            }
....
        }
```
When we use this x-dialog directive, we want to use it as other UI framework does
```html
<div x-dialog x-data="{type:'error'}" ... >...</div>
```
The problem is x-data is executed before x-dialog, the type of this dialog is alway 'info'. It will cause some extra side effects. 
```html
<div x-dialog x-data="{}" :mytype="type" >...</div>
```
The above code will cause error, because :mytype is executed via x-bind directive, it is executed before x-dialog. It means `type` property is not yet been initiated.

Now, the solution is allowing set the order of directive when defining it.

Firstly, change directiveOrder from an array to a map (name -> order) and make `byPriority` use the assigned order
```javascript

const DEFAULT_ORDER = 10000

let directiveOrder = {
    'ignore'        : 100,
    'ref'           : 200,
    'data'          : 300,
    'id'            : 400,
    'bind'          : 500,
    'init'          : 600,
    'for'           : 700,
    'mask'          : 800,
    'model'         : 900,
    'modelable'     : 1000,
    'transition'    : 1100,
    'show'          : 1200,
    'if'            : 1300,
    // DEFAULT      : DEFAULT_ORDER = 10000
    'teleport'      : 20000,
}

function byPriority(a, b) {
    let orderA = directiveOrder[a.type] || DEFAULT_ORDER
    let orderB = directiveOrder[b.type] || DEFAULT_ORDER

    return orderA - orderB
}

```
Then we add order option to directive function
```javascript

export function directive(nameOrOptions, callback) {
    let name = nameOrOptions
    if (nameOrOptions.name){
        name = nameOrOptions.name
        let order = nameOrOptions.order
        if (order){
            if (!Number.isInteger(order)){
                try{
                    order = new Function('orders', `return ${order}`)(directiveOrder);
                }catch(e){
                    console.log(`Fails to evaluate "${order}" to order`, e)
                }
            }
            directiveOrder[name] = order
        }
    }
    directiveHandlers[name] = callback
}
```
Now let's re-define the x-dialog as
```javascript

export default function (Alpine) {
    Alpine.directive({name:'dialog', order: 250}, (el, directive) => {

```
The x-dialog will be execute before x-data, we solve the previous issues.
If we don't want hard code the order as 250. We could use an expression instead

```javascript

export default function (Alpine) {
    Alpine.directive({name:'dialog', order: 'orders.data - 50', (el, directive) => {

```

This change is essential for creating Alpine UI framework

### DIFF
--- a/index.html
+++ b/index.html
@@ -8,7 +8,7 @@
     <script src="./packages/ui/dist/cdn.js" defer></script>
     <script src="./packages/alpinejs/dist/cdn.js" defer></script>
     <script src="//cdn.tailwindcss.com"></script>
-    <!-- <script src="https://unpkg.com/alpinejs@3.0.0/dist/cdn.min.js" defer></script> -->
+    <!-- <script src="https://cdn.jsdelivr.net/npm/alpinejs@3.0.0/dist/cdn.min.js" defer></script> -->
 
     <!-- <div x-data="{ value: null }">
         Value: <span x-text="value"></span>
--- a/morph.html
+++ b/morph.html
@@ -1,7 +1,7 @@
 <html>
     <script src="./packages/morph/dist/cdn.js" defer></script>
     <script src="./packages/alpinejs/dist/cdn.js" defer></script>
-    <!-- <script src="https://unpkg.com/alpinejs@3.0.0/dist/cdn.min.js" defer></script> -->
+    <!-- <script src="https://cdn.jsdelivr.net/npm/alpinejs@3.0.0/dist/cdn.min.js" defer></script> -->
 
 <div id="before">
 <!-- Before markup goes here: -->
--- a/packages/alpinejs/package.json
+++ b/packages/alpinejs/package.json
@@ -2,6 +2,12 @@
     "name": "alpinejs",
     "version": "3.10.5",
     "description": "The rugged, minimal JavaScript framework",
+    "homepage": "https://alpinejs.dev",
+    "repository": {
+        "type": "git",
+        "url": "https://github.com/alpinejs/alpine.git",
+        "directory": "packages/alpinejs"
+    },
     "author": "Caleb Porzio",
     "license": "MIT",
     "main": "dist/module.cjs.js",
--- a/packages/alpinejs/src/directives/x-model.js
+++ b/packages/alpinejs/src/directives/x-model.js
@@ -1,6 +1,7 @@
 import { evaluateLater } from '../evaluator'
 import { directive } from '../directives'
 import { mutateDom } from '../mutation'
+import { nextTick } from '../nextTick'
 import bind from '../utils/bind'
 import on from '../utils/on'
 
@@ -33,6 +34,17 @@ directive('model', (el, { modifiers, expression }, { effect, cleanup }) => {
     el._x_removeModelListeners['default'] = removeListener
 
     cleanup(() => el._x_removeModelListeners['default']())
+
+    // If the input/select/textarea element is linked to a form
+    // we listen for the reset event on the parent form (the event
+    // does not trigger on the single inputs) and update
+    // on nextTick so the page doesn't end up out of sync
+    if (el.form) {
+        let removeResetListener = on(el.form, 'reset', [], (e) => {
+            nextTick(() => el._x_model && el._x_model.set(el.value))
+        })
+        cleanup(() => removeResetListener())
+    }
 
     // Allow programmatic overiding of x-model.
     let evaluateSetModel = evaluateLater(el, `${expression} = __placeholder`)
--- a/packages/alpinejs/src/directives/x-teleport.js
+++ b/packages/alpinejs/src/directives/x-teleport.js
@@ -4,7 +4,7 @@ import { mutateDom } from "../mutation"
 import { addScopeToNode } from "../scope"
 import { warn } from "../utils/warn"
 
-directive('teleport', (el, { expression }, { cleanup }) => {
+directive('teleport', (el, { modifiers, expression }, { cleanup }) => {
     if (el.tagName.toLowerCase() !== 'template') warn('x-teleport can only be used on a <template> tag', el)
 
     let target = document.querySelector(expression)
@@ -31,7 +31,16 @@ directive('teleport', (el, { expression }, { cleanup }) => {
     addScopeToNode(clone, {}, el)
 
     mutateDom(() => {
-        target.appendChild(clone)
+        if (modifiers.includes('prepend')) {
+            // insert element before the target
+            target.parentNode.insertBefore(clone, target)
+        } else if (modifiers.includes('append')) {
+            // insert element after the target
+            target.parentNode.insertBefore(clone, target.nextSibling)
+        } else {
+            // origin
+            target.appendChild(clone)
+        }
 
         initTree(clone)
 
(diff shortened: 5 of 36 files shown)


packages/alpinejs/src/directives.js
```diff
@@ -15,7 +15,22 @@ export function setPrefix(newPrefix) {
...
 
-export function directive(name, callback) {
+export function directive(nameOrOptions, callback) {
+    let name = nameOrOptions
+    if (nameOrOptions.name){
+        name = nameOrOptions.name
+        let order = nameOrOptions.order
+        if (order){
+            if (!Number.isInteger(order)){
+                try{
```
ekwoka, line 25: Idea is great, I do remember something about this coming out even at Alpine Day when v3 was announced.

But I think this is needlessly complicated way to write this.

Why use a function to to...what exactly?

If it's not a number, just discard it and maybe `console.warn` about it. No need for extra steps that don't, to my eyes, add any value. Or is the point to allow someone to pay anything that resolves to a number? but why can't the end user just resolve the number themselves?

ekwoka (changes requested):

packages/alpinejs/src/directives.js
xinjie-zhang: `order` could be a number or an expression, like "`orders.data - 1`". Since there are no exported const or enums for the internal `directiveOrder` definition, you have to dive into the source code to find out the order value of `data`
```javascript

let directiveOrder = {
    'ignore'        : 100,
    'ref'           : 200,
    'data'          : 300,
    'id'            : 400,
    'bind'          : 500,
    'init'          : 600,
    'for'           : 700,
    'mask'          : 800,
    'model'         : 900,
    'modelable'     : 1000,
    'transition'    : 1100,
    'show'          : 1200,
    'if'            : 1300,
    // DEFAULT      : DEFAULT_ORDER = 10000
    'teleport'      : 20000,
}
```
With expression supported, if you want to insert a directive with higher priority than `bind`, just give an expression as "`orders.bind - 1`". Lower priority than `model`? Just use "orders.model + 1".  You do not have to know what's the exact order value of target directive.

ekwoka: Ah, okay, very smart!!!

Had not thought of that kind of use case.

Accudio:
I would be a fan of something like this. I work on a library that adds async/on-demand component loading to Alpine called `[Async Alpine](https://github.com/Accudio/async-alpine)`. It currently uses attribute renaming, but I am currently re-working it to use `x-ignore`.

If we were to use Alpine's plugin system we would have to require developers to put `x-ignore` on their elements in addition to our attributes in order to prevent Alpine from processing components immediately.

Ideally however we'd be able to create a directive that runs first, and we can add `x-ignore` to components ourselves. That would make it easier for developers to use and allow us to integrate with Alpine as a plugin rather than a wrapper library.

In terms of implementation @vimeshjs's proposal seems like a good idea to me, and a priority system like that is fairly common and familiar to many developers I think.

joshhanley:
Thanks for the PR! We can certainly see a need for this, we're not sure if this is the ideal solution though. We're going to leave this open for now to see if we can come up with any other possible solutions. We're open to any ideas!

SimoTod:
How about
```html
Alpine.directive('dialog', (el, directive) => {
   ...
}).before('bind');
```
basically, if you call the directive normally, it will go in the default order but if you chain before, you can decide to run it before an existing directive.

I have a working PR with just minimal changes to a single function if we like the option.

https://codepen.io/SimoTod/pen/WNyypEG

xinjie-zhang:
> How about
> 
> ```
> Alpine.directive('dialog', (el, directive) => {
>    ...
> }).before('bind');
> ```
> 
> basically, if you call the directive normally, it will go in the default order but if you chain before, you can decide to run it before an existing directive.
> 
> I have a working PR with just minimal changes to a single function if we like the option.
> 
> https://codepen.io/SimoTod/pen/WNyypEG

It looks better and cleaner

xinjie-zhang:
Now with current directive implementation, we developed a new library [Vimesh UI](https://github.com/vimeshjs/vimesh-ui). It will help to build sophisticated UI components. Some examples:

> /counters.html
```html
<head>
    <script src="https://unpkg.com/@vimesh/style" defer></script>
    <script src="https://unpkg.com/@vimesh/ui"></script>
    <script src="https://unpkg.com/alpinejs" defer></script>

    <script>
        $vui.config.importMap = {
            "*": './components/${component}.html'
        }
    </script>
    <style>
        [x-cloak] {
            display: none !important;
        }
    </style>
</head>

<body x-cloak x-import="counter" class="p-2" x-data="{name: 'Counter to rename', winner: 'Jacky'}">   
    Rename the 2nd counter : <input type="text" x-model="name" class="rounded-md border-2 border-blue-500">
    <vui-counter x-data="{step: 1}" :primary="true" title="First" x-init="console.log('This is the first one')" owner-name="Tom"></vui-counter>
    <vui-counter x-data="{step: 5}" :title="name + ' @ ' + $prop('owner-name')" owner-name="Frank"></vui-counter>
    <vui-counter x-data="{step: 10, value: 1000}" :owner-name="winner"></vui-counter>
</body>
```
> /components/counter.html
```html
<template x-component.unwrap="counter" :class="$prop('primary') ? 'text-red-500' : 'text-blue-500'"
    x-data="{ step : 1, value: 0}" x-init="$api.init && $api.init()" title="Counter" owner-name="nobody">
    <div>
        <span x-text="$prop('title')"></span><br>
        Owner: <span x-text="$prop('owner-name')"></span><br>
        Step: <span x-text="step"></span><br>
        Value : <span x-text="value"></span><br>
        <button @click="$api.increase()"
            class="inline-block rounded-lg bg-indigo-600 px-4 py-1.5 text-white shadow ring-1 ring-indigo-600 hover:bg-indigo-700 hover:ring-indigo-700">
            Increase
        </button>
    </div>
    <script>
        return {
            init() {
                console.log(`Value : ${this.value} , Step : ${this.step}`)
            },
            increase() {
                this.value += this.step
            }
        }
    </script>
</template>
```

PR #3122, opened 2022-08-28
x-mask: Allow overriding thousands separator

julianwachholz:
- [x] Allow second argument to `$money`  to override thousands separator
- [x] Backwards compatible
- [x] Added tests
- [x] Added documentation

### DIFF
--- a/packages/mask/src/index.js
+++ b/packages/mask/src/index.js
@@ -163,9 +163,8 @@ export function buildUp(template, input) {
     return output
 }
 
-function formatMoney(input, delimeter = '.', thousands) {
-    thousands = (delimeter === ',' && thousands === undefined)
-        ? '.' : ','
+export function formatMoney(input, delimiter = '.', thousands) {
+    thousands = thousands ?? delimiter === "," ? "." : ","
 
     let addThousands = (input, thousands) => {
         let output = ''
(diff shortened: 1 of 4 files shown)


packages/mask/src/index.js
```diff
@@ -163,9 +163,9 @@ export function buildUp(template, input) {
     return output
 }
 
-function formatMoney(input, delimeter = '.', thousands) {
+export function formatMoney(input, delimeter = '.', thousands) {
     thousands = (delimeter === ',' && thousands === undefined)
-        ? '.' : ','
+        ? '.' : thousands || ','
```
ekwoka, line 168: Simple and clean.

I'd probably do this as 

```js
thousands = thousands ?? delimiter === ',' ? '.' : ','
```
This would allow it to short circuit on a provided value quicker, and then also the ternary is clearer.

Maybe also fix the misspelling of `delimeter` while we're here.

julianwachholz: Hi Eric, thank you for the nice feedback. Indeed the `??` operator is clearer. I went ahead and renamed the delimiter argument.

gaetan-hexadog:
Hi,

any news regarding this ?

calebporzio:
Looks great, thanks!

joshhanley:
@julianwachholz and @ekwoka FYI there was a bug with the ternary condition taking precedence over the custom thousands separator, so have pushed a fix in this commit https://github.com/alpinejs/alpine/commit/d0b91cb5af3c636abeb77b51a9e175275f2a47c7. Just needed to wrap the ternary condition in parenthesis.

PR #3134, opened 2022-09-04
:bug: prevents coercion of space and underscore to minus

ekwoka:
Addresses bug found in https://github.com/alpinejs/alpine/discussions/3124

This PR:
- Prevents `' '` and `'_'` from being coerced into `'-'` in the `kebabcase` function.
- Adds test for these cases.
- Adds the modifier `minus` and `underscore` as aliases for `-` and `_` when using them as `@keydown` modifiers.

This will break any apps that actually depends on this coercion, but I don't think that would be an expected behavior that anyone would be relying on.

### DIFF
--- a/packages/alpinejs/src/utils/on.js
+++ b/packages/alpinejs/src/utils/on.js
@@ -93,6 +93,8 @@ function isNumeric(subject){
 }
 
 function kebabCase(subject) {
+    if ([' ','_'].includes(subject
+    )) return subject
     return subject.replace(/([a-z])([A-Z])/g, '$1-$2').replace(/[_\s]/, '-').toLowerCase()
 }
 
@@ -149,8 +151,8 @@ function keyToModifiers(key) {
     let modifierToKeyMap = {
         'ctrl': 'control',
         'slash': '/',
-        'space': '-',
-        'spacebar': '-',
+        'space': ' ',
+        'spacebar': ' ',
         'cmd': 'meta',
         'esc': 'escape',
         'up': 'arrow-up',
@@ -159,6 +161,8 @@ function keyToModifiers(key) {
         'right': 'arrow-right',
         'period': '.',
         'equal': '=',
+        'minus': '-',
+        'underscore': '_',
     }
 
     modifierToKeyMap[key] = key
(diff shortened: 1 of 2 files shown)


packages/alpinejs/src/utils/on.js
```diff
@@ -93,6 +93,8 @@ function isNumeric(subject){
 }
 
 function kebabCase(subject) {
+    if ([' ','_'].includes(subject
+    )) return subject
```
SimoTod, line 97: If that is the intended behaviour, wouldn't you just remove `.replace(/[_\s]/, '-')` from the next line instead?

ekwoka: Then it won't kebab case strings.

I'm not positive what the kebabcase is doing here (maybe I can't think up a good reason it exists?).

So I elected to not remove something I didn't understand the use-case for.

I would assume that removing the kebabcase part entirely would work.

SimoTod: if there are spaces or underscores, they will never hit the last line because of your addition making that instruction redundant anyway, though.
If `Arrow Right` needs to be converted to `arrow-right` then your addition would break things as well.

Unfortunately I don't have a good reference of key names on all browser, but that should probably be looked at before introducing this change.

ekwoka: No it won't, this will only catch ` ` and `_` not anything that includes those.

I think you're reading the flow backwards.

This is equivalent to `if (subject === ' ' || subject === '_')`

SimoTod: Ah, right. You are looking for exact matches. Fair enough, just ignore my comment. 😅

tests/cypress/integration/directives/x-on.spec.js
```diff
@@ -330,6 +330,30 @@ test('keydown modifiers',
...
+        get('span').should(haveText('1'))
+        get('#space').type('-')
+        get('span').should(haveText('1'))
+        get('#minus').type('-')
+        get('span').should(haveText('2'))
+        get('#minus').type(' ')
+        get('span').should(haveText('2'))
+        get('#underscore').type('_')
+        get('span').should(haveText('3'))
+        get('#underscore').type(' ')
```
marcreichel, line 354:
```suggestion
        get('span').should(haveText('3'))
        get('#underscore').type(' ')
        get('span').should(haveText('3'))
```

The last assertion is missing I think.

marcreichel (changes requested):

marcreichel (approved):

joshhanley:
Looks good, thanks!

PR #3143, opened 2022-09-06
Form reset support for x-model

SimoTod:
If an element using x-model is included in a form and the form provides a reset button, clicking said button will desync the UI from the alpine state (all input elements get emptied but the Alpine scope won't see the update).

Failing test before fix: https://github.com/SimoTod/alpine/runs/8214754814?check_suite_focus=true#step:6:572

### DIFF
--- a/packages/alpinejs/src/directives/x-model.js
+++ b/packages/alpinejs/src/directives/x-model.js
@@ -35,6 +35,17 @@ directive('model', (el, { modifiers, expression }, { effect, cleanup }) => {
 
     cleanup(() => el._x_removeModelListeners['default']())
 
+    // If the input/select/textarea element is linked to a form
+    // we listen for the reset event on the parent form (the event
+    // does not trigger on the single inputs) and update
+    // on nextTick so the page doesn't end up out of sync
+    if (el.form) {
+        let removeResetListener = on(el.form, 'reset', [], (e) => {
+            nextTick(() => el._x_model && el._x_model.set(el.value))
+        })
+        cleanup(() => removeResetListener())
+    }
+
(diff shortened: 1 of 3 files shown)


packages/alpinejs/src/directives/x-model.js
```diff
@@ -35,6 +35,17 @@ directive('model', (el, { modifiers, expression }, { effect, cleanup }) => {
 
     cleanup(() => el._x_removeModelListeners['default']())
 
+    // If the input/select/textarea element is linked to a form
+    // we listen for the reset event on the parent form (the event
+    // does not trigger on the single inputs) and update
+    // on nextTick so the page doesn't end up out of sync
+    if (el.form) {
+        let removeResetListener = on(el.form, 'reset', [], (e) => {
+            nextTick(() => el._x_model && el._x_model.set(el.value))
```
ekwoka, line 44: Is there really risk that `el._x_model` not be defined here?

could `el._x_model?.set(el.value)` not work?

SimoTod: I don't think it will ever be undefined but who knows?
I assume you can build a complex case where the reset button has an @click that toggles a boolean, acting on an x-if that removes the input and, because this handler waits for the next tick it may or may not error.

I just thought it was just safer to add a small check. 

the below will work too but I don't have time to check compatibility for all browsers (I'm not a js/frontend dev 😂)

ekwoka: Haha true.

Conditional chaining can be transpiled by bundlers when the consumer is targeting older things.

calebporzio:
You da man @SimoTod - thanks!

PR #3146, opened 2022-09-07
allow await in invalid expression wrapper

ekwoka:
Solves: https://github.com/alpinejs/alpine/discussions/3118

Problem: expressions that are identified as invalid right hand expressions are wrapped in an IIFE, which brakes any `await` calls.

Solution: make IIFE async

Expressions are normally treated as async functions, but this was one situation in which they were not.


joshhanley:
@ekwoka have pushed an empty commit to trigger CI, as I think the failing test is unrelated, but just want to make sure

joshhanley:
@SimoTod if you get the chance, can you comment on Caleb's concern above?

ekwoka:
It shouldn't change the behavior for anything else, in this case.

Async functions, when called, run synchronously up to the first await, and then the resolution is handled in the queue, but the promises from the outer `AsyncFunction` and this inner `async()={}` should virtually collapse together.

joshhanley:
Yep makes sense to me. Thanks for the PR!

PR #3156, opened 2022-09-12
Replace inital-focus with x-dialog:focus

jonnitto:
I think it would be more comfortable to add `x-dialog:focus` instead of `:initial-focus`.

Based on https://github.com/alpinejs/alpine/discussions/3155


jonnitto:
Can someone re-run the tests?

SimoTod:
I think only Caleb can re-run it. The failing test is in the main branch anyway so it will fail again.

calebporzio:
If we weren't trying to keep API parity with Headless UI, I'd say this is a logical addition. However, to keep as similar to Headless UI as possible, I'm going to close this.

PR #3174, opened 2022-09-26
remove letters from money mask

PH7-Jack:
This PR solves the problem of adding letters on empty inputs using the money mask, the mask stops working

**Unexpected Behavior:**
Ability to type letters on money input
image

**Expected Behavior:**
Letters and non-numeric characters should be removed

**Live example:**
1. Go to the alpine mask (documentation)[https://alpinejs.dev/plugins/mask#money-inputs]
2. Click in the money input
3. Type **"ABC"**


packages/mask/src/index.js
```diff
@@ -164,6 +164,8 @@ export function buildUp(template, input) {
 }
 
 function formatMoney(input, delimeter = '.', thousands) {
+    if (!input.replace(/\D+/g, '')) return '9'
```
ekwoka, line 167: Shouldn't this be a regex test and not a replace?

ekwoka (changes requested):

packages/mask/src/index.js
PH7-Jack: @ekwoka Yes, it is possible to use the both

ekwoka: I don't mean whats possible, I mean what's the correct tool for the job?

Wouldn't

```js
if (!/^\D+$/.test(input)) return '9'
```

Be more explicit to what you're actually doing?

joshhanley:
@PH7-Jack I have merged in main and there are some failing tests. Could you please address them and also change to using `.test()` as suggested by @ekwoka? Thanks!

joshhanley:
@PH7-Jack turns out the failing test is unrelated, I've pushed a fix to main, so leave it with me for the moment and I will merge in changes and check it's all good.

PH7-Jack:
@joshhanley Ok, if you need something just let me know.
Thanks

joshhanley:
@PH7-Jack thanks for the PR! I've updated it to use regex test and merged it.

PR #3224, opened 2022-10-23
Make store setter a constructor also

dy:
Related: #3223 

It seems _constructor_ pattern is more widespread with store than _setter_. It's unlikely store is set multiple times during the app lifetime, whereas returning just created store is more natural (coming from [@preact/signals](https://www.npmjs.com/package/@preact/signals) or react hooks world).

```js
let store = Alpine.store('auth', {user});
onAuthChanged(user => store.user = user);
```

### DIFF
--- a/packages/alpinejs/src/store.js
+++ b/packages/alpinejs/src/store.js
@@ -18,6 +18,8 @@ export function store(name, value) {
     }
 
     initInterceptors(stores[name])
+    
+    return stores[name]
 }
 
 export function getStores() { return stores }


ekwoka:
It's a bit against the Alpine philosophy, but it is a very small change.

dy:
I am a bit surprised there's a philosophy for returning a result of constructor, which is natural expectation.
If that's the case, I don't think I like alpine philosophy.

ekwoka:
Well, the philosophy is that you don't really have other JavaScript accessing and manipulating data in Alpine components.

The philosophy is that if it affects a component, it should be done in the component and by the component.

Not unlike how in React you don't have other code outside of react stepping in and changing variables.

dy:
@ekwoka speaking about stores, not component state. There's plenty of react stores solutions, from redux to preact-signals, main point of which is controlling component from outside. Not willing to take part in debate, that philosophy of imposing itself instead of openness is alien to me. Wish alpinejs well.

ekwoka:
Sure, but that's more often side effect of the implementation.

Alpine has it...just use the store as a getter.

It's unlikely that you really need such behavior anyway. I've worked on rather large and complex Alpine apps and never needed to access a store or component like that.

SimoTod:
@dy none of us is involved with the development of the library. 

You can just leave the PR open and wait for the maintainer to chime in.

PR #3243, opened 2022-11-02
Fix issue #2208 x-collapse caused by zoom levels

hoektoe:
Zoom levels other than 100% is returning 3 decimal points values for height compared to standard 2 decimal used in browsers. Thus the condition to remove overflow:hidden never becomes truthy as 92.312  == 92.31 => false


hoektoe:
Unsure if I should make this change for closing action also. But I doubt could be issue since zero is zero

calebporzio:
I dig. Thank you.

PR #3252, opened 2022-11-03
Nested-templates-order-fix

Tim-Wils:
There's is this bug, where the order of rendering is wrong if there's multiple templates.

Referred discussions:
#3245 
#3032


ekwoka:
How will this work with situation in which the outer forloop changes counts?

ekwoka:
I think the only way is to have the wrapper div around the inner loop or to flatmap the inner stuff you want.

ekwoka:
I think the only way is to have the wrapper div around the inner loop or to flatmap the inner stuff you want.

Alpine could potentially inject a div with display: contents as an opinionated fix...

PR #3268, opened 2022-11-10
Introduce x-teleport values/modifiers

comes:
Sometimes it is needed to teleport a node after or before an existing node.

Example:
```html
<table>
  <template x-for="n in 10">
    <tr x-bind:id="`row-${n}`">
      <template x-teleport:append="#row-5">
        <tr>
          <td colspan="2">EVERYTHING ABOVE 5</td>
        </tr>
      </template>
      <td>#</td>
      <td x-text="n"></td>
    </tr>
  </template>
</table>
```
The example above will add a new `<tr>` correctly

Especially when you deal with `<tr>`, you are hitting a problem here because you can only have one single root element within a `<template>`. Sometimes, you need to have a second one.

With `x-teleport:append`, you can append a node as a sibling of the target instead of a child. Works same wise with `x-teleport:prepend`

### DIFF
--- a/packages/alpinejs/src/directives/x-teleport.js
+++ b/packages/alpinejs/src/directives/x-teleport.js
@@ -4,7 +4,7 @@ import { mutateDom } from "../mutation"
 import { addScopeToNode } from "../scope"
 import { warn } from "../utils/warn"
 
-directive('teleport', (el, { expression, value  }, { cleanup }) => {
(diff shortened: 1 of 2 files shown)


joshhanley:
@comes thanks for the PR! We've updated prepend/append to be modifiers so you would use it like `x-teleport.append` instead.

We've merged this one now, but is there any chance you could PR some documentation for this? Thanks!

PR #3307, opened 2022-12-01
Custom directive order

SimoTod:
Failing test before the change: https://github.com/alpinejs/alpine/actions/runs/3594849581/jobs/6053624912#step:6:833

### DIFF
--- a/packages/alpinejs/src/directives.js
+++ b/packages/alpinejs/src/directives.js
@@ -17,6 +17,23 @@ let directiveHandlers = {}
 
 export function directive(name, callback) {
     directiveHandlers[name] = callback
+
+    return {
+        before(directive) {
+            if (!directiveHandlers[directive]) {
+                console.warn(
+                    "Cannot find directive `${directive}`. "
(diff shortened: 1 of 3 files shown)


packages/alpinejs/src/directives.js
```diff
@@ -17,6 +17,23 @@ let directiveHandlers = {}
...
+        before(directive) {
+            if (!directiveHandlers[directive]) {
+                console.warn(
+                    "Cannot find directive `${directive}`. "
+                    + "`${name}` will use the default order of execution"
+                );
+                return;
+            }
+            const pos = directiveOrder.indexOf(directive)
+                ?? directiveOrder.indexOf('DEFAULT');
```
SimoTod, line 31: this is because some of the directives (x-text for example) are not explicitly in the directiveOrder list

calebporzio:
Love this, thanks @SimoTod!!

SimoTod:
@calebporzio no worries. If you like, you can use it for the ui package so you can tidy up the directive order list in the core package.

PR #3309, opened 2022-12-01
Fix recursion problem with x-trap inert

austenc:
Using `x-trap.inert` causes some extra recursion that I don't think is intended.

The `crawlSiblingsUp` function is working overtime and re-crawling elements that have already had `aria-hidden="true"` applied. This is even more apparent when there are tons of DOM elements, and it causes some significant performance problems in that case. The codepen below illustrates the problem in the console. 

**Codepen (see the console in codepen)**
https://codepen.io/austenc/pen/XWYxeoB

When the button is clicked, the console produces four log lines:
CleanShot 2022-12-01 at 14 40 48

I'm not sure how to write a test for this recursion case. Perhaps we could do something similar to the codepen with mutation observer to check those sibling elements are only mutated once?

Any help on the tests is much appreciated!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -42,6 +42,10 @@
             "resolved": "packages/intersect",
             "link": true
         },
+        "node_modules/@alpinejs/mask": {
+            "resolved": "packages/mask",
+            "link": true
+        },
         "node_modules/@alpinejs/morph": {
             "resolved": "packages/morph",
             "link": true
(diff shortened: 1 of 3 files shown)


austenc:
Added a test for this. All credit goes to @cmoses1020 who actually wrote the test 🙇

austenc:
Okay, I thought this test was failing because of this PR, but it looks like it's just a flaky test.

See this github actions run from yesterday:
https://github.com/alpinejs/alpine/actions/runs/3594967514/jobs/6053885520

It passes most of the time on my local machine, but occasionally fails there too.

calebporzio:
Legends. Thanks @austenc!!! (and @cmoses1020)

PR #3312, opened 2022-12-02
Bug (UI) Popover not focusing correctly

SimoTod:
I believe the reason why the test fails inconsistently is that there is a race condition between showing the popup and focusing on the first element. If an element is not visible, it can't be focused. This means that the focus stays on the button and pressing `tab` is actually moving the focus inside the popover (rather than tabbing away).

I suppose this could also happen on a real website under certain conditions.


austenc:
That dang test got me today, too! Thanks @SimoTod

SimoTod:
Closing as the test is still failing

PR #3318, opened 2022-12-05
Add visibility modifier for x-show (#2831)

stepanjakl:
This PR implements visibility modifier for the x-show directive (#2831)

### DIFF
--- a/packages/alpinejs/src/directives/x-show.js
+++ b/packages/alpinejs/src/directives/x-show.js
@@ -9,16 +9,27 @@ directive('show', (el, { modifiers, expression }, { effect }) => {
     // We're going to set this function on the element directly so that
     // other plugins like "Collapse" can overwrite them with their own logic.
     if (! el._x_doHide) el._x_doHide = () => {
-        mutateDom(() => {
-            el.style.setProperty('display', 'none', modifiers.includes('important') ? 'important' : undefined)
-        })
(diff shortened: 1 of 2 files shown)


stepanjakl:
I was just testing the visibility modifier used alongside the `x-transition` directive and I can't make it to transition in (using opacity). It just appears without transition, even when the transition duration is defined. Any ideas what could hinder it?

stepanjakl:
To expand, it should be doable to use visibility style property with opacity. Here's an example: https://stackoverflow.com/a/29318542

SimoTod:
The article changes opacity and visibility at the same time while the transition code does that in steps. There must be some conflict with animating opacity when you change visibility.

It's not exactly what the initial request was but it works correctly if you toggle the hidden attribute.

SimoTod:
For transition out, the visibility attribute is only added after the transition has finished so, whatever property you set it will always work. Transition in removes the attribute in the middle of the transition.

stepanjakl:
Okay, that makes sense. Should we attempt to modify the code to work with the visibility property? I tested the hidden attribute, but it seems to work same as applying `display: none;` to the element.

joshhanley:
@stepanjakl thanks for being willing to contribute! Unfortunately, in the interest of keeping Alpine's API as minimal as possible and keeping the maintenance burden to a minimum, we're going to hold off for now on this. Thanks again!

PR #3334, opened 2022-12-11
Bug - Custom prefix not supported in same use cases

SimoTod:
Alpine provides a way to define a custom prefix rather than using `x-` but the internal code itself hardcodes the `x-` version in a lot of places so people using custom prefixes are going to get unexpected errors.
Also, the prefix for `x-on` and `x-bind` is called too early so the short versions are never working with a custom prefix.
Sorry for the huge PR but there were a lot of files to change.

### DIFF
--- a/packages/alpinejs/src/binds.js
+++ b/packages/alpinejs/src/binds.js
@@ -1,4 +1,4 @@
-import { attributesOnly, directives } from "./directives"
+import { attributesOnly, directives, prefix } from "./directives"
 
 let binds = {}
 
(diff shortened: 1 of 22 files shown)


calebporzio:
Ok, first, thank you @SimoTod for this pull request. It's thorough and does it's job well.

Unfortunately, I'm going to close it. I added that internal "prefix" code as sort-of internal trial to leave that door open. But I intentionally didn't document the API and as you can see didn't stick with it internally.

I think I'd like to move away from this extension point eventually, so in a weird way, I'm glad it doesn't really work right now (because that means people aren't relying on it)

So again, thank you, but I'm going to close this.

PR #3336, opened 2022-12-11
Remove Alpine UI references from Alpine core

SimoTod:
Mini PR to tidy up the core package

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -4,7 +4,7 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v1
       - uses: actions/setup-node@v2
         with:
           node-version: '15'
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -4,7 +4,7 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
       - uses: actions/setup-node@v2
         with:
           node-version: '15'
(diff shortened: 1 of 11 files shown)


joshhanley:
@SimoTod dialog tests were failing but it wasn't bound early, so have removed the before for dialog and it's now passing. Popover wasn't bound early either, so I wonder whether we should remove that too?

Entangle is failing but I believe that is unrelated, as it's failing on the main branch.

SimoTod:
HI @joshhanley, I might have been over zealous there. Yeah, if it wasn't in the original list, it doesn't need to be there. I'll tidy up the PR in the next few days.

joshhanley:
@SimoTod ok sweet, thanks!

joshhanley:
@SimoTod FYI it looks like `@entangle` is failing due to a change made on the main branch. But that was made after this branch was made, so it shouldn't have those changes yet. Going to run some tests against this PR to find out why GitHub actions seemingly is using the main branch build of Alpine.

joshhanley:
Ok I've fixed the entangle test failures, they were unrelated to this PR. It turns out when GitHub actions runs the tests on a PR, it actually merges the PR with the main branch and runs the tests against that. So I had to fix the failing tests on main to get them to work here.

But the `popover.spec.js` is consistently failing, not it's normal flaky fail, as I've run the tests on the last commit quite a few times. So yeah I think that one will need to have the `.before` removed too.

SimoTod:
@joshhanley I've updated the PR. I think the popover test has already been randomly failing for a while in several PRs so it may fail again in the future (https://github.com/alpinejs/alpine/pull/3339#issuecomment-1360465213).

joshhanley:
@SimoTod great, thanks! All green now 😁

I'm going to try and give that popover flakyness another crack maybe sometime this week (not on this PR). But it never fails locally which makes it hard to test 😕

calebporzio:
This is great, thanks @SimoTod !

PR #3339, opened 2022-12-14
Accept x-data="true" as a synonym for x-data="{}"

aral:
As per #3337, some frameworks (like htm) convert `<div x-data … />` to `<div x-data="true" … />` whereas Alpine.js currently only accepts the equivalent of `<div x-data="" … />`. This patch (based on https://github.com/alpinejs/alpine/discussions/3337#discussioncomment-4397836), also makes it accept `x-data="true"` to mean data is `{}`.

### DIFF
--- a/packages/alpinejs/src/directives/x-data.js
+++ b/packages/alpinejs/src/directives/x-data.js
@@ -21,7 +21,7 @@ directive('data', skipDuringClone((el, { expression }, { cleanup }) => {
 
     let data = evaluate(el, expression, { scope: dataProviderContext })
 
-    if (data === undefined) data = {}
+    if (data === undefined || data === true) data = {}
 
     injectMagics(data, el)
 


aral:
Looks like it’s causing a UI component test to fail. Afaik those are closed-source, so I don’t think I can help with that. If someone with access to the source can take a look when they get a moment I’d appreciate it. Otherwise, if there’s anything else I can do, please do let me know.

ekwoka:
Technically the ui components are not closed source, they're right there in the repo under package/ui, but a few of them are flaky/incomplete, so if a quick look doesn't seem like this is related, then the ui tests can be ignored for now.

ekwoka (approved):

aral:
> Technically the ui components are not closed source, they're right there in the repo under package/ui, but a few of them are flaky/incomplete, so if a quick look doesn't seem like this is related, then the ui tests can be ignored for now.

Ah, so they’re not the Alpine UI Components from https://alpinejs.dev/components#components (that’s why I was confused; those have a commercial license and these are apparent under MIT but they’re both referred to as UI components under the Alpine name) :)

ekwoka:
Well, yes and no. The documentation for the headless components are all behind that, but the package itself is just right in the repo and npm, just no documentation for it.

joshhanley:
@aral the failing test is `popover.spec.js` which is frequently flaky, so I wouldn't worry about that. 🙂

joshhanley:
@aral thanks for the PR! Could you add a test for this? Thanks!

joshhanley:
@aral Caleb said to merge anyway! Thanks for the PR!

PR #3351, opened 2022-12-21
Fix x-model checkbox weirdness

austenc:
Seems like `x-model` should also use the `skipForClone` function. Normally extra listeners wouldn't cause a problem, but if you're binding to an array (for example, with multiple checkboxes) it causes incorrect values to be bound. Each value is doubled after a clone happens. 

This was discovered while investigating https://github.com/livewire/livewire/pull/5454

I've included a test for this that fails prior to `skipForClone` being added.

### DIFF
--- a/packages/alpinejs/src/directives/x-model.js
+++ b/packages/alpinejs/src/directives/x-model.js
@@ -5,8 +5,9 @@ import { nextTick } from '../nextTick'
 import bind from '../utils/bind'
 import on from '../utils/on'
 import { warn } from '../utils/warn'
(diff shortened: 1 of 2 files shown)


calebporzio:
I'm going to close this because I feel iffy about wrapping the entirety of x-model in `skipDuringClone`.

As it's the event listeners we don't want registered and not the "binding of values" that also happens.

I'm closing for now because I'm trying to keep these repos clean and not in a wip state, but I'm super grateful and feel free to re-open with a different approach (maybe just wrapping the event listening in the skip function?)

Thanks!

PR #3366, opened 2023-01-02
Don't restore cursor during initialization.

tsjason:
Don't restore the cursor's position during initialization in order to avoid the input from automatically taking focus.

### DIFF
--- a/packages/mask/src/index.js
+++ b/packages/mask/src/index.js
@@ -32,10 +32,10 @@ export default function (Alpine) {
                 // - Initializing the mask on the input if it has an initial value.
                 // - Running the template function to set up reactivity, so that
                 //   when a dependency inside it changes, the input re-masks.
-                processInputValue(el)
+                processInputValue(el, false)
             })
         } else {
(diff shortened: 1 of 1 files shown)


calebporzio:
Yes, a test would be ideal. If it's impossible to accomplish with Cypress, then a codepen or something demoing the fix will suffice as well. Thank you!

tsjason:
For some reason, I can't get the tests to pass locally. It gives me `AssertionError: Timed out retrying after 5000ms: Expected to find element: `[alpine-is-ready]`. 

I tried to make a codepen to demonstrate the issue, but a simple example wouldn't do it. I'm not sure what was different about our specific use. If you don't want to merge it without one, that's fine; just close the PR. We are using a fork with the fix in production now.

calebporzio:
Sounds good, I'll merge this. Thanks!

calebporzio:
Actually there are merge conflicts I can't commit to because this PR is off a main branch on your fork. I'm just going to push the fix myself. Thanks

PR #3375, opened 2023-01-09
Update lifecycle.md

cfjedimaster:
Change After to AFTER to match the style used in the previous section.


joshhanley:
@cfjedimaster thanks!

PR #3393, opened 2023-01-27
Fix x-model event listener doubling when cloned

austenc:
### Problem
The `x-model` directive's event listeners are getting registered **twice** when an element is cloned with `Alpine.clone()`.

This happens once via `initTree()` when clone walks the DOM initializing stuff. Secondly, the mutation observer picks up on `x-model` and then registers the listeners again. A couple other things:

### Notes
- Only noticeable when `x-model` is bound to an array (i.e. checkboxes), but seems to happen for any input with `x-model`.
- Affects Livewire in particular, due to it cloning Alpine components when making DOM updates. livewire/livewire#5454.
- Tried writing a test for the event doubling on any input, but doesn't seem like JS has a native way to retrieve a list (or count) of attached event listeners.

### Questions
- This line doesn't feel great... thoughts? https://github.com/alpinejs/alpine/pull/3393/files#diff-a45b9bae1e2842675ba9349395f0347d4f2a53390bcca48d06fdfb4075e69b08R66
- Changes or other tests you'd like to see included?

Had to open a new PR, because the closed one (#3351) wasn't updated with the latest commits.

### DIFF
--- a/packages/alpinejs/src/clone.js
+++ b/packages/alpinejs/src/clone.js
@@ -2,7 +2,7 @@ import { effect, release, overrideEffect } from "./reactivity"
 import { initTree, isRoot } from "./lifecycle"
 import { walk } from "./utils/walk"
 
-let isCloning = false
+export let isCloning = false
 
 export function skipDuringClone(callback, fallback = () => {}) {
     return (...args) => isCloning ? fallback(...args) : callback(...args)
--- a/packages/alpinejs/src/directives/x-model.js
+++ b/packages/alpinejs/src/directives/x-model.js
@@ -63,16 +63,12 @@ directive('model', (el, { modifiers, expression }, { effect, cleanup }) => {
         || modifiers.includes('lazy')
             ? 'change' : 'input'
 
-    let removeListener = {}
-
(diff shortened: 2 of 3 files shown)


joshhanley:
@austenc failing test is unrelated, so I've re-triggered the tests. 

In regards to that line of concern, maybe you could change it to a one liner? And I think it should be an empty callback

```js
// We only want to register the event listener when we're not cloning, since the
// mutation observer handles initializing the x-model directive already when
// the element is inserted into the DOM. Otherwise we register it twice.
let removeListener = isCloning ? () => {} : on(el, event, modifiers, (e) => {
    setValue(getInputValue(el, modifiers, e, getValue()))
})
```

austenc:
@joshhanley yeah, that seems way better. Done!

Thanks also for re-triggering. Pesky focus tests are such a pain 🥵

joshhanley:
@austenc sweet! Yeah they are! I wish we could fix that one up, but it basically never fails locally, so it's hard to debug.

calebporzio:
Thanks @austenc! Looks great!

nuernbergerA:
finally released :tada:

PR #3395, opened 2023-01-28
Add x-on:.body modifier

arcanemachine:
This pull request adds a `.body` modifier to the `x-on` directive.

It is similar to the `.document` and `.window` modifiers, but with a small difference: **This modifier allows you to easily reference the event listener's parent element using a query selector (e.g. `document.querySelector('body')`).**

### Why?

The use case that inspired this change was that I wanted to dispatch an event from the server side using Phoenix LiveView, which requires a CSS selector as its target.

Although using the `.document` modifier _appears_ to attach the listener to the `<html>` element (at least according to the Chrome DevTools' 'Event Listener' panel), dispatching an event like so does not produce the desired result:

`document.querySelector('html').dispatchEvent(new CustomEvent('hello-world'))`

However, with the new `.body` modifier, you can now create 'global' event listeners that can be targeted by their query selector:

`document.querySelector('body').dispatchEvent(new CustomEvent('hello-world'))`

### Demo

CodePen demo here: https://codepen.io/arcanemachine/pen/XWBYxvg

---

P.S. There is a flaky Cypress test (#3394) that may cause the tests to fail, but it is unrelated to my pull request.

### DIFF
--- a/packages/alpinejs/src/utils/on.js
+++ b/packages/alpinejs/src/utils/on.js
@@ -18,6 +18,7 @@ export default function on (el, event, modifiers, callback) {
     if (modifiers.includes('capture')) options.capture = true
     if (modifiers.includes('window')) listenerTarget = window
     if (modifiers.includes('document')) listenerTarget = document
+    if (modifiers.includes('body')) listenerTarget = document.body
     if (modifiers.includes('prevent')) handler = wrapHandler(handler, (next, e) => { e.preventDefault(); next(e) })
     if (modifiers.includes('stop')) handler = wrapHandler(handler, (next, e) => { e.stopPropagation(); next(e) })
     if (modifiers.includes('self')) handler = wrapHandler(handler, (next, e) => { e.target === el && next(e) })
--- a/packages/docs/src/en/directives/on.md
+++ b/packages/docs/src/en/directives/on.md
@@ -188,6 +188,13 @@ Adding `.window` to listeners is extremely useful for these sorts of cases where
 
 `.document` works similarly to `.window` only it registers listeners on the `document` global, instead of the `window` global.
 
+<a name="body"></a>
+### .body
+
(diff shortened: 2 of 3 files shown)


joshhanley:
@arcanemachine yeah if the failure is popover.spec.js you can ignore that as it’s frequently flaky. 

I’ll can re-trigger tests to run if they fail.

In regards to the PR, all looks good to me! But I don’t follow the part about Phoenix and needing a css selector (don’t use phoenix so could just be me 😅). Are you able to show some code of how you dispatch the event on the phoenix end and how it uses the selector?

arcanemachine:
Sure, here's a contrived example (the important part is the `JS.dispatch` function at the end):

```elixir
  def handle_event("update_content", _data, socket) do
    # business logic skipped for brevity

    # trigger success toast event
    {:noreply, socket |> JS.dispatch("update-content-success", to: "body"), detail: %{item_id: }}
  end
```

Now, on the client side, I can trigger a global-ish type of event. Say, popping up a toast message, or something like that. Something that's not necessarily tied to a specific element, but to the app's state as a whole.

That's how I think this whole concept could be useful.

ekwoka:
I'd recommend just using `document.body` since it's already available with no query.

arcanemachine:
> I'd recommend just using `document.body` since it's already available with no query.

I just posted an example that shows why I need to use a selector. It's basically a workaround for the limitations of LiveView's syntax (which demands a selector).

ekwoka:
if the listener is on the window or document, anything dispatched on the body could bubble...

arcanemachine:
@ekwoka I'm sure it does. My original example was using the wrong function. LiveView events target the window object by default.

I'm going to bail on this one since I was totally wrong and I don't need the workaround that I thought I did. I'll be back if I can find a genuine need for this one.

PR #3398, opened 2023-01-30
Fixes @keyup and @keydown listeners with .capture modifier never executed

mgschoen:
When added to a `@keyup` or `@keydown` event listener, the `.capture` modifier is currently interpreted as a keycode, e.g. the listener is never executed.

Reproduction of the current bug: https://codepen.io/mgschoen/pen/PoBBVMX

This PR
- fixes that `.capture` also works with `keyup` and `keydown` events
- extends `.capture` tests accordingly
- adds docs for `.capture`

### DIFF
--- a/packages/alpinejs/src/utils/on.js
+++ b/packages/alpinejs/src/utils/on.js
@@ -104,7 +104,7 @@ function isKeyEvent(event) {
 
 function isListeningForASpecificKeyThatHasntBeenPressed(e, modifiers) {
     let keyModifiers = modifiers.filter(i => {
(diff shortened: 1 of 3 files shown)


mgschoen:
The failing test was previously [mentioned to be flaky](https://github.com/alpinejs/alpine/pull/3339#issuecomment-1360465213), so I'm assuming I shouldn't worry about the failure?

joshhanley:
@mgschoen thanks for the PR! Yeah that popover test is flaky, I've triggered tests to re-run but I wouldn't be too concerned about it.

joshhanley:
@mgschoen thanks!

PR #3408, opened 2023-02-08
:bug: fixes improper use of nullish coalescing

ekwoka:
`Array.prototype.indexOf` returns `-1` when an item is not found, which would never utilize the nullish coalescence.

To satisfy the apparent intent, this instead simply uses a ternary in the target index of splice

@SimoTod

### DIFF
--- a/packages/alpinejs/src/directives.js
+++ b/packages/alpinejs/src/directives.js
@@ -27,11 +27,8 @@ export function directive(name, callback) {
                 );
                 return;
             }
-            const pos = directiveOrder.indexOf(directive)
-                ?? directiveOrder.indexOf('DEFAULT');
-            if (pos >= 0) {
-                directiveOrder.splice(pos, 0, name);
-            }
+            const pos = directiveOrder.indexOf(directive);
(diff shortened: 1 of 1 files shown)


SimoTod (approved):
Makes sense, thanks

andrewpitts15:
@calebporzio Is this going to be merged in? I have an alpine project that has to run in an app with CEF using an old Chrome version (76) that doesn't support nullish coalescing that the existing code is breaking.

SimoTod:
You can use the previous tag while you are waiting for the next release.
https://cdn.jsdelivr.net/npm/alpinejs@3.10.5/dist/cdn.min.js

joshhanley:
@ekwoka thanks!

ekwoka:
@andrewpitts15 in the future, you can also bundle Alpine yourself in your project and use a different target. Nullish coalescing is within the bounds of the target for the Alpine distributables, and while this change removes it, I don't feel there is any reason to believe another one won't be introduced at some point in the future.

Nullish coalescing is transpilable, so your own tooling can handle making it backwards compatible.

PR #3409, opened 2023-02-08
Feature/UI/combobox

SimoTod:


joshhanley:
@SimoTod I noticed when you merged the main branch in, it has added all the extra commits in the history and if you look at files changed it has all those extra files. It's also tagged all those contributors.

I think we should actually merge the `main` branch into the `jlb/combobox` branch instead of into your PR to prevent all those changes showing up.

I can sort that out if you agree. If so, can you remove the merge commit from your branch and force push it, and then I will merge `main` into the `jlb/combobox` branch so it's up to date?

SimoTod:
> @SimoTod I noticed when you merged the `main` branch in, it has added all the extra commits in the history and if you look at files changed it has all those extra files. It's also tagged all those contributors.
> 
> I think we should actually merge the `main` branch into the `jlb/combobox` branch instead of into your PR to prevent all those changes showing up here.
> 
> I can sort that out if you agree. If so, can you remove the merge commit from your branch and force push it, and then I will merge `main` into the `jlb/combobox` branch so it's up to date?

sorted

PR #3414, opened 2023-02-11
Update index.js

Restartz:
fix: mask on initial x-model values

https://github.com/alpinejs/alpine/discussions/2863#discussioncomment-4942036

### DIFF
--- a/packages/mask/src/index.js
+++ b/packages/mask/src/index.js
@@ -32,10 +32,14 @@ export default function (Alpine) {
                 // - Initializing the mask on the input if it has an initial value.
                 // - Running the template function to set up reactivity, so that
                 //   when a dependency inside it changes, the input re-masks.
-                processInputValue(el)
+                queueMicrotask(() => {
+                    processInputValue(el)
+                })
             })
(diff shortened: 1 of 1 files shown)


ekwoka:
Is there a way for you to add a test for this?

Restartz:
> Is there a way for you to add a test for this?

Looking at other pull requests I think this should be it:
```
test('mask initial x-model value',
    html`<input x-data="{ amount: 132558 }" x-mask:dynamic="$money($input)" x-model="amount">`,
    ({ get }) => { get('input').should(haveValue('132,558')) }
)
```

But I don't really know where to add it.

Restartz:
I think I found it. I've added a test the `tests/cypress/integration/plugins/mask.spec.js` file, also changed it a bit to fit in more with the other tests in that file.

calebporzio:
This makes sense, only hesitancy with this is `effect()` will no longer track reactive dependancies used in `templateFn` now that we've introduced `queueMicrotask`.

Can you verify this isn't a problem?

PR #3415, opened 2023-02-11
x-mask custom decimal precision

Restartz:
Allows for custom decimal precision using x-mask $money magic as requested here: https://github.com/alpinejs/alpine/discussions/2873

Example usage:
- To allow for a decimal precision of 3 digits: `<input x-data x-mask:dynamic="$money($input, '.', ',', 3)">`
- To only allow for round numbers: `<input x-data x-mask:dynamic="$money($input, '.', ',', 0)">`

It defaults to 2, so the current behaviour without specified precision should still be the same:
`<input x-data x-mask:dynamic="$money($input)">`

### DIFF
--- a/packages/docs/src/en/plugins/mask.md
+++ b/packages/docs/src/en/plugins/mask.md
@@ -171,3 +171,16 @@ You may also choose to override the thousands separator by supplying a third opt
     <input type="text" x-mask:dynamic="$money($input, '.', ' ')"  placeholder="3 000.00">
 </div>
 <!-- END_VERBATIM -->
+
+
+You can also override the default precision of 2 digits by using any desired number digits as the fourth optional argument:
+
+```alpine
+<input x-mask:dynamic="$money($input, '.', ',', 8)">
+```
+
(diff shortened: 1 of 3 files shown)


ekwoka:
Looks good. Maybe at this point it needs to be made into an options object?

Restartz:
I've changed the $money helper to accept an options object as the second parameter.
So precision can now be set as follows: 
`$money($input, { precision: 3 })`

The delimiter and thousands params can also be passed trough the options object: 
`$money($input, { delimiter: ',', precision: 0 })`

Calls like `$money($input, ',')` or `$money($input, ',', ' ')` also still work.

calebporzio:
Hey @Restartz, thanks for this. I actually prefer the 4th param instead of the options object. Can you remove all the options object stuff and we'll merge? Thanks!

Oh, and can you also add documentation for this?

Restartz:
@joshhanley Done and done! 🙂

joshhanley:
@Restartz thanks!

PR #3436, opened 2023-02-24
Fix the x-mask focus issue in Safari browser

orchidcode:
Solves the problem in Safari browsers with x-mask setting initial focus on the wrong input element on page when page loads.

### DIFF
--- a/packages/mask/src/index.js
+++ b/packages/mask/src/index.js
@@ -58,7 +58,7 @@ export default function (Alpine) {
 
             let setInput = () => { lastInputValue = el.value = formatInput(input, template) }
 
-            if (shouldRestoreCursor) {
+            if (shouldRestoreCursor && el.activeElement) {
                 // When an input element's value is set, it moves the cursor to the end
                 // therefore we need to track, estimate, and restore the cursor after
                 // a change was made.


joshhanley:
@orchidcode thanks for the PR! It seems your changes have broken the mask tests. Any chance you can have a look at that? Thanks!

orchidcode:
> Also are you able to provide some more details on the issue you are solving or even a codepen that demonstrates the issue so we can confirm it?

The issue is only with Safari browser. When you use x-mask plugin on some fields, Safari focuses one of them on page load. So if you have a form with 10 fields and only the 6th uses x-mask, it will grab focus on page load.

You may replicate this behavior if you use Safari on your docs page for x-mask: https://alpinejs.dev/plugins/mask – on page load, the second example gets focus on an input field. Have a look also at https://github.com/alpinejs/alpine/discussions/2890 thread.

Unfortunately, I am not good enough to find better solution for now.

joshhanley:
@orchidcode ok thanks for the details 🙂

You will need to ensure that any changes you make don't break the current test suite. I'd suggest making sure you can run the tests locally, so as you make changes you can review if everything is still working as expected.

joshhanley:
@orchidcode thanks for being willing to contribute! We're going to close this for now, but if you manage to fix the issue with the existing tests, please resubmit the PR 🙂

Jackky90:
There should be `document.activeElement === el` to be correct. Cuz `activeElement` property is unique to the `document` object.

PR #3438, opened 2023-02-25
Tidy up play files in root directory

jameschensmith:
#### Summary

Removes play files that were in the root directory. An alternative could be to move these into another directory, such as `examples/`. Open to suggestions ☺️ 

Relates to #3434.


joshhanley:
@jameschensmith thanks for being willing to contribute. Caleb said he uses these files, so wants to just leave them where they are 🙂 so going to close this.

PR #3449, opened 2023-03-07
Replace brotli-size with zlib

jameschensmith:
#### Summary

Replaces `brotli-size` with `node:zlib`. Tested, and both produce the exact same output.

#### Why?

> [`brotli-size` is] only used to indicate during building the current size of the bundled packages, so it has no actually effect.
> 
> `brotli-size` was a convenience package for compressing and displaying the size, while nodejs `zlib` is a core module that can brotli compress and then we can easily get the buffer size from there without involving third party code.
> 
> Here it looks like it's even an exact API match. So its just knocking off an unneeded dev dependency.

&mdash; @ekwoka ([source](https://github.com/alpinejs/alpine/pull/3449#issuecomment-1457698301))

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,6 @@
             },
             "devDependencies": {
                 "axios": "^0.21.1",
-                "brotli-size": "^4.0.0",
                 "chalk": "^4.1.1",
                 "cypress": "^7.0.0",
                 "cypress-plugin-tab": "^1.0.5",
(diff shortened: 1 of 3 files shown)


joshhanley:
@jameschensmith thanks for the PR(s)! Can you make sure to add a “why” and some more details to your descriptions? I personally don’t understand why this change is being made (because I have no idea what either library does without looking 😆), so would be good to read your reasoning and makes it easier for maintainers to merge if all relevant information is in the PR 🙂 thanks!

ekwoka:
@joshhanley In this case, it's only used to indicate during building the current size of the bundled packages, so it has no actually effect.

brotli-size was a convenience package for compressing and displaying the size, while nodejs `zlib` is a core module that can brotli compress and then we can easily get the buffer size from there without involving third party code.

Here it looks like it's even an exact API match. So its just knocking off an unneeded dev dependency.

jameschensmith:
My apologies! Yes, it's exactly what @ekwoka has said. I'll put that in the description (I hope you don't mind if I quote you, @ekwoka 😅).

joshhanley:
@jameschensmith no worries! Ah makes sense, thanks @ekwoka

calebporzio:
Sounds good, thanks!

PR #3459, opened 2023-03-12
Exclude dependencies from non-CDN builds

jameschensmith:
#### Summary

Marks packages as "external" for non-CDN builds. Currently, if you install `v3.12.0` of alpinejs, you will pull down `@vue/reactivity` and it's dependency into `node_modules` while also having it bundled within alpinejs. This PR excludes those dependencies only from the ESM / CJS bundles.

See somewhat related discussion / recommendation [here](https://github.com/alpinejs/alpine/discussions/3434).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7,9 +7,6 @@
             "workspaces": [
                 "packages/*"
             ],
-            "dependencies": {
-                "nprogress": "^0.2.0"
-            },
             "devDependencies": {
                 "axios": "^0.21.1",
                 "chalk": "^4.1.1",
@@ -4933,7 +4930,8 @@
         },
         "node_modules/nprogress": {
             "version": "0.2.0",
-            "license": "MIT"
+            "resolved": "https://registry.npmjs.org/nprogress/-/nprogress-0.2.0.tgz",
+            "integrity": "sha512-I19aIingLgR1fmhftnbWWO3dXc0hSxqHQHQb3H8m+K3TnEn/iSeTZZOyvKXWqQESMwuUVnatlCnZdLBZZt2VSA=="
         },
         "node_modules/nwsapi": {
             "version": "2.2.0",
@@ -7104,7 +7102,10 @@
         "packages/navigate": {
             "name": "@alpinejs/navigate",
             "version": "3.10.2",
-            "license": "MIT"
+            "license": "MIT",
+            "dependencies": {
+                "nprogress": "^0.2.0"
+            }
         },
         "packages/persist": {
             "name": "@alpinejs/persist",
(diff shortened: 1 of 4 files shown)


ekwoka:
I'm all about this

joshhanley:
@jameschensmith thanks for the PR! Makes sense to me.

Can you explain why `nprogress` was moved from the base `package.json` to `packages/navigate/package.json` in this PR? Is it related to the external dependency changes?

jameschensmith:
> @jameschensmith thanks for the PR! Makes sense to me.
> 
> Can you explain why `nprogress` was moved from the base `package.json` to `packages/navigate/package.json` in this PR? Is it related to the external dependency changes?

@joshhanley, no problem! Yes, it is related to the changes. Because the non-CDN builds no longer have the dependencies bundled with them, `@alpinejs/navigate` needs to have this dependency defined in its `package.json` because this package is used in `packages/navigate/src/bar.js`. Let me know if you have any more questions and I'll do my best to answer them 😊

joshhanley:
@jameschensmith ok, great! Thanks for that 🙂

RobertBoes:
This change seems to cause issues in our project, after upgrading `alpinejs` to `3.12.1` we're getting the following error:
```sh
ERROR in ./node_modules/alpinejs/dist/module.esm.js 1562:0-88
Module not found: Error: Can't resolve '@vue/reactivity' in '/[path-to-project]/node_modules/alpinejs/dist'
```
Even adding `"@vue/reactivity": "^3.3.1",` to our own project doesn't solve the issue and still gives the same error. The weird thing is looking in the `node_modules/alpinejs` folder I do see `@vue/reactivity` is installed in `node_modules/alpinejs/node_modules/@vue/reactivity`

ekwoka:
That's quite strange. I'd try deleting node modules and package-lock and reinstalling

ekwoka:
What bundler are you using?

RobertBoes:
Removing the node_modules had no effect. We're using Laravel Mix (Webpack) and the `"type": "module",` is only really recently introduced/become the standard in fresh Laravel apps.

But if this change would require one to have `"type": "module",` in their `package.json` it seems like a big breaking change to me

ekwoka:
It doesn't and wouldn't. Although, many would say using webpack and cjs was already broken.

what is important is the module resolution, not the type of module.

Basically that it should walk up the tree checking for `node_modules` folders when the path isn't relative.

But if you are not using type module, and it is still going to `esm.js` that may be an issue, but most likely with your system, since it should be resolving the cjs module.

ekwoka:
That would do it!!!!

This is why it's most common to have your aliases be `@/`

Igloczek:
@calebporzio can you elaborate on what were the issues that made you revert those changes, besides the aliases conflicts in @RobertBoes project?

i'd like to bring this PR to life, since it's a very non-npm way of doing things to ship bundled dependencies in npm packages, since it's kinda like shipping CDN-ish build, but to NPM, it makes updating the underlying deps impossible without falling back to use `/src` from the packages and figuring out what is missing to run them, for example missing `ALPINE_VERSION` global variable, that is set only during the build

and because of this unwanted bundling exists, some packages (`anchor` and `sort`) don't even have deps listed in the right place, since it hides the problem, because it can "borrow" them from the monorepo root - related PR #4550

Issue #383, opened 2020-04-15
$watch doesn't work with Array.prototype.push


FYITom:
CodePen show's the issue, check the console.log when toggling filters.

https://codepen.io/TomS-/pen/NWGxdzv

ryangjchandler:
Hi @FYITom, I'd probably use `this.selected = this.selected.concat([item])` instead of `this.selected.push(item)`. The behaviour of `.push()` is a bit temperamental with the reactivity core and the Proxy won't always react to changes. Explicitely setting a property will definitely work.

I modified your CodePen here (https://codepen.io/ryangjchandler/pen/rNOxyLE?editors=1111) with the changes and the watcher is now reacting.

If you're okay with this, then please close the issue. If you want some further help, just ask, there's always someone here!

HugoDF:
Is this a `$watch` issue? Isn't it a generic reactivity issue on array mutation methods?

ryangjchandler:
> Is this a `$watch` issue? Isn't it a generic reactivity issue on array mutation methods?

Yeah exactly. I've had trouble with `.push()` before, without the `$watch`er.

SimoTod:
`$watch` triggers a function when the value changes (note that, when dealing with objects and arrays, the value is just a reference to the object so $watch only works if you reassign the variable as Ryan suggested).
At the moment, Alpine does not support 'deep' watchers (They would trigger when something inside the object or array changes at any level).
There is an initial PR for it #294 but it hasn't been reviewed yet.

ryangjchandler:
> `$watch` triggers a function when the value changes (note that, when dealing with objects and arrays, the value is just a reference to the object so $watch only works if you reassign the variable as Ryan suggested).
> At the moment, Alpine does not support 'deep' watchers (They would trigger when something inside the object or array changes at any level).
> There is an initial PR for it #294 but it hasn't been reviewed yet.

Yeah exactly that. I think the PR could be tidied up a little before being reviewed / merged too.

FYITom:
Thank you very much for that, makes complete sense to me.

pohlt:
I had the same issue with a `Set`. Adding or deleting an item does not trigger a watch or set/unset a class in an `x-bind:class`. After changing the set, I did a `selected = new Set(selected)` which triggers any updates.

SimoTod:
Deep watchers are not supported yet. The watcher triggers only if you reassign the variable but not if you change it's internal state.

stuartjnelson:
[...]

PR #3463, opened 2023-03-14
Add modifier `defer` to control deferring storage

hirasso:
The persist plugin right now stores the value immediately on initialization. This prevents me from being able to use it in quite a few  cases, where nothing should be stored in the browser before user interaction (GDPR compliance).

This PR attempts to solve this problem by adding a new modifier `$persist('my-value').defer()` that will wait for the first mutation of the data before storing it in the browser.

In my (manual) tests, it's already working fine for primitives, but for arrays and objects there seems to be something I'm missing (cypress tests are failing for arrays/objects right now). 

If anyone from the maintainers thinks this could be a useful feature, I would be very grateful if they could have a look at why this fails with arrays. (Something inside `Alpine.effect()`, I guess...)

Actually, I'll add a failing test for arrays, so that it's easier for you guys.

### DIFF
--- a/packages/persist/src/index.js
+++ b/packages/persist/src/index.js
@@ -8,8 +8,6 @@ export default function (Alpine) {
         return Alpine.interceptor((initialValue, getter, setter, path, key) => {
             let lookup = alias || `_x_${path}`
 
-            console.log('defer:', defer);
-
             let initial = storageHas(lookup, storage)
                 ? storageGet(lookup, storage)
                 : initialValue
--- a/packages/persist/src/index.js
+++ b/packages/persist/src/index.js
@@ -2,10 +2,13 @@ export default function (Alpine) {
     let persist = () => {
         let alias
         let storage = localStorage
+        let defer = false
         let isInitial = true
 
         return Alpine.interceptor((initialValue, getter, setter, path, key) => {
             let lookup = alias || `_x_${path}`
+
+            console.log('defer:', defer);
 
             let initial = storageHas(lookup, storage)
                 ? storageGet(lookup, storage)
@@ -16,7 +19,7 @@ export default function (Alpine) {
             Alpine.effect(() => {
                 let value = getter()
 
-                if (!isInitial) storageSet(lookup, value, storage)
+                if (!defer || !isInitial) storageSet(lookup, value, storage)
 
                 setter(value)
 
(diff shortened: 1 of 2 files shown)


hirasso:
The persist plugin right now stores the value immediately on initialization. This prevents me from being able to use it in quite a few  cases, where nothing should be stored in the browser before user interaction (GDPR compliance).

This PR attempts to solve this problem by adding a new modifier `$persist('my-value').defer()` that will wait for the first mutation of the data before storing it in the browser.

In my (manual) tests, it's already working fine for primitives, but for arrays and objects there seems to be something I'm missing (cypress tests are failing for arrays/objects right now). 

If anyone from the maintainers thinks this could be a useful feature, I would be very grateful if they could have a look at why this fails with arrays. (Something inside `Alpine.effect()`, I guess...)

Actually, I'll add a failing test for arrays, so that it's easier for you guys.

hirasso:
Maybe a pointer to a solution: The test passes if I use `concat` instead of `push`.

This fails:

```js
test.only('can defer persisting an array',
    [html`
        <div x-data="{ things: $persist(['foo', 'bar']).defer() }">
            <button @click="things.push('baz')"></button>

            <span x-text="things.join('-')"></span>
        </div>
    `],
    ({ get }, reload) => {
        get('span').should(haveText('foo-bar'))
        get('button').click()
        get('span').should(haveText('foo-bar-baz'))
        reload()
        get('span').should(haveText('foo-bar-baz'))
    },
)
```


This passses:

```js
test.only('can defer persisting an array',
    [html`
        <div x-data="{ things: $persist(['foo', 'bar']).defer() }">
            <button @click="things = things.concat(['baz'])"></button>

            <span x-text="things.join('-')"></span>
        </div>
    `],
    ({ get }, reload) => {
        get('span').should(haveText('foo-bar'))
        get('button').click()
        get('span').should(haveText('foo-bar-baz'))
        reload()
        get('span').should(haveText('foo-bar-baz'))
    },
)
```

Maybe this is somehow related to #383 ?

hirasso:
That's exactly the thing. Strictly speaking, storing the data before mutating it is not technically necessary. 

In my opionion, the behavior introduced here should be the default behavior. I only added the modifier to make it opt-in and prevent breaking changes (consumers might already rely on the local storage entry being present from the beginning...)

What do you think?

hirasso:
You are right - it wouldn't be possible to get the user's permission before persisting the value right now. I'll close this. Anyways, nice little exercise for the history plugin :)

PR #3468, opened 2023-03-15
Document the built-in `destroy()` lifecycle method

arcanemachine:
I recently discovered that Alpine has a built-in (but undocumented) `destroy()` lifecycle method, as discussed [here](https://github.com/alpinejs/alpine/discussions/3420).

Having used it for personal purposes, I have found the method to work as expected (cleanup logic is executed before the element is removed from the DOM). It even [has tests already written for it](https://github.com/alpinejs/alpine/blob/main/tests/cypress/integration/custom-data.spec.js#L144)!

This pull request contains a bit of documentation explaining the use of this method.

### DIFF
--- a/packages/docs/src/en/essentials/lifecycle.md
+++ b/packages/docs/src/en/essentials/lifecycle.md
@@ -28,6 +28,23 @@ Alpine.data('dropdown', () => ({
 }))
 ```
 
+<a name="element-destruction"></a>
+## Element destruction
+
+The `destroy()` method, if present, is called when an element is removed from the DOM.
+
+Alpine will automatically call any `destroy()` methods stored on a data object. For example:
+
+```js
+Alpine.data('dropdown', () => ({
+    destroy() {
+      // I get called before the element using this data is removed from the DOM.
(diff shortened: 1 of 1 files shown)


joshhanley:
@arcanemachine thanks for the PR! There has been a PR recently also wanting to document the `destroy()` method, see it here PR #2717.

I'm closing this for the same reason we closed that one, which is, destroy() is undocumented due to not being as reliable as we'd like and that there should also probably be an x-destroy directive with comprehensive tests around all of it.

Thanks for being willing to contribute!

arcanemachine:
I apologize, I did not check for prior commits (rookie mistake).

You mentioned that the feature currently isn't reliable. Would you be able to describe the issues that exist with the current implementation? I see that [this pull request](https://github.com/alpinejs/alpine/pull/2733) mentions that the `destroy()` callback does not have access to the current scope. Is there anything else?

joshhanley:
@arcanemachine honestly, I'm not sure, I was just passing on Caleb's feedback on that PR.

PR #3473, opened 2023-03-17
Make x-transition delay syntax consistent with duration syntax

trych:
Hey there,

I noticed that with x-transition.duration it was possible to use a `x-transition.duration.500ms` syntax as well as `x-transition.duration.500` (as also noted in the source code). For `x-transition.delay` only the `ms` syntax was possible. This PR makes the delay syntax consistent with the duration syntax, so both work with or without the `ms` suffix.

(My first Alpine.js PR, so please let me know if something is wrong)


joshhanley:
@trych thanks for the PR! Makes sense to me. Could you please add tests? Thanks!

trych:
Ok, will close this PR and open another one that submits from a different branch (could not figure out, if I can change the source branch after the fact ...)

PR #3481, opened 2023-03-23
throttle/debounce user callback after preventDefault/stopPropagation is executed

hudon:
This fixes the issue outlined here: https://github.com/alpinejs/alpine/discussions/3478

There is a tradeoff worth noting here and perhaps a breaking change. This fix implies that if a user types `dragover.prevent.throttle`, they do intend the dragover events to _all_ preventDefault(), even the ones that end up not firing the callback because they're getting throttled. If some users however are depending on the current behavior and expecting the entire wrapped callback to be throttled (stopProp/preventDef included), this patch will break their use case. It would be inconsistent behavior, though, to have the event sometimes propagate and sometimes not.

_repost of https://github.com/alpinejs/alpine/pull/3479 from feature branch_

### DIFF
--- a/packages/alpinejs/src/utils/on.js
+++ b/packages/alpinejs/src/utils/on.js
@@ -21,7 +21,7 @@ export default function on (el, event, modifiers, callback) {
 
     // By wrapping the handler with debounce & throttle first, we ensure that the wrapping logic itself is not
     // throttled/debounced, only the user's callback is. This way, if the user expects
-    // `e.preventDefault()` to happen, it'll still happen even if the callback gets throttled.
+    // `e.preventDefault()` to happen, it'll still happen even if their callback gets throttled.
     if (modifiers.includes('debounce')) {
         let nextModifier = modifiers[modifiers.indexOf('debounce')+1] || 'invalid-wait'
         let wait = isNumeric(nextModifier.split('ms')[0]) ? Number(nextModifier.split('ms')[0]) : 250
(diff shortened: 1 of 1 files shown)


packages/alpinejs/src/utils/on.js
```diff
@@ -18,6 +18,23 @@ export default function on (el, event, modifiers, callback) {
...
+    // `e.preventDefault()` to happen, it'll still happen even if their callback gets throttled.
+    if (modifiers.includes('debounce')) {
+        let nextModifier = modifiers[modifiers.indexOf('debounce')+1] || 'invalid-wait'
+        let wait = isNumeric(nextModifier.split('ms')[0]) ? Number(nextModifier.split('ms')[0]) : 250
+
+        handler = debounce(handler, wait)
+    }
+    if (modifiers.includes('throttle')) {
+        let nextModifier = modifiers[modifiers.indexOf('throttle')+1] || 'invalid-wait'
+        let wait = isNumeric(nextModifier.split('ms')[0]) ? Number(nextModifier.split('ms')[0]) : 250
```
ekwoka, line 33: This isn't yours, but I think just

```
Number(nextModifier.replace('ms','')) || 250
```
would be more efficient and get the same job done.

ekwoka (approved):

packages/alpinejs/src/utils/on.js
hudon: good catch, reads better too. I've made the change and confirmed existing tests cover this

joshhanley:
@hudon thanks for the PR! Can you revert your last commit? I think that should be changed in a separate PR, that way if we have to revert this PR (or the new PR) if any changes cause issues, then we're not having to revert everything. Thanks!

hudon:
@joshhanley done

joshhanley:
@hudon sweet, thanks!

PR #3486, opened 2023-03-25
Respect reduced motion preference

ekwoka:
Solves #3484 

This only adds transition behaviors when the browser does not match `prefers-reduced-motion`

Should this handle the browser changing back and forth (this code currently does) or only check on initial registration of the directive?

I couldn't get the cypress test to really work. cypress does not make it easy to modify media queries, and the transitionend event was very flaky. I validated this separately in a local sandbox.

Might be able to get tests to work with `https://github.com/bahmutov/cypress-cdp` plugin.

### DIFF
--- a/packages/alpinejs/src/directives/x-transition.js
+++ b/packages/alpinejs/src/directives/x-transition.js
@@ -5,29 +5,68 @@ import { directive } from '../directives'
 import { mutateDom } from '../mutation'
 import { once } from '../utils/once'
 
-directive('transition', (el, { value, modifiers, expression }, { evaluate }) => {
-    if (typeof expression === 'function') expression = evaluate(expression)
+directive(
+    'transition',
+    (el, { value, modifiers, expression }, { evaluate, cleanup }) => {
+        if (typeof expression === 'function') expression = evaluate(expression)
 
-    if (! expression) {
-        registerTransitionsFromHelper(el, modifiers, value)
-    } else {
-        registerTransitionsFromClassString(el, expression, value)
+        cleanup(
+            registerTransitionWithReducedMotion(
+                el,
+                expression || modifiers,
+                value
+            )
+        )
     }
-})
+)
+
+const prefersReducedMotion = window.matchMedia(
+    '(prefers-reduced-motion: reduce)'
+)
+
+function registerTransitionWithReducedMotion(el, expressionOrMod, stage) {
+    const handler = ({ matches }) => {
+        if (matches) return removeTransitions(el)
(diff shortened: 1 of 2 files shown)


kslstn:
I sympathize with the intent, but I see some issues with deciding for the developer whether or not to render transitions. I don't fully understand the proposed change, so I'm wondering if it considers these issues.

The system/browser setting refers to 'prefers _reduced_ motion', which was originally created for people who get nausea from things moving on screens. A good reduced motion implementation can include more subtle transitions, f.e. opacity fades instead of things moving on and off the viewport.

Some (of my) implementations rely on things happening at the end of a transition using `@transitionend`. For `prefers-reduced-motion` situations, I've set the transition duration to invisibly short (something like 1 or 10 ms). That way the transition isn't visible, but the `transitionend` event is fired.

ekwoka:
@kslstn So then, perhaps what may really need to happen is for the built in plugins like x-collapse and helper modifiers to have this (as those are opinionated by alpine) and leave custom transitions to the developer to consider? Or even limit the helpers to the scale ones, but not the opacity?

I do agree that this implementation here may be too aggressive. Maybe a config option that can be flipped on `Alpine` before start that can adjust the behavior from `no transitions, only user defined transtions, all transitions` even. 

The transitionend could still be handled with no transition (alpine could still emit a `transitionend` event from the element regardless).

Would be helpful to get some input @joshhanley so I can see which direction this PR should take.

calebporzio:
Yeah, tricky. I can see the desire for something like this.

For starters, PR's like this are easier to review (and more likely to get merged) if they contain the minimal changes necessary: specifically referring to all the code style changes.

Has more than one person requested that Alpine have a story for this? I'm not denying it should, just because it's not an obvious solution, I want to make sure there's enough demand.

I think in the meantime, if it's one or two people, I'd recommend they create their own `x-accessible-transition` directive or something that wraps the original one. That's just a solution off the top of my head.

If there IS sufficient demand, the options are:
A) provide a userland story for people who want this disabled
B) provide a global config option (would be the first)
C) provide a modifier called "reduce-motion" or something that disables transitions if the OS has that set.
D) make it the default behavior

I lean towards "A", then if it's not a clean recommendation and if enough people want it, "C"

Thoughts?

ekwoka:
Sorry, some of the style changes come from the fact whatever settings you use for style/linting aren't included in the repo, so just an errant save changes everything here.

True, a separate directive plugin is a solution. I'll close this and see about if that can be abstracted easily in a separate plugin package of my own.

PR #3494, opened 2023-03-29
:bug: Enabled morphing @event handlers

ekwoka:
fixes #3493 

This PR:
- when needing to set `@event` attributes on elements during morph, transforms them into properly prefixed `x-on:` attributes instead
- Tests the above behavior

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7046,7 +7046,7 @@
             }
         },
         "packages/alpinejs": {
-            "version": "3.11.1",
+            "version": "3.12.0",
             "license": "MIT",
             "dependencies": {
                 "@vue/reactivity": "~3.1.1"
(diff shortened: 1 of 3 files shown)


joshhanley:
@ekwoka in Livewire, we actually monkey patch the `setAttribute` method to allow `@` symbols, see here https://github.com/livewire/livewire/blob/019b1e69d8cd8c7e749eba7a38e4fa69ecbc8f74/js/index.js#L142-L163

We're now using Alpine Morph for Livewire V3 and currently we're also monkey patching it there too.
I wonder whether it's worth adding something like that into the morph plugin instead?

ekwoka:
It would make sense to have it live in Alpine morph. I'd probably not go with the monkey patching native behaviors though.

PR #3498, opened 2023-03-31
feat: Adds descriptive warning for undefined x-for key and updates tests.

nickdawes:
Description:
This pull request addresses the issue where the x-for directive would throw a generic error message when the provided key is undefined or invalid.

The changes include:
Updating the Alpine.js code to throw a more descriptive error message when the key is undefined or invalid in the x-for directive.
Adding a new test case to the Cypress test suite to check for the updated error message when the key is undefined.

Here is the summary of the changes made:
Updated the error message for an undefined or invalid key in the x-for directive.
Added a new test case to the Cypress test suite to check for the updated error message.

Why was this change necessary? 
By providing a more descriptive error message, developers can quickly identify the issue and fix their code accordingly.

Please review the changes and let me know what you think.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7046,7 +7046,7 @@
             }
         },
         "packages/alpinejs": {
-            "version": "3.11.1",
+            "version": "3.12.0",
             "license": "MIT",
             "dependencies": {
                 "@vue/reactivity": "~3.1.1"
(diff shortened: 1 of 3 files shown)


ekwoka:
Is this the correct place for this message to appear? Couldn't it be checking the result of the key evaluation?

PR #3504, opened 2023-04-01
Bug: Updates correct scope when x-for looping over element with x-data

ekwoka:
solves #3447  (demos and discussion exist there)

```html
<template x-for="item in my_array" :key="item.k">
        <div x-data="{ foo: 'bar' }" >
            <div x-text="item.x"></div>
        </div>
    </template>
 ```
 
 When looping over an iterable with x-for, and the first child has an `x-data` scope, if the array changed, now containing new objects but which had the same `key` value, the scope would impropely update.
 
 This fixes that.
 
 ## The cause

The x-for does update the scope of the child, but it naively would update only the newest scope on the element. The problem is that this would be, during the update, the scope created by the `x-data` and not the `x-for`. The effects registered to the looped over object/values would be within the context of the `x-for` provided synthetic scope. This scope would continue to have the initial value.

## The fix

Store reference to the scope provided by `x-for` and use that to directly update the synthetic scope, instead of naively updating the latest scope.

### DIFF
--- a/packages/alpinejs/src/directives/x-for.js
+++ b/packages/alpinejs/src/directives/x-for.js
@@ -166,7 +166,7 @@ function loop(el, iteratorNames, evaluateItems, evaluateKey) {
                 marker.remove()
             })
 
-            refreshScope(elForSpot, scopes[keys.indexOf(keyForSpot)])
+            refreshScope(elForSpot, scopes[keys.indexOf(keyForSpot)], true)
         }
 
         // We can now create and add new elements.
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: packages/alpinejs/src/directives/x-for.js [hunk 1]
```diff
@@ -185,7 +185,15 @@ function loop(el, iteratorNames, evaluateItems, evaluateKey) {
 
             let clone = document.importNode(templateEl.content, true).firstElementChild
 
-            addScopeToNode(clone, reactive(scope), templateEl)
+            let reactiveScope = reactive(scope)
+
+            addScopeToNode(clone, reactiveScope, templateEl)
+
+            clone._x_refreshXForScope = (newScope) => {
+                Object.entries(newScope).forEach(([key, value]) => {
```
ekwoka, line 193: This can actually be optimized to just

```js
Object.assign(reactiveScope[key], newScope)
```

PR #3522, opened 2023-04-10
Feat: create new prefix parameter to magic method $money

pedrobruning:
# Description

Added a new option to $money magic method called prefix, which enables the developer to add a prefix to the input field

## Files

- packages\mask\src\index.js
- tests\cypress\integration\plugins\mask.spec.js
- packages\docs\src\en\plugins\mask.md

## Tests

Tested if the new functionality is working as expected and tested if the other functionalities are not being impacted by the change.

### DIFF
--- a/packages/docs/src/en/plugins/mask.md
+++ b/packages/docs/src/en/plugins/mask.md
@@ -184,3 +184,16 @@ You can also override the default precision of 2 digits by using any desired num
     <input type="text" x-mask:dynamic="$money($input, '.', ',', 4)"  placeholder="0.0001">
 </div>
 <!-- END_VERBATIM -->
+
+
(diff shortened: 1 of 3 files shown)


joshhanley:
@pedrobruning thanks for the PR!

It all looks good, the only issue I can see is if there is negative numbers, such as `-1000`, with your current implementation, it would output `$-1000` which should actually be `-$1000`.

But doing a quick google, it seems other currencies may actually put the minus after the currency symbol. Or even some actually put the currency symbol at the end of the number. https://learn.microsoft.com/en-us/globalization/locale/currency-formatting

So I wonder whether this will still work or whether it is better to just add the currency symbol next to the input rather than as apart of the value?

PR #3532, opened 2023-04-17
`x-model.boolean` modifier

gdebrauwer:
In a project I had a `select` with 2 options: one was value, the other was false. But currently that is not very easy to work with, because alpine.js either just returns a string ('true'/'false' when option values are strings) or an integer (1/0 when option values are integers). This also causes issues when entangling it with a boolean property on a Livewire component. To fix this, I added a `boolean modifier to `x-model`.
```html
<select x-model.boolean="bar">
    <option value="true">yes</option>
    <option value="false">no</option>
</select>
```


packages/alpinejs/src/directives/x-model.js
```diff
@@ -169,6 +187,18 @@ function safeParseNumber(rawValue) {
     return isNumeric(number) ? number : rawValue
 }
 
+function safeParseBoolean(rawValue) {
+    if ([1, '1', 'true', true].includes(rawValue)) {
+        return true
+    }
+
+    if ([0, '0', 'false', false].includes(rawValue)) {
+        return false
```
ekwoka, line 196: I was thinking "How could we do this with only one comparison?" and my brain went to 

```js
Boolean(JSON.parse(rawValue.toString()))
```

But surely that would stupid right?

gdebrauwer: I think that does not work, because if the rawValue is not 1/0/true/false, then we want to just return that raw value. With that code, I don't think that is possible 🤔

ekwoka: Yes, I do see that. To me though, the use case of this modifier seems fundamentally contradictory to ever having it be any other values. I'd rather have it fail early (at a json parsing error) than continue to have the bad value end up somewhere else somehow.

But that's an opinionated design.

gdebrauwer: The `x-model.number` works like that as well so I thought it might be best to let it behave in the same way
https://github.com/alpinejs/alpine/blob/6142a83956685b4ddf67d95cf70a46fa3efe5fdf/packages/alpinejs/src/directives/x-model.js#L166-L170

ekwoka: Makes sense 👍

calebporzio:
Thanks for this PR. I've resurrected it and fixed a small bug and we're good to go. Thanks!

PR #3535, opened 2023-04-18
Handle special binding case for 'checked' and 'selected'

inxilpro:
I ran into an interesting issue with `x-bind` and the `checked` property today. [Here's the Twitter thread](https://twitter.com/inxilpro/status/1648402525966729219) summarizing the issue, and [here's a codesandbox demo](https://codesandbox.io/s/elated-sinoussi-jvdlrv) showing it in action.

Basically, the `value`, `checked` and `selected` properties are special in that they're not kept in sync with their corresponding attributes. That means that the following is perfectly legit javascript:

```js
element.setAttribute('checked', 'checked');
element.checked = false;
```

This will result in a checkbox that is not checked, but has a checked attribute set (which, as far as I can tell, is visually checked but not "technically" checked).

Anyway, 90% of the time this won't come up because both `x-bind:value` and `x-model` have special bind logic for checkbox values that set the property rather than the attribute:

https://github.com/alpinejs/alpine/blob/6142a83956685b4ddf67d95cf70a46fa3efe5fdf/packages/alpinejs/src/utils/bind.js#L54-L58

There are some cases, though, where you want to handle both sides of the binding on your own (in our case, the checkboxes are driven from Algolia instant search state, so changing the checkbox triggers an event that then updates the state). This PR makes that case possible.

### DIFF
--- a/packages/alpinejs/src/utils/bind.js
+++ b/packages/alpinejs/src/utils/bind.js
@@ -22,6 +22,14 @@ export default function bind(el, name, value, modifiers = []) {
         case 'class':
             bindClasses(el, value)
             break;
+        
+        // 'selected' and 'checked' are special attributes that aren't necessarily
+        // synced with their corresponding properties when updated, so both the 
+        // attribute and property need to be updated when bound.
+        case 'selected':
+        case 'checked':
+            bindAttributeAndPropery(el, name, value)
+            break;
 
         default:
             bindAttribute(el, name, value)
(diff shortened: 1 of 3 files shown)


packages/alpinejs/src/utils/bind.js
```diff
@@ -22,6 +22,14 @@ export default function bind(el, name, value, modifiers = []) {
         case 'class':
             bindClasses(el, value)
             break;
+        
+        // 'selected' and 'checked' are special attributes that aren't necessarily
+        // synced with their corresponding properties when updated, so both the 
+        // attribute and property need to be updated when bound.
+        case 'selected':
+        case 'checked':
+            bindAttributeAndPropery(el, name, value)
```
ekwoka, line 31:
```suggestion
            bindAttributeAndProperty(el, name, value)
```

packages/alpinejs/src/utils/bind.js
```diff
@@ -78,6 +86,11 @@ function bindStyles(el, value) {
     el._x_undoAddedStyles = setStyles(el, value)
 }
 
+function bindAttributeAndPropery(el, name, value) {
```
ekwoka, line 89:
```suggestion
function bindAttributeAndProperty(el, name, value) {
```

packages/alpinejs/src/utils/bind.js
inxilpro: Good catch! Thanks :)

PR #3545, opened 2023-04-25
add configurable flag to $persist property definition

stepanjakl:
This pull request adds a configurable flag to the definition of the $persist property. Previously, attempting to redefine the $persist property would throw an error.

Screenshot 2023-04-25 at 12 33 37

I would specifically receive this error when the `alpine:init` event was triggered again in order to initialize newly added Alpine components.

With this change, the property can be redefined without issue. Please review and merge if acceptable.

### DIFF
--- a/packages/persist/src/index.js
+++ b/packages/persist/src/index.js
@@ -27,7 +27,7 @@ export default function (Alpine) {
         })
     }
 
-    Object.defineProperty(Alpine, '$persist', { get: () => persist() })
+    Object.defineProperty(Alpine, '$persist', { get: () => persist(), configurable: true })
     Alpine.magic('persist', persist)
     Alpine.persist = (key, { get, set }, storage = localStorage) => {
         let initial = storageHas(key, storage)


ekwoka:
I think this is a bit of the wrong solution.


Do not call `Alpine.start()` more than once.

Newly added elements with Alpine in them are automatically evaluated,or you can do `Alpine.initTree` on the root element if it was not evaluated upon being added

stepanjakl:
Hi @ekwoka, I know this is not an ideal solution and I would love to find an alternative. However, I am not able to initialize a newly added component with the `initTree()` method as suggested, when the component uses `Alpine.data` (unlike having the inline syntax). I tried it in the past and now as well.

stepanjakl:
I tried both, initializing Alpine.data inside the alpine:init listener and without. No chance unfortunately.

Essentially, this is part of a project that uses Barba.js. The newly added DOM doesn’t seem to get registered by Alpine. That’s why I have to call the alpine:init event again.

stepanjakl:
I think that Alpine still crawls the newly added DOM since I get errors for undefined variables and functions, but the data themselves don’t get registered is my suspicion. Is there anywhere to see (i.e. some global object) which Alpine.data have been registered?

stepanjakl:
After a bit of thinking and testing the aformentioned issue, it's most likely that Alpine crawls the newly added DOM before the `Alpine.data` is registered. I'll do some architectural changes in my projec and probably make use of the [Async Alpine](https://github.com/accudio/async-alpine) package to load the components. I do still think that this PR is valid, as all other plugins don't return errors of this kind.

ekwoka:
Yes, if none of the others do this, then this would also be good to not do it.

Though `start` still shouldn't be called multiple times. Maybe there needs to be a message on THAT, or see if there is a way to make `start` smarter.

To do the architectural changes, you can use `Alpine.mutateDom` or just put `x-ignore` on the newly added things, then run `Alpine.data` then run `Alpine.initTree` or remove the `x-ignore`

PR #3565, opened 2023-05-10
Add warning to prevent Alpine from being loaded twice on the same page

calebporzio:


anantakrishna:
So, it's only a warning, right? Seems so after looking at the changes. Actually, this does not “prevent Alpine from being loaded twice on the same page”. At first reading, the title is a bit confusing.

PR #3597, opened 2023-05-25
:bug: Prevents double bundling of Alpine in UI Package

ekwoka:
Before this, the `@alpinejs/ui` package included inside itself a whole other copy of `alpinejs` since it referenced the internal file, not through the module system. This causes a few issues during actual runtime.

This also fixes the long standing issue where `tabbable` was not actually available to the `focus` plugin.


calebporzio:
Thank you!

PR #3615, opened 2023-06-08
$focus: Add 'onlyTabbables' modifier

valincius:
This PR adds `tabbables` modifier to `$focus` plugin to allow traversing ONLY the tabbable elements.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7825,17 +7825,19 @@
             }
         },
         "packages/alpinejs": {
-            "version": "3.12.0",
+            "version": "3.12.2",
             "license": "MIT",
             "dependencies": {
                 "@vue/reactivity": "~3.1.1"
             }
         },
         "packages/collapse": {
-            "version": "3.12.0",
+            "name": "@alpinejs/collapse",
+            "version": "3.12.2",
             "license": "MIT"
         },
         "packages/csp": {
+            "name": "@alpinejs/csp",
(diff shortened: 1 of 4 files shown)


valincius:
Failing tests are outside the scope of this PR.
If you'd like, I can address these in another PR.

ekwoka:
What's the difference between a tabbable and a focusable?

valincius:
> What's the difference between a tabbable and a focusable?

`tabbable` respects the `tabindex`

More info in their readme: https://github.com/focus-trap/tabbable

SimoTod:
> What's the difference between a tabbable and a focusable?

an input with tabindex = -1 is focusable but not tabbable

@valincius isn't the syntax a bit inconsistent? `focusables` returns the list of focusable elements whilst `tabbables` returns this. I can already see people doing `$focus.tabbables()` and `$focus.tabbable($el)`

valincius:
@SimoTod I went with this syntax to allow for reusing all of the other methods that build off of `all` and thus `focusables` so it wouldn't need duplication. I can see how this would be confusing though.

I will draft up something else.

valincius:
I've updated this to better fit the current api, could I get a review?

calebporzio:
Thanks for contributing!

In order to keep Alpine as focused and maintainable as possible, I'm going to close this pull request.

PR #3617, opened 2023-06-09
Add enhancement for discusssion #3293

puckowski:
* Add watch enhancement for discussion #3293.

### DIFF
--- a/packages/alpinejs/src/magics/$watch.js
+++ b/packages/alpinejs/src/magics/$watch.js
@@ -7,38 +7,25 @@ magic('watch', (el, { evaluateLater, effect }) => (key, callback) => {
 
     let oldValue
 
-    const deepClone = (obj) => {
-        if (typeof obj !== 'object' || obj === null) {
-            return obj; // Return primitives and null as is
-        }
-
-        const clone = Array.isArray(obj) ? [] : {};
-
-        for (let key in obj) {
-            if (obj.hasOwnProperty(key)) {
-                if (typeof obj[key] === 'function') {
-                    clone[key] = new Function('return ' + obj[key].toString())();
-                } else {
-                    clone[key] = deepClone(obj[key]);
-                }
-            }
-        }
-
-        return clone;
+    let deepClone = (obj) => {
+        return JSON.parse(JSON.stringify(obj))
     }
 
     let effectReference = effect(() => evaluate(value => {
-        // JSON.stringify touches every single property at any level enabling deep watching
+        // JSON.stringify touches every single property at any level enabling deep watching...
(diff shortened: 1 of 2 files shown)


puckowski:
CodePen with cdn.min.js from main_3293 forked from https://codepen.io/dinyo/pen/eYVaKgj?editors=1111 https://codepen.io/puckowski/pen/poQJzzv?editors=1111

ekwoka:
I think it could be beneficial to check for and prefer `structuredClone`.

Maybe set this up more as a polyfill type thing.

At script init do something like `const deepClone = globalThis.structuredClone ?? customClone`

So we can prefer the future :)

puckowski:
When ```structuredClone``` encounters a proxy object, it throws a ```DataCloneError``` and fails to clone the object.

I ran into ```DataCloneError``` issues when I was using ```structuredClone``` for https://github.com/alpinejs/alpine/pull/3046

This presented issues as there is no great way to tell if an object is proxied in JavaScript, thus I had to support a simple shallow clone implementation along with ```structuredClone```. There were concerns about the number of lines added with 3046.

Plus, depending on how many browser versions the Alpine user wishes to support, they may need to polyfill ```structuredClone``` (https://caniuse.com/?search=structuredClone). Or, we add the ```structuredClone``` polyfill to Alpine for all users at the expense of bundle size. Again, that may not be ideal in terms of number of lines added.

I think 3617 as is allows users of Alpine to reliably get the old and new value with $watch with few lines of code added while passing tests.

calebporzio:
Thanks for the PR - I'm wondering if we can just use the simple JSON.parse(JSON.stringify(...)) clone method

It obv won't keep functions around, but I think it's ok to have that as a limitation when accessing "old" right?

ekwoka:
Is there risk of circular references?

calebporzio:
Thanks for contributing!

In order to keep this project as focused and maintainable as possible, I'm going to close this pull request.

Suall1969:
This PR and #3046 were both closed, but the relevant documentation section on [Deep watching](https://alpinejs.dev/magics/watch#deep-watching) still falsely states that the old value is accessible.
It turns out you can access the old value if you define a getter that clones the property and watch that instead. This way the nuances of the cloning algorithm can be altered per requirements, though in most cased a shallow clone would suffice:
```js
<div x-data="{ foo: { bar: 'baz' }, get fooCopy() { return { ...this.foo } }}" x-init="$watch('fooCopy', (value, oldValue) => console.log(value, oldValue))">
    <button @click="foo.bar = 'bob'">Update</button>
</div>
```

PR #3624, opened 2023-06-17
:sparkles: Adds x-for.hydrate

ekwoka:
This pr adds `x-for.hydrate` to help with hydrating server rendered loop results, instead of generating new ones, and includes accompanying docs and tests.

Here's the updated docs explaining how it works, since I don't think restating it makes much sense.

Some RFC:
- Should non-hydrated elements found (those with `key` attributes not present in the looped data) be removed or simply left in place?
- Should the `key` just be `:key` to line up with the `x-for` `:key` naming?

## Hydration

A common issue with server rendering templates and wanting to take advantage of Alpines loop handling behavior (to hook in sorting and filtering after the fact) is deciding what should be static markup and what should be handled by Alpine. Fortunately, Alpine is capable of looking at the elements in the DOM near the template and selectively hydrating the elements, adding the appropriate scopes and reordering elements as necessary.

For example, you might server render a component like this:

```html
<ul x-data="{ colors: [
    { id: 1, label: 'Red' },
    { id: 2, label: 'Orange' },
    { id: 3, label: 'Yellow' },
]}">
    <template x-for.hydrate="color in colors" :key="color.id">
        <li x-text="color.label"></li>
    </template>
    <li x-text="color.label" key="1">Red</li>
    <li x-text="color.label" key="2">Orange</li>
    <li x-text="color.label" key="3">Yellow</li>
</ul>
```

Here Alpine will see the static markup with `key` attributes (`key` is REQUIRED for hydration to work) and will hydrate the template with the appropriate data. From the attached scopes and treat them just like any element generated by the template itself.

It will event correct the order and text like so:

```html
<ul x-data="{ colors: [
    { id: 1, label: 'Red' },
    { id: 2, label: 'Orange' },
    { id: 3, label: 'Yellow' },
]}">
    <template x-for.hydrate="color in colors" :key="color.id">
        <li x-text="color.label"></li>
    </template>
    <li x-text="color.label" key="2">Orange</li>
    <li x-text="color.label" key="3">Red</li>
    <li x-text="color.label" key="1">Red</li>
</ul>
```

### Limitations

Currently, extra elements as siblings of the template that have a `key` attribute will not be removed and may cause errors.

Without Morph (see below), the static markup will be used as is. It will not be updated if differences exist between the static markup and the template markup.

### Using with Morph

If you are using [Morph](https://alpinejs.dev/plugins/morph), you can leave the directives off of the static markup and Morph will automatically add them for you. This is force the static markup to also have a matching layout to the template when difference exist, while still allowing you to server render the template.

```html
<ul x-data="{ colors: [
    { id: 1, label: 'Red' },
    { id: 2, label: 'Orange' },
    { id: 3, label: 'Yellow' },
]}">
    <template x-for.hydrate="color in colors" :key="color.id">
        <li x-text="color.label" style="color: blue"></li>
    </template>
    <li key="2">Orange</li>
    <li key="3">Red</li>
    <li key="1">Red</li>
</ul>
```


andrewbelcher:
I was, without realising, working on the same feature at the same time! I've got a slightly different implementation which I've pushed in #3628. As mentioned there, I'm really happy to collaborate to get the best solution in!

It also didn't work with my markup when I then made changes to the data as a result of an AJAX lookup intended to supplement the data (add price/availability to a list of products). I'll see if I can get a minimum working scenario which we can ensure is covered in tests for the eventual solution!

ekwoka:
@CodiumAI-Agent /improve

calebporzio:
Going to close this (as I did with the other x-for hydration PR). Thanks though!

PR #3625, opened 2023-06-18
:construction_worker: Transpiles CDN builds to es2017

ekwoka:
Addresses: #3619 

ES2017 is the lowest absolutely required version of Alpine and is a common target still (but hopefully not too long).

This PR targets CDN builds to es2017.

Transpiling of modules to other targets is left to the consumer, as it should be.


dgknca:
This should be merged! I had to replace my app with jquery because of this.

https://github.com/alpinejs/alpine/discussions/3287#discussioncomment-7661216

PR #3628, opened 2023-06-19
Adds hydration for x-for

andrewbelcher:
This PR adds the ability to hydrate an `x-for` directive, allowing server side rendered loop results to be taken over for management by Alpine!

### Note #3624 was opened whilst I was working on this
I hadn't seen that #3624 until I was prepping to push my version. It takes a slightly different approach, doesn't support a couple things, but also works in my testing where #3624 didn't. Really happy to try and collaborate with @ekwoka to get a "best of both" solution!

From a user's perspective, it's almost identical. Though I had some issues with numeric vs string keys and added a `hydrate.int` modifier that typecast the keys and I opted for `:key` (for the exact reason called out in #3624.

Under the covers, the code is a little simpler, though possible more repetitive as scoping is duplicated rather than abstracted. This version _does_ handle additional keys that need to be removed (called out as a limitation in #3624), and opts to stop when it encounters an element without a `:key`, considering those to be the end of the scope of the loop.

### DIFF
--- a/packages/alpinejs/src/directives/x-for.js
+++ b/packages/alpinejs/src/directives/x-for.js
@@ -8,7 +8,7 @@ import { flushJobs } from '../scheduler'
 import { warn } from '../utils/warn'
 import { dequeueJob } from '../scheduler'
 
-directive('for', (el, { expression }, { effect, cleanup }) => {
+directive('for', (el, { modifiers, expression }, { effect, cleanup }) => {
     let iteratorNames = parseForExpression(expression)
 
     let evaluateItems = evaluateLater(el, iteratorNames.items)
@@ -19,6 +19,22 @@ directive('for', (el, { expression }, { effect, cleanup }) => {
 
     el._x_prevKeys = []
     el._x_lookup = {}
+
+    if (modifiers.includes('hydrate')) {
+        const intKey = modifiers[modifiers.indexOf('hydrate') + 1] === 'int'
+        let curEl = el;
+        while (curEl = curEl.nextElementSibling) {
+            const key = curEl.getAttribute(':key')
+                curEl.getAttribute(':key')
+            if (key === null) {
+                break
+            }
+
+            const castKey = intKey ? parseInt(key) : key
+            el._x_lookup[castKey] = curEl
+            el._x_prevKeys.push(castKey)
+        }
+    }
 
     effect(() => loop(el, iteratorNames, evaluateItems, evaluateKey))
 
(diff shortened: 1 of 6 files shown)


ekwoka:
Interesting way to go about this.

Quite a bit less code 🤔

If my understanding is correct, it basically just tosses all the keyed siblings into the lookups as if they had already rendered in the last loop.

And during the main rendering loop, it does a check to see if that node has been initialized and then does so there.

andrewbelcher:
> If my understanding is correct, it basically just tosses all the keyed siblings into the lookups as if they had already rendered in the last loop.
> 
> And during the main rendering loop, it does a check to see if that node has been initialized and then does so there.

Yes, that's exactly what my approach was! This may not be the right way to go about it... I worked backwards from what else was done in the for directive, rather than from a place of understanding the underlying Alpine internals. But it seemed to work in all the scenarios I tested once I'd handled initialising rhs reactivity. 

I really like your integration with Morph, so whatever approach we take, we should keep that as an option!

packages/alpinejs/src/directives/x-for.js
```diff
@@ -20,7 +20,23 @@ directive('for', (el, { expression }, { effect, cleanup }) => {
     el._x_prevKeys = []
     el._x_lookup = {}
 
-    effect(() => loop(el, iteratorNames, evaluateItems, evaluateKey))
+    if (modifiers.includes('hydrate')) {
+        const intKey = modifiers[modifiers.indexOf('hydrate') + 1] === 'int'
+        let curEl = el;
+        while (curEl = curEl.nextElementSibling) {
+            const key = curEl.getAttribute(':key')
+                curEl.getAttribute(':key')
```
ekwoka, line 28:
```suggestion

```

packages/alpinejs/src/directives/x-for.js
```diff
@@ -20,7 +20,23 @@ directive('for', (el, { expression }, { effect, cleanup }) => {
...
+            if (key === null) {
+                break
+            }
+
+            const castKey = intKey ? parseInt(key) : key
+            el._x_lookup[castKey] = curEl
+            el._x_prevKeys.push(castKey)
```
ekwoka, line 35: ah, okay, now I see where you meant you had issues with there being int keys.

ekwoka (commented):
Initial glance :)

packages/alpinejs/src/directives/x-for.js
andrewbelcher: Yes, if the keys from the `:key` attribute weren't the same type as the keys in the `x-data`, it would treat them as different items.

ekwoka: Yes, since mine just used the calculated keys and not the ones directly from the dom, it avoided this...
hmm...

calebporzio:
Curious how much demand there is for this. I understand the use case; I just want to know if there are enough people that stand to benefit or are already invested in it before I merge. I'm concerned this feature will spawn many bugs because of either edge cases or mis-use

andrewbelcher:
> Curious how much demand there is for this. I understand the use case; I just want to know if there are enough people that stand to benefit or are already invested in it before I merge. I'm concerned this feature will spawn many bugs because of either edge cases or mis-use

I fully appreciate that! I can definitely see cases where this would be beneficial, not least the seo side of things! But I did this for a proof of concept which isn't going anywhere immediately, so I wouldn't merge it just for me!

I guess the question is does alpinejs want to support the hydration use case. If so, a solution for handling templates in both x-for and x-if will be needed. Personally, it would be an edge case rather than the norm. Without it, it will limit alpinejs to enhancements that don't matter in the nojs/seo world and administrative uses. But that is where I see the most value personally!

SimoTod:
Just throwing another option in the mix. Theoretically, you can add server-side the same content you expect to be generated by Alpine if important for a nojs/seo scenario. 
Just add an x-init="$el.remove" to your server side list items (or divs, spans, etc) so when Alpine kicks in, what would be a duplicate cintent gets removed. That achieves the same result as this PR with no changes (in both approaches, nothing guarantees that the server side content matches the initial Alpine state).

calebporzio:
That's a really cool idea @SimoTod - thanks!

Going to close this as I don't want to introduce this feature right now without having more time to gauge demand and think it through. Thanks so much for the contribution and conversation!

PR #3631, opened 2023-06-20
Fix tests broken by e579cc8.

andrewbelcher:
e579cc8 introduced some changes to tests that appear to have been unnecessary and broke the tests.


calebporzio:
Thank you!!

PR #3644, opened 2023-07-05
Add named exports to build files

ypconstante:
Like discussed on #3637, Node handles default export differently from other tools, causing the imported object to be wrapped in another object with a `default` key. 

This PR adds named export to all packages, allowing imports that work with both import interpretations. 
For plugins the exported name has a `Plugin` suffix, since `@alpinejs/morph`  already has a export named only `morph`.

[This solution is the recommendation from esbuild](https://esbuild.github.io/content-types/#default-interop).


packages/morph/builds/module.js
```diff
@@ -1,5 +1,5 @@
-import morphPlugin, { morph } from '../src/index.js'
```
calebporzio, line 1: I think these are two separate exports and should remain that way.

But now I see that this would mean this plugin diverges from the rest of the named plugin exports.

Hmmmm.... the "morphPlugin" variable is actually the plugin, and "morph" is the morphing function, not the full plugin.

So this change would cause a breakage for other people consuming { morph }

It's not documented so I'm not sure how much of a breaking change it would be, but it at least will probably break something internally? or with Livewire?

ypconstante: Just found at least one usage of this. [Jest tests are using this export](https://github.com/alpinejs/alpine/blob/main/tests/jest/morph/external.spec.js#L1).

Are this tests still used? They are not running on CI, and currently are failing.

PR #3658, opened 2023-07-15
Morph key lookup issue

joshhanley:
This PR fixes an issue with morph, where if the key already exists in the DOM and is to be moved up during morph, it wasn't getting moved correctly.

The reason was, that `fromKeyDomNodeMap` was empty, so the `fromKeyNode` in the below snippet was always empty, so the move "from" key was never running.

https://github.com/alpinejs/alpine/blob/a9206161f0bf878d41a001197553613cf31cbde9/packages/morph/src/morph.js#L259-L265

Removing the commented out `keyToMap(fromChildren)` got it working, but then caused an infinite loop. This was due to `dom.replace()` replacing the old with the new, but leaving the new one in its place in the array, causing two references to the dom node. So then when any `dom.()` methods run and try to find the index of the node, there would be two of them but only the first was ever returned. I opted to update `dom.replace()` but this could also be done in the above from key move code.

This fixes livewire/next#99

There is a failing test `directives/x-bind-style.spec.js ` but running those tests locally they pass, so I believe it's unrelated.

Hope this helps!

### DIFF
--- a/packages/morph/src/dom.js
+++ b/packages/morph/src/dom.js
@@ -14,11 +14,17 @@ export let dom = {
     replace(children, old, replacement) {
         let index = children.indexOf(old)
 
+        let replacementIndex = children.indexOf(old)
+
         if (index === -1) throw 'Cant find element in children'
 
         old.replaceWith(replacement)
 
         children[index] = replacement
+
+        if (replacementIndex) {
+            children.splice(replacementIndex, 1)
+        }
 
         return children
     },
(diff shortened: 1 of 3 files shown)


joshhanley:
Turns out the failing `x-bind-style` tests are due to the formatting in the test file. Someone has already submitted a PR to fix  https://github.com/alpinejs/alpine/pull/3631

packages/morph/src/dom.js
```diff
@@ -14,12 +14,18 @@ export let dom = {
     replace(children, old, replacement) {
         let index = children.indexOf(old)
 
+        let replacementIndex = children.indexOf(old)
```
calebporzio, line 17: Heads up, this should be "replacement" instead of "old" - tracked it down because of a pagination test failing. Fixed now. Thanks!

PR #3666, opened 2023-07-20
fix alpine/focus build issues

bty888:


ekwoka:
This doesn't address the issue.

The dependency bundled into the package is set in the package.json.

It has nothing to do with caches or anything.

PR #3667, opened 2023-07-20
bump focus-trap version

bty888:
Shadow DOM support  is added by focus-trap in 6.8.0 , updating the dependency so that x-trap working with Shadow DOM . 
https://github.com/alpinejs/alpine/discussions/3663


calebporzio:
Nice, thanks!

PR #3668, opened 2023-07-20
:sparkles: Allows returning a function from init to run after unmount

ekwoka:
This PR enables the common behavior in component based UI systems, where an `onmount` function can return a function that will then be called at cleanup.

This would be an alternative to `destroy`, especially for cleaning up any `effect` made in `init` but with the benefit of not needing to manage storing data outside of the closure.

This also handles `async init` methods that would return a promise that resolves to the cleanup


calebporzio:
This makes sense and I appreciate the contribution, however as it is a fairly impactful change I'm going to hold off. Thanks again

ekwoka:
I know you mentioned possibly adding it in the past, but wasn't sure if it was backburnered or you decided against it.

PR #3672, opened 2023-07-21
:bug: Fixes reference bug for uninitalized variables

ekwoka:
Addresses: #3671 

Theres also some other questions I have about further optimizations of this code.

- The `garbageCollector` stuff there is only increasing memory, not decreasing it
- quite a few extra calls to `Array.from` for using map, or just getting an array, but using `Array.prototype.map.call` can use the method without needing to create an intermediary array, as well as just using `childNodes` to get a NodeList that then also has `forEach` methods...

edit: added commit that does some refactoring. That's separate from the first that is the bug fix.

### DIFF
--- a/packages/navigate/src/page.js
+++ b/packages/navigate/src/page.js
@@ -54,7 +54,7 @@ function mergeNewHead(newHead) {
     // Only add scripts and styles that aren't already loaded on the page.
     let garbageCollector = document.createDocumentFragment()
 
-    for (child of Array.from(newHead.children)) {
+    for (const child of Array.from(newHead.children)) {
         if (isAsset(child)) {
             if (! headChildrenHtmlLookup.includes(child.outerHTML)) {
                 if (isScript(child)) {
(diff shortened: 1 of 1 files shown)


calebporzio:
Yeah, I think there are a few things going on here that I have questions about.

I'd be happy to merge a PR where you remove instances of Array.from() to not be as memory intensive, however my other hangups are:

* I think the purpose of the "garbage" collector is to remove the desired elements from the "newHead" before adding them to the current head. By appending to the garbage collector you are actually removing the element from it's previous place. So removing that breaks the desired behavior
* I don't mind some of the other refactors but they should be a separate PR
* Little things like using semicolons isn't inline with the style of the surrounding code

Thank you for being willing to contribute

ekwoka:
Appending to the other garbage document isn't needed in this factor that only loops through the whole list a single time. The `newHead` will get garbage collected with all those nodes attached.

On the note of the style, it would probably be very helpful if you pushed whatever formatting settings you are using in vscode to the repo.

Or even add like a precommit linting hook or something.

I'll move the refactors to another PR.

PR #3678, opened 2023-07-24
Livewire3

calebporzio:


packages/navigate/src/links.js
```diff
@@ -1,51 +1,52 @@
...
-            })
-        })
+export function whenThisLinkIsHoveredFor(el, ms = 60, callback) {
+    el.addEventListener('mouseenter', e => {
+        let timeout = setTimeout(() => {
+
+        }, ms)
+
+        let handler = () => {
+            clear
```
atduarte, line 37: @calebporzio was this line intentional? Leading to some "Uncaught ReferenceError: clear is not defined"

PR #3679, opened 2023-07-27
Fixed webpack 4 builds breaking due to nullish coalescing operator

mgschoen:
### What's the issue?

After upgrading to the latest version of `alpinejs` in a webpack 4 project, I discovered that my build was broken due to webpack encountering the unexpected token `??` in Alpine's source. This is the [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) and it was introduced in `alpinejs@3.12.1` (https://github.com/alpinejs/alpine/pull/3483). It's fairly common nowadays. It wasn't though when webpack 4 was a thing, and it turned out webpack 4 never got support for it. See this nice write-up of the same problem in a completely unrelated project: https://github.com/PaulLeCam/react-leaflet/issues/883

### Proposal

The problem could be solved on the consumer side by adding transpilation for `??`. I'd still suggest fixing it on the library side for the following reasons:

- this kind of deprecation should probably only occur in a major version
- the webpack 4 -> webpack 5 migration is pretty painful especially in complex setups and I suppose my project is not the only one that didn't yet make the jump
- `??` can easily be replaced and we get extended webpack 4 support with virtually no effort

What do you guys think?

### DIFF
--- a/packages/alpinejs/src/directives/x-model.js
+++ b/packages/alpinejs/src/directives/x-model.js
@@ -134,7 +134,7 @@ function getInputValue(el, modifiers, event, currentValue) {
         // Safari autofill triggers event as CustomEvent and assigns value to target
         // so we return event.target.value instead of event.detail
         if (event instanceof CustomEvent && event.detail !== undefined)
-            return event.detail ?? event.target.value
+            return event.detail !== null && event.detail !== undefined ? event.detail : event.target.value
         else if (el.type === 'checkbox') {
             // If the data we are binding to is an array, toggle its value inside the array.
             if (Array.isArray(currentValue)) {
--- a/packages/mask/src/index.js
+++ b/packages/mask/src/index.js
@@ -174,7 +174,9 @@ export function formatMoney(input, delimiter = '.', thousands, precision = 2) {
     if (input === '-') return '-'
     if (/^\D+$/.test(input)) return '9'
 
-    thousands = thousands ?? (delimiter === "," ? "." : ",")
+    if (thousands === null || thousands === undefined) {
+        thousands = delimiter === "," ? "." : ","
+    }
 
     let addThousands = (input, thousands) => {
         let output = ''


ekwoka:
This seems improper. 

Nullish coalescing is basically from the same time that Alpines minimum features are from.

If you haven't updated Webpack in 3 years, then why would you need to update Alpine now?

The CDN version is already transpiled to target ES 2017 if you need that.

mgschoen:
Thanks for your thoughts. I agree we're quite late with the webpack update. But I'd say outdated build tools are not a super uncommon thing, aren't they? We're trying to keep at least the code we ship to users up to date, so updating Alpine is definitely a priority.

SimoTod:
Don't you just need to use https://babeljs.io/docs/babel-plugin-transform-nullish-coalescing-operator in your build process?

mgschoen:
@SimoTod Sure, that works. But my proposal is fixing it on the library side so people don't need a transpiler to be able to use the library. Feel free to reject if this is not a priority for you, we'll be fine without the change!

SimoTod:
> Sure, that works. But my proposal is fixing it on the library side so people don't need a transpiler to be able to use the library. Feel free to reject if this is not a priority for you, we'll be fine without the change!

No worries, I wasn't sure you were aware of the option. I don't have decisional power, the maintainer will review and decide.

ekwoka:
@UnleashedMySelf 
> it's just the only issue that makes AlpineJS not work on ios 12-13

The CDN already transpiles this, and you can set your own bundling target (and should be).

nullish coalescing is transpilable, so it's better to set up your tooling for the goals you have than to hold the rest of the crabs in the bucket.

calebporzio:
Because there are so few instances of "??" I'm going to merge this, but I agree we should be able to use these syntaxes and may in the future. Thanks. everyone

PR #3681, opened 2023-07-29
fix: Entangle inner state undefined

danharrin:
This was causing a serious Livewire issue for us where a `$wire.entangle()` was being removed after an AJAX request, which then caused `inner` to be `undefined` here, which crashed Livewire completely and made all further interactions break.

To fix this, I've reached for the already-stringified `innerHashLatest` and ensured that it is `null` instead of `undefined` in this case. If you don't want to reuse `innerHashLatest` here, I can swap it back for `JSON.stringify(inner ?? null)`.


calebporzio:
Thanks Dan!

PR #3685, opened 2023-08-01
[livewire] Update navigate page.js

PhiloNL:
Fixes https://github.com/livewire/livewire/discussions/6144


calebporzio:
Thanks Philo

PR #3702, opened 2023-08-08
clarify execution order of `init`

browner12:
there are multiple ways to run code on Alpine initialization, so this clarifies the order of execution of the object `init()` method vs the `x-init` directive.


ekwoka:
Nice!

calebporzio:
Good context, thanks!

PR #3707, opened 2023-08-12
Add ability to get registered data

apih:
This PR adds the ability to get registered data in `Alpine.data()`. This will allow a use case like this:

```js
Alpine.data('Component1', {
  // bla bla
});

Alpine.data('Component2', {
  // bla bla
});

Alpine.data('Component3', {
  ...Alpine.data('Component1'),
  ...Alpine.data('Component2'),
});
```

### DIFF
--- a/packages/alpinejs/src/datas.js
+++ b/packages/alpinejs/src/datas.js
@@ -1,7 +1,11 @@
 
 let datas = {}
 
-export function data(name, callback) {
+export function data(name, callback = null) {
+    if (callback === null) {
+        return datas[name]
+    }
+
     datas[name] = callback
 }
 


ekwoka:
You can just do that in the x-data

```js
x-data="{...Component1, Component2}"
```

Just know this will break getters/setters

apih:
I am not aware that getters and setters will break when merged using spread operator. Mainly because I rarely use getters and setters. Will need to be careful in the future then.

ekwoka:
Yeah, it makes sense when you think about it. If you enumerate the properties and values on an object, getters won't be getters, they will be the key and the value got from running that getter at that moment.

similarly, setters aren't enumerable at all.

the `defineProperties` method actually finds out the full description of the property (enumerable? configurable? read only? getter? setter?) and replicates all of that to the new object.

calebporzio:
Good insight @ekwoka - thanks!

PR #3716, opened 2023-08-21
Document destroy method in Alpine.data

maritaria:
Document the `destroy()` method of components on the docs page for `Alpine.data`. Includes an example allowing the reader to easily play and experiment with the behavior.


maritaria:
Additionally, the `essentials/lifecycle` may also benefit from a link to the `Alpine.data` page, as the `destroy()` method is not mentioned in the lifecycle at all.

ekwoka (approved):
I like it 👍

PR #3727, opened 2023-08-25
Fixed a variable typo

sakanjo:
Fixed an issue that causes an error to happen when the magic method is not registered: `caught ReferenceError: directiveName is not defined`


calebporzio:
Thanks!

PR #3729, opened 2023-08-25
fix(morph): currentNode can also be null, not just undefined, fixes #3728

levu42:


ekwoka:
#3762

joshhanley:
@levu42 thanks for the PR! Any chance you could add a test for this? Thanks!

levu42:
@joshhanley not easily, since, as I described in #3728 I can't reproduce it in a minimal test case. I have it happen in a bigger Livewire project, but my tries to make a minimal test case weren't successful so far.

joshhanley:
@levu42 yeah I just tried too, and couldn't replicate it using morph alone outside of a Livewire context.

calebporzio:
Thanks!

PR #3737, opened 2023-08-30
x-if warning if used on non-template

jonas-grobe:
Added a warning if x-if is used on a tag other than template (similar to x-teleport warning).

### DIFF
--- a/packages/alpinejs/src/directives/x-if.js
+++ b/packages/alpinejs/src/directives/x-if.js
@@ -5,8 +5,11 @@ import { initTree } from '../lifecycle'
 import { mutateDom } from '../mutation'
 import { walk } from "../utils/walk"
 import { dequeueJob } from '../scheduler'
+import { warn } from "../utils/warn"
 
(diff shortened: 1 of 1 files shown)


packages/alpinejs/src/directives/x-if.js
```diff
@@ -5,8 +5,11 @@ import { initTree } from '../lifecycle'
 import { mutateDom } from '../mutation'
 import { walk } from "../utils/walk"
 import { dequeueJob } from '../scheduler'
+import { warn } from "../utils/warn"
 
 directive('if', (el, { expression }, { effect, cleanup }) => {
+    if (el.tagName.toLowerCase() !== 'template') warn('x-if can only be used on a <template> tag', el)
```
ekwoka, line 11:
```suggestion
    if (!el instanceof HTMLTemplateElement) warn('x-if can only be used on a <template> tag', el)
```

ekwoka (changes requested):

calebporzio:
Good call, thanks!

PR #3743, opened 2023-09-02
:bug: Passes through modifiers in x-bind

ekwoka:
Fixes #3742 

When attempting to bind attributes that contain `.`, x-bind greedily consumes them even when they are not valid modifiers for x-bind.

this allow unrecognized bind modifiers to be passed through to the bound attribute

Maybe it should just pass all modifiers through for simplicity?


calebporzio:
Thanks for contributing!

In order to keep this project as focused and maintainable as possible, I'm going to close this pull request.

PR #3756, opened 2023-09-12
:white_check_mark: Allows underscore in event names

ekwoka:
Solves #3023 

Allowing underscores in event names

Test includes


calebporzio:
Thanks!

PR #3762, opened 2023-09-14
update CSP build/installation docs

browner12:
there has been a lot of confusion on how to use the CSP build, so this commit hopefully clears them up, and provides more explicit instructions on how to "build from source".

some of the wording and code examples make it seem like an NPM package is available, so we'll update that to avoid confusion.


calebporzio:
Thanks!

PR #3764, opened 2023-09-14
use normal single quotes

browner12:
use a standard single quote, which is common practice, and is consistent with the rest of the repo.


ekwoka (approved):

calebporzio:
Thanks!

PR #3765, opened 2023-09-15
:sparkles: Names expression evaluators

ekwoka:
This PR adds the expression as a name to the evaluator function. While Alpine logging information about the expression is helpful, many tools like sentry won't fully track or associate those warning logs with the actual error.

This change allows the expression to show as the name of the function in the stack trace itself making hunting down errors much easier.

Screenshot 2023-09-15 at 1 35 53 PM


calebporzio:
Good calls - thanks!

PR #3794, opened 2023-10-04
Use Internet Explorer conditional comment syntax to preserve morph markers on Cloudflare-proxied sites

markjaquith:
This changes the morph markers that Alpine recognizes to match the ones in this Livewire PR: https://github.com/livewire/livewire/pull/6974

`<!--[if BLOCK]><![endif]-->`
`<!--[if ENDBLOCK]><![endif]-->`

Read that PR for more details, but basically the issue is that some proxies strip HTML comments, which means that the morph markers that a backend like Livewire renders might get stripped by things like Cloudflare (with HTML minification on). The Internet Explorer conditional syntax above will be ignored by Cloudflare (and likely other HTML-minification pipelines) because historically it was used to enable or disable functionality on certain browsers.

### DIFF
--- a/packages/morph/src/morph.js
+++ b/packages/morph/src/morph.js
@@ -156,8 +156,8 @@ export function morph(from, toHtml, options) {
             }
 
             // Handle conditional markers (presumably added by backends like Livewire)...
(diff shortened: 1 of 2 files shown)


calebporzio:
Great! Thank you Mark!

PR #3807, opened 2023-10-12
:bug: Fixes proxy stack setters bug

ekwoka:
Solves #3804 

This was a regression in the behaviors from the merge proxy optimization.

This still seems problematic in many ways. but this more faithfully represents the older logic.

If there is a getter setter pair, the `this` refers to the proxy stack. If it's just a setter, it doesn't (this was how the original logic worked).

Similarly to before, using getters/settings/methods,etc on the parent component can modify the child component when keys match. Not great, but that's how it has been.

It could be possible to instead limit it only to that level and parents, but that would change the behavior.

This also adds tests to ensure this continues to work :)

### DIFF
--- a/packages/alpinejs/src/scope.js
+++ b/packages/alpinejs/src/scope.js
@@ -64,14 +64,14 @@ let mergeProxyTrap = {
         )
     },
 
-    set({ objects }, name, value) {
-        return Reflect.set(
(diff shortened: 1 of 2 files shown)


calebporzio:
Thanks @ekwoka for handling this!

PR #3808, opened 2023-10-12
typo in 'x-transitions' duration syntax

itopizarro:
This changes the syntax of the reference example from `x-transition.duration.500ms` to `x-transition:duration.500ms`

### DIFF
--- a/packages/docs/src/en/directives/transition.md
+++ b/packages/docs/src/en/directives/transition.md
@@ -51,7 +51,7 @@ Initially, the duration is set to be 150 milliseconds when entering, and 75 mill
 You can configure the duration you want for a transition with the `.duration` modifier:
 
 ```alpine
-<div ... x-transition.duration.500ms>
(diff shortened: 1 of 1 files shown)


ekwoka:
But it's not a type...

itopizarro:
Maybe it was my implementation?
I had an element with `x-transition` directives, where I was ok with setting both the `enter` and `leave` to the same value as follows:
```
<div x-show="showExample"
  x-transition:enter="foo"
  x-transition:leave="bar"
  x-transition.duration.1000ms>
```
But that didn't affect the duration at all. However, when I changed the instance regarding duration to `x-transition:duration.1000ms` it worked. I do see the need for the `.` notation when using `x-transition:enter.duration.1000ms`.

ekwoka:
This is because you aren't supposed to mix value based transitions and the default transition helpers.

I can look more into it if/when I have time, but it's more of a "happy accident" that what you described worked...

PR #3814, opened 2023-10-16
:bug: Fixes reference to incorrect attribute

ekwoka:
Seems like this is an incorrect reference, since `_x_teleportBack` exists in the codebase but `teleportBack` does not.

Edit: Turns out it isn't even needed at all, and the typo is what was letting it not break.

Obligatory: TypeScript would have prevented this.


calebporzio:
🤦 - thanks @ekwoka ! (lol re: typescript)

PR #3844, opened 2023-11-02
:bug: Joins selectors to meet spec

ekwoka:
While an array of strings can work when the runtime chooses to coerce the array to a string, it is outside of the DOM spec, which allows only a string. Some DOM runtimes, like Happy-dom do not support selectors that are arrays.

This just brings the code inline with the spec.


calebporzio:
Ah interesting, I didn't know that thanks

PR #3851, opened 2023-11-05
Update refs.md

Simsz:
Added details on $refs not supporting dynamic binding anymore in V3


Simsz (commented):
Updated the example to reflect the real output of $refs

calebporzio:
Thank you!

PR #3881, opened 2023-11-23
fix: reset input on init so default value is displayed

coder2000:
Currently when setting a default value for the combobox it isn't displayed right away. You would have to open and close it before the display value would be visible. This will reset the input as part of the init so the value is displayed right away.

### DIFF
--- a/packages/ui/src/combobox.js
+++ b/packages/ui/src/combobox.js
@@ -1,74 +1,83 @@
-import { generateContext, renderHiddenInputs } from './list-context'
+import { generateContext, renderHiddenInputs } from "./list-context";
 
 export default function (Alpine) {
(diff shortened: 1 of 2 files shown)


coder2000:
Sorry. My editor has some defaults that are different than the codebase that I didn't notice until I opened the PR.

bakerkretzmar:
@coder2000 can you revert your formatting changes so it's possible to review this? This bug has been killing me but it took me forever to figure out what actually changed in this PR.

coder2000:
> @coder2000 can you revert your formatting changes so it's possible to review this? This bug has been killing me but it took me forever to figure out what actually changed in this PR.

Done.

calebporzio:
fixed and added tests, thanks!

PR #3908, opened 2023-12-08
x-mask: Cleanup of event listeners

beholdr:
Hi,
Found a bug when using x-mask with Livewire. Problem is sometimes input changed by Livewire and mask directive intiated every time. Adding cleanup of event listeners fix this issue.

Unfortunately don't know how to make correct tests for this situation without LW part, but you can check it on this wirebox: https://wirebox.app/b/4kwle

To reproduce:
1. Fill both inputs with `12345` string
2. Click toggle button
3. Try to enter `6` (to get `123456` string) in both fields. First field has weird behaviour, second field is OK.

Second field has patched version of mask plugin.

### DIFF
--- a/packages/mask/src/index.js
+++ b/packages/mask/src/index.js
@@ -1,6 +1,6 @@
 
 export default function (Alpine) {
-    Alpine.directive('mask', (el, { value, expression }, { effect, evaluateLater }) => {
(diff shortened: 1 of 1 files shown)


ekwoka:
Definitely should be there!

I am a bit surprised to learn event listeners accept abort signals. That's pretty cool.

calebporzio:
Thank you!

PR #3911, opened 2023-12-09
[Docs] Add range type to x-model inputs

peterfox:
# Changes

* Adds range input to examples in `model.md` documentation

# Why

I've been playing around with the range input type with AlpineJS and there seems to be no issue. It feels sensible to add it to the docs as a supported input type.

### DIFF
--- a/packages/docs/src/en/directives/model.md
+++ b/packages/docs/src/en/directives/model.md
@@ -62,6 +62,7 @@ Now when the `<button>` is clicked, the input element's value will instantly be 
 * `<input type="checkbox">`
 * `<input type="radio">`
 * `<select>`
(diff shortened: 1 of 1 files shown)


ekwoka:
Makes sense. All inputs work (even custom ones that conform to the dom specs)!

peterfox:
@ekwoka I assume that's the case but when looking at the docs it makes it seem like only those limited inputs will work from the way it's worded.

ekwoka:
Agreed. Might be worth making an earlier callout that it is supposed to work correctly with all native controls and custom controls that follow standards.

But then again, nobody reads the docs that closely 😂

PR #3928, opened 2023-12-14
add explanation of passing an object to `x-for`

browner12:
`x-for` also accepts objects in addition to arrays. the docs currently only mention arrays, so users might not be sure if objects are allowed. this explicitly states they are and gives an example how to use them.

I was trying to find the relevant PR for when this functionality was added, but was unable to. I found some PRs from Ryan Chandler that were trying to add this feature but were closed.


calebporzio:
Thank you!

PR #3932, opened 2023-12-15
Fix typo

aerni:
I copy-pasted the code from the docs and the bundling failed because of wrong single quotes. Simple fix.


calebporzio:
Thanks!

PR #3935, opened 2023-12-18
Wait for custom elements to be defined before x-model.fill

saeldion:
Fixes https://github.com/alpinejs/alpine/discussions/3813

x-model needs to wait for custom elements to be defined before setting up the directive.

### DIFF
--- a/tests/cypress/integration/directives/x-model.spec.js
+++ b/tests/cypress/integration/directives/x-model.spec.js
@@ -279,3 +279,34 @@ test(
     }
 );
 
+test(
+    'x-model with fill on undefined custom element',
+    html`
+        <script>
+            class CustomElement extends HTMLElement {
+                connectedCallback() {
+                    this.innerText = "Loaded"
+                }
+
+                get value() {
+                    return "This is the internal value"
+                }
+
+            }
+
+            setTimeout(() => {
+                customElements.define("custom-element", CustomElement)
+            }, 300)
+        </script>
+
+        <div x-data="{ a: null }">
+            <span x-text="a"></span>
+            <custom-element x-model.fill="a">Not loaded</custom-element>
+        </div>
+    `,
+    ({get}) => {
+        get("custom-element").should(haveText("Not loaded"))
+        get("custom-element").should(haveText("Loaded"))
+        get("span").should(haveText("This is the internal value"))
+    }
+)
--- a/packages/alpinejs/src/directives/x-model.js
+++ b/packages/alpinejs/src/directives/x-model.js
@@ -7,8 +7,13 @@ import on from '../utils/on'
 import { warn } from '../utils/warn'
 import { isCloning } from '../clone'
 
-directive('model', (el, { modifiers, expression }, { effect, cleanup }) => {
+directive('model', async (el, { modifiers, expression }, { effect, cleanup }) => {
     let scopeTarget = el
+
+    const tagName = el.tagName.toLowerCase()
+    if (tagName.includes("-") && !customElements.get(tagName)) {
+        await customElements.whenDefined(tagName);
+    }
 
     if (modifiers.includes('parent')) {
         scopeTarget = el.parentNode


tests/cypress/integration/directives/x-model.spec.js
```diff
@@ -279,3 +279,34 @@ test(
...
+
+                get value() {
+                    return "This is the internal value"
+                }
+
+            }
+
+            setTimeout(() => {
+                customElements.define("custom-element", CustomElement)
+            }, 300)
```
saeldion, line 299: I'm don't know Cypress very well. So I appreciate feedback on how to delay the definition of the custom element.

saeldion:
We are using custom elements in our project, some of which are lazy imported when discovered. If Alpine runs before the custom element and the custom element is using x-model the reactivity is not setup and the elements value will not be picked up even after it is registered, at least in my experience.

So preferably should Alpine wait custom elements to be defined before running its directives. Maybe this could be applied on all directives and maybe on elements that are children of a custom elements?

ekwoka:
Well, there is no way to know that an element is a custom element that will be registered later.

It may be better to use `x-ignore` on those and have the components connected callback remove it.

saeldion:
> It may be better to use `x-ignore` on those and have the components connected callback remove it.

That could be a work around, sure. However, I might be doing something wrong, but it doesn't seem like the `x-model` directive is re-run when `x-ignore` is removed.

ekwoka:
🤔 It definitely should....

May need to call `Alpine.initTree` on it...

There's gotta be a pattern that works best for this...

saeldion:
`Alpine.initTree(this)` doesn't work either 🤔 

I wrote this test to "reproduce" it. Am I missing anything?

```
test(
    'x-model is re-run after x-ignore is removed',
    html`
        <script>
            class CustomElement extends HTMLElement {
                get value() {
                    return 'This is the internal value'
                }

                connectedCallback() {
                    this.removeAttribute("x-ignore")
                    Alpine.initTree(this)

                    // this.dispatchEvent(new Event("change", {bubbles: true})) // Even tried dispatching a change event to trigger alpine
                }
            }

            setTimeout(() => {
                customElements.define("custom-element", CustomElement)
            }, 500)
        </script>

        <div x-data="{ a: null }">
            <custom-element x-ignore x-model.fill="a"></custom-element>
            <span x-text="a"></span>
        </div>
    `,
    ({ get }) => {
        get('span').should(haveText("This is the internal value"));
    }
)
```

saeldion:
Haven't tried it in my codebase yet but in [this Codepen ](https://codepen.io/Saeldion/pen/ExMjjdR) I wrote, executing `Alpine.destroyTree` before `Alpine.initTree` seems to work. I'll give it a proper test tomorrow.

ekwoka:
That's curious that it would require both 🤔

saeldion:
> That's curious that it would require both 🤔

Yeah... It would be nice if x-ignore was more "dynamic" and it being added and removed as picked up automatically.  
I also noticed that if `x-ignore` is added at a later time things still have reactivity in that subtree.

calebporzio:
Thanks for this report. The change you proposed will have other ripple effects by making x-model's handler async. For example it will now be bumped to the end of the microtask queue for non-custom element registrations and ruin the predictable order we have for directive registration.

Because .fill is a smaller feature, I don't think it's worh the core re-working to make it work.

There are probably cleaner workarounds than what you have but I can't think of them off the top of my head.

Perhaps just making sure your custom element registrations happen before Alpine initializes.

Anyhow, thanks for the contribution!

saeldion:
> The change you proposed will have other ripple effects by making x-model's handler async. For example it will now be bumped to the end of the microtask queue for non-custom element registrations and ruin the predictable order we have for directive registration.

Yeah I figured as much. Was mostly an attempt to showcase my problem and probably not as the final solution.

Maybe .fill is a smaller feature but we use it quite extensively when building more interactive forms so it would be nice to have some kind of support or at least a suggestion for a "workaround".
As mentioned above `x-ignore` could work for this but having to use `destroyTree` and `initTree` feels clunky. It would be really nice if Alpine could handle `x-ignore` being added/removed automatically.

ekwoka:
That is something that could be done.

Right now `x-ignore` is a not a "real" directive, it doesn't run anything. the tree initialization just checks for it.

some logic could be added as an actual `x-ignore` directive for just a cleanup that then initializes the tree on that element.

```js
Alpine.directive('ignore',
  (el, { modifiers }, { cleanup }) => 
    !modifiers.includes('self') && cleanup(() => el.isConnected() && Alpine.initTree(el)))
```

As a sloppy one.

This should also handle if the element is jsut completely removed so it doesn't try to initialize a detached tree.

saeldion:
Okay I will give that a try.

Hopefully this is something that could be part of Alpine some day?

ekwoka:
I would probably prefer custom elements just initialize on themselves when they start.

PR #3942, opened 2023-12-20
Bug in code example of documentation for x-anchor to an ID

MaximVanhove:
The code example uses `#` in the `getElementById` function, which breaks the code

```js
// Wrong
document.getElementById('#trigger')

// Right
document.getElementById('trigger')
```

Full code example

```html
    <button id="trigger" @click="open = ! open">Toggle</button>

    <div x-show="open" x-anchor="document.getElementById('trigger')">
        Dropdown content
    </div>
</div>
```


calebporzio:
Thanks!

PR #3951, opened 2023-12-27
fix: already init component

Matsa59:
Currently we defined if a mutation is a move operation if the node is in the `addedNodes` and in the `removedNodes`. 

However in some case, an element is added, removed and added again. So it should considered as `add` + `move` operations. But alpine will ignore it, considering it as a move operation only (so nothing will be init).

The PR focus on introducing a new state `_x_isInit` that will be true for component that are initialized.

Also it may speed up a little bit the `onMutate` function. Previously we checked every deleted nodes before adding any new nodes. For big mutations, iterate over the `deletedNodes` several times could be CPU intensive.

> Conversation on : https://github.com/alpinejs/alpine/discussions/3947

### DIFF
--- a/packages/alpinejs/src/mutation.js
+++ b/packages/alpinejs/src/mutation.js
@@ -194,10 +194,6 @@ function onMutate(mutations) {
         node._x_ignore = true
     })
     for (let node of addedNodes) {
-        // If an element gets moved on a page, it's registered
-        // as both an "add" and "remove", so we want to skip those.
-        if (removedNodes.includes(node)) continue
-
         // If the node was eventually removed as part of one of his
         // parent mutations, skip it
         if (! node.isConnected) continue
--- a/packages/alpinejs/src/mutation.js
+++ b/packages/alpinejs/src/mutation.js
@@ -194,14 +194,9 @@ function onMutate(mutations) {
         node._x_ignore = true
     })
     for (let node of addedNodes) {
-        // If the node is already init, it means it's a move operation
-        if (node._x_isInit) continue
-
         // If the node was eventually removed as part of one of his
         // parent mutations, skip it
         if (! node.isConnected) continue
-
-        node._x_isInit = true
 
         delete node._x_ignoreSelf
         delete node._x_ignore
--- a/packages/alpinejs/src/mutation.js
+++ b/packages/alpinejs/src/mutation.js
@@ -194,9 +194,14 @@ function onMutate(mutations) {
         node._x_ignore = true
     })
     for (let node of addedNodes) {
+        // If the node is already init, it means it's a move operation
+        if (node._x_isInit) continue
+
         // If the node was eventually removed as part of one of his
         // parent mutations, skip it
         if (! node.isConnected) continue
+
+        node._x_isInit = true
 
         delete node._x_ignoreSelf
         delete node._x_ignore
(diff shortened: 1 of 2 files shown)


packages/alpinejs/src/mutation.js
```diff
@@ -194,14 +194,15 @@ function onMutate(mutations) {
...
-        // as both an "add" and "remove", so we want to skip those.
-        if (removedNodes.includes(node)) continue
+        // If the node is already init, it means it's a move operation
+        if (node._x_isInit) continue
 
         // If the node was eventually removed as part of one of his
         // parent mutations, skip it
         if (! node.isConnected) continue
 
+        node._x_isInit = true
```
ekwoka, line 204: This is the incorrect place to handle this, as this does not actually do any initializing.

ekwoka (changes requested):
Should probably handle tracking initialization tracking in the initialization process, not the mutation process.

packages/alpinejs/src/mutation.js
Matsa59: Do you mean something like 

```js
export function initTree(el, walker = walk, intercept = () => {}) {
    deferHandlingDirectives(() => {
        walker(el, (el, skip) => {
            intercept(el, skip)

            initInterceptors.forEach(i => i(el, skip))

            directives(el, el.attributes).forEach(handle => handle())

            if (el._x_ignore) {
                skip()
            } else {
                el._x_isInit = true;
            }
        })
    })
}
```

I'm really newbie for the Alpine code source ;)

ekwoka: Almost. More like on every walk of an element. This should only do it on tree roots.

Matsa59: hum, you mean in the `walker` callback ?
```diff
export function initTree(el, walker = walk, intercept = () => {}) {
    deferHandlingDirectives(() => {
        walker(el, (el, skip) => {
+           el._x_isInit = true
+
            intercept(el, skip)

            initInterceptors.forEach(i => i(el, skip))

            directives(el, el.attributes).forEach(handle => handle())
            el._x_ignore && skip()
        })
    })
}
```

the previous code works well for the moment, I could try this one too if you want

Matsa59: Okay I get it, I simply do this

```js
export function initTree(el, walker = walk, intercept = () => {}) {
    deferHandlingDirectives(() => {
        walker(el, (el, skip) => {
            if (el._x_isInit) return;
            el._x_isInit = true
            // ...
```

in the lifecycle and everything works great, we have a couple of edge case they all works perfectly \o/

Matsa59: hum that's strange cause there is a test that failed, however everything works. There is something that I don't catch

Matsa59:
FINALLY I found a way to make it works ... waiting for CI to be sure

calebporzio:
Thanks! I do think the implementation is a little heavy handed. I instead just changed the addedNodes and removedNodes to JS `Set`s instead of arrays so that they enforce uniqueness in the case of duplicate addeds and removes. Thanks!

ekwoka:
Does that solve the actual issue this is meant to solve?

It still will incorrectly consider an element that is added and immediately moved as being a "moved" element and never initialize it, which was the issue.

Matsa59:
I just say the `console.log` oupsi x)
As you can see hard debugging things

> Thanks! I do think the implementation is a little heavy handed. I instead just changed the addedNodes and removedNodes to JS `Set`s instead of arrays so that they enforce uniqueness in the case of duplicate addeds and removes. Thanks!

It was my first though but it was more complex and didn't fix well the issue :/ 

Thanks for merging, really appreciated and sorry for the `console.log`

SimoTod:
@Matsa59 I think this change caused https://github.com/alpinejs/alpine/discussions/3994

PR #3956, opened 2024-01-02
Bug Fix in Docs: Update mask plugin docs to include 'x-data' attribute

ARehmanMahi:
The 'x-data' attribute has been now included in the input fields of the examples given in the mask plugin documentation. This helps to demonstrate its usage correctly; plus, placeholders added to match the demo fields.

Without x-data Mask doesn't work. 
Reproducable [JSFidle demo](https://jsfiddle.net/hLd24ey7/)

### DIFF
--- a/packages/docs/src/en/plugins/mask.md
+++ b/packages/docs/src/en/plugins/mask.md
@@ -69,12 +69,12 @@ The primary API for using this plugin is the `x-mask` directive.
 Let's start by looking at the following simple example of a date field:
 
 ```alpine
-<input x-mask="99/99/9999" placeholder="MM/DD/YYYY">
+<input type="text" x-data x-mask="99/99/9999" placeholder="MM/DD/YYYY">
 ```
 
 <!-- START_VERBATIM -->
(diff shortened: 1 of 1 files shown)


ekwoka:
I'm not sure this sends the right message though.

As it stands, the docs as a whole are quite clear that `x-data` is required for alpine directives to work, but they do not need to be on every element.

The docs do not include x-data in any examples where the content of the x-data is not relevant to the example itself.

This change would break that consistency and potentially introduce miscommunication about x-mask requiring x-data on the individual element.

ARehmanMahi:
Your reasing is good, I'm no an Aline user but wanted to try Mask plugin and fased the issue, I also saw this post [https://github.com/alpinejs/alpine/discussions/2920](https://github.com/alpinejs/alpine/discussions/2920 ) where no recent asnwers wre given, hence I thought would be less ambigious for new users like myself. Coming from Jquery I'm used to copy paste demo code.

I agree with your reasing though. Please close this issue as you see fit. Thank you

PR #3959, opened 2024-01-04
Add CSP-friendly Alpine build

calebporzio:
This PR provides a version of AlpineJS that is compatible with strict content security policies that would otherwise prevent the usage of Alpine because of "unsafe-eval" violations.

It removes the ability to evaluate JavaScript expressions inside HTML attributes; however, it maintains ALL functionality of Alpine otherwise.

All plugins and features are CSP-compatible when using this build.


loicy:
Hi,

After test of 3.13.5 & 3.13.7, It seems that without the use of unsafe-eval is still mandatory.

ekwoka:
@loicy using @alpinejs/csp ? Just to be clear.

PR #3962, opened 2024-01-05
Listbox compare null value bugfix

gdebrauwer:
When there is no value selected and the `by` attribute is used, then an error occurs because it tries to access a property on `null`. This PR fixes that. I got the fix from the combobox.js file, as the issue was already fixed there.


calebporzio:
Thanks!

PR #3966, opened 2024-01-08
Combobox and listbox: Save optionKey in x-data to fix usage in Livewire

gdebrauwer:
When optionKey is not saved in x-data, the value is lost after a Livewire update. That causes the optionElement of the $listboxOption magic to not be found and that results in the combobox/listbox not working completely anymore.
```js
Alpine.magic('listboxOption', (el) => {
	let data = Alpine.$data(el)
	
	let optionEl = Alpine.findClosest(el, i => i.__optionKey)
	
	// This error will be thrown after a Livewire update (same in comboxOption-magic)
	if (! optionEl) throw 'No x-combobox:option directive found...'
	
	// ...
});
```

This PR fixes that by saving the optionKey in x-data.

### DIFF
--- a/packages/ui/src/combobox.js
+++ b/packages/ui/src/combobox.js
@@ -43,19 +43,19 @@ export default function (Alpine) {
     Alpine.magic('comboboxOption', el => {
         let data = Alpine.$data(el)
 
-        let optionEl = Alpine.findClosest(el, i => i.__optionKey)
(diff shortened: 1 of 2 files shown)


ekwoka:
hmm.

So the element is changing out, but the data is being carried over.

so the element attributes are lost...

calebporzio:
Going to close this as there are no tests, but when you get the chance @gdebrauwer - please re-submit with tests or at least  a better explanation of how to reproduce this issue. Thank you!

PR #3977, opened 2024-01-12
Add x-trap noinitialfocus modifier

richcarni:
Adds a new x-trap modifier: .noinitialfocus to trap focus without giving focus to the first element in the focus trap

Test included

Relates to discussion #2806

Use case is to be able to create a modal with the option of not focusing the first input. For example, Wire Element Modal is a Livewire component for creating modals, which uses x-trap, and currently has no mechanism (besides tricks like hidden inputs) to create a modal that opens without focusing its first input: [https://github.com/wire-elements/modal/discussions/367](https://github.com/wire-elements/modal/discussions/367)

### DIFF
--- a/packages/docs/src/en/plugins/focus.md
+++ b/packages/docs/src/en/plugins/focus.md
@@ -301,6 +301,13 @@ For example:
 </div>
 <!-- END_VERBATIM -->
 
+<a name="noautofocus"></a>
+#### .noautofocus
+
+By default, when `x-trap` traps focous within an element, it focuses the first focussable element within that element. This is a sensible default, however there are times where you may want to disable this behavior and not automatically focus any elements when `x-trap` engages.
+
+By adding `.noautofocus`, Alpine will not automatically focus any elements when trapping focus.
+
 <a name="focus-magic"></a>
 ## $focus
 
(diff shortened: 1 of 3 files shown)


ekwoka:
How does this handle (or communicate to devs) that the focus may now behave unpredictably for assistive technologies?

richcarni:
Thanks for the feedback @ekwoka 

I'll have to do some testing around how this affects the behaviour of assistive technologies. Will report back and update documentation accordingly

richcarni:
After testing and consideration of how this feature isn't compatible with assistive technologies, I've come to the conclusion that there are better solutions to the problem I thought it solved (see #2806)

Therefore, unless someone wants to argue to the contrary, I'll close this PR

calebporzio:
I like the addition - I've heard of cases where people want to disable autofocus on mobile dialogs because autofocusing automatically pulls up the keyboard when you open a modal and that's not ideal.

Anyhow, I like it and changed it to `.noautofocus`

Thanks!

richcarni:
Apologies for not getting around to writing the documentation @calebporzio 

I didn't end up closing this because I realised in my use case, which is for a Livewire modal, autofocus has to be implemented by the modal because it has no contents when the page is first loaded. So you end up with an ugly switch in focus. More annoyingly my first field is an address autocompletion, which incurs a fee if activated

Anyway, thanks for merging!

PR #3990, opened 2024-01-22
Combobox and listbox: Save optionKey in x-data to fix usage in Livewire

gdebrauwer:
I created a listbox (or combobox) with options that are present in HTML and not generated by javascript.
```blade
<!-- in the blade file of a livewire component -->

<div>
    <button wire:click="$refresh" type="button">trigger livewire update</button>

    <div x-data="{ value: null }">
        <div x-listbox x-model="value">
            <button x-listbox:button>
                Select Framework
            </button>

            <ul x-listbox:options>
                <li :class="{
                    'text-blue-100': $listboxOption.isActive,
                    'text-yellow-100': !$listboxOption.isActive,
                    '!text-gray-60 opacity-50 cursor-not-allowed': $listboxOption.isDisabled,
                }" x-listbox:option value="laravel">
                    Laravel
                </li>

                <li :class="{
                    'text-blue-100': $listboxOption.isActive,
                    'text-yellow-100': !$listboxOption.isActive,
                    '!text-gray-60 opacity-50 cursor-not-allowed': $listboxOption.isDisabled,
                }" x-listbox:option value="rails">
                    Ruby on Rails
                </li>

                <li :class="{
                    'text-blue-100': $listboxOption.isActive,
                    'text-yellow-100': !$listboxOption.isActive,
                    '!text-gray-60 opacity-50 cursor-not-allowed': $listboxOption.isDisabled,
                }" x-listbox:option value="phoenix" :disabled="true">
                    Phoenix
                </li>
            </ul>
        </div>

        Selected: <span x-text="value"></span>
    </div>
</div>
```

When I use the listbox/combobox in a livewire component and a livewire update is triggered (by clicking the "trigger livewire update" button in the example above), then the following javascript warnings and errors occur:

Screenshot 2024-01-22 at 08 23 49

When I debugged the issue, I discovered that the `__optionKey` on the option HTML element disappears after the Livewire update. To fix that, I updated the listbox/combobox code so that the "optionKey" is saved in the Alpinejs' x-data as that data will not be disappear after a Livewire update.

I did not add tests because I'm unsure how to test this as Livewire causes it.

_This issue and another Livewire-related issue (https://github.com/livewire/livewire/discussions/7566) are currently the only 2 bugs preventing me from using the listbox/combobox in Livewire. I have tried to debug the other issue, but after many hours of searching, I have still not found what is causing the infinite loop._

### DIFF
--- a/packages/ui/src/combobox.js
+++ b/packages/ui/src/combobox.js
@@ -43,19 +43,19 @@ export default function (Alpine) {
     Alpine.magic('comboboxOption', el => {
         let data = Alpine.$data(el)
 
-        let optionEl = Alpine.findClosest(el, i => i.__optionKey)
+        let optionEl = Alpine.findClosest(el, i => Alpine.$data(i).optionKey)
 
         if (! optionEl) throw 'No x-combobox:option directive found...'
 
         return {
             get isActive() {
-                return data.__context.isActiveKey(optionEl.__optionKey)
+                return data.__context.isActiveKey(Alpine.$data(optionEl).optionKey)
             },
             get isSelected() {
                 return data.__isSelected(optionEl)
             },
             get isDisabled() {
-                return data.__context.isDisabled(optionEl.__optionKey)
+                return data.__context.isDisabled(Alpine.$data(optionEl).optionKey)
             },
         }
     })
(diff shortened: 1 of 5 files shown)


calebporzio:
Ok thanks for the extra context. I was able to reproduce this and can confirm your fix is good.

I added tests and found a couple problems with the current implementation, but a made a few tweaks and hopefully we're good now!

Maybe you can submit a failing test PR to Livewire for the other issue and we can tackle that as well? Thanks @gdebrauwer for the effort and contributions!

gdebrauwer:
@calebporzio I created a PR with some failing tests: https://github.com/livewire/livewire/pull/7776

PR #3995, opened 2024-01-24
:bug: Fixes reinitializing moved elements

ekwoka:
Fixes #3994

Introduced in #3951

Adds a test for it :)

### DIFF
--- a/packages/alpinejs/src/mutation.js
+++ b/packages/alpinejs/src/mutation.js
@@ -196,6 +196,7 @@ function onMutate(mutations) {
     for (let node of addedNodes) {
         // If the node was eventually removed as part of one of his
         // parent mutations, skip it
+        if (removedNodes.has(node)) continue
         if (! node.isConnected) continue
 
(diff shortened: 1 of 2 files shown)


SimoTod:
I think this will reintroduce the issue for the guy who originally made the change. I suspect the problem is that he sets the init flag after processing the directives while he should do it at the very beginning (inside the if).

ekwoka:
I do agree.

but I think fixing this should be priority over ensuring that case also works.

I will investigate an additional PR that can handle both situations more easily.

But it seems clear to me that now it is likely breaking all elements that move. or at least POTENTIALLY breaking them depending on the side effects of their directives.

calebporzio:
Thanks @ekwoka, I think I actually removed that line by mistake when I was refactoring that PR? I changed addedNodes and removedNodes to `Set`s to ensure uniqueness, but we also need this check in there for "move" scenarios.

Thanks again. Will tag a patch release asap

PR #4005, opened 2024-01-29
Ensure `$listboxOption.isSelected` / `$comboboxOption.isSelected` works if item value is `false` or `0`

gdebrauwer:
In my application, I noticed that the selected option is not styled correctly when the value of the selected option is `false` or `0`.

https://github.com/alpinejs/alpine/assets/22586858/e3495434-e47d-4858-a6aa-7a09fa79a7cc

This is caused by the `isSelected` getter on `$listboxOption` / `$comboboxOption`. That getter always returns false if the item's value is falsy. That is not correct if the item's value is false or 0. This PR fixes that.

### DIFF
--- a/packages/ui/src/combobox.js
+++ b/packages/ui/src/combobox.js
@@ -222,7 +222,7 @@ function handleRoot(el, Alpine) {
                     let item = this.__context.getItemByEl(el)
 
                     if (! item) return false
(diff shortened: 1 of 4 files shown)


calebporzio:
Thanks for this! Super valuable!

PR #4008, opened 2024-01-30
Fix context of listbox to prevent error when using wire:navigate

gdebrauwer:
When I combine Livewire's `wire:navigate` and listbox components, I get the following error when navigating:

Screenshot 2024-01-30 at 14 15 03

The combobox component does not have that issue. When looking at the source code of the combobox, I saw that the `__activateSelectedOrFirst` method is directly accessed on `this` instead of `this.$data`. This PR changes that in the listbox component in the same way, as that fixes the error.


calebporzio:
Ideally we have a test for this, but I get that it happens because of Livewire, so I'll merge. Thanks!

PR #4031, opened 2024-02-13
Add failing test for keyboard controls with `x-teleport`

Ahrengot:
Adds failing test for #4030 

If the `this.$refs.__items` are replaced with `this.$el` in the following section then the test passes, so it seems to be related to teleported refs.

https://github.com/alpinejs/alpine/blob/84fd08d62c1041ed73358a4a0d337656d8d488cd/packages/ui/src/menu.js#L98-L110

### DIFF
--- a/tests/cypress/integration/plugins/ui/menu.spec.js
+++ b/tests/cypress/integration/plugins/ui/menu.spec.js
@@ -219,6 +219,54 @@ test('keyboard controls',
     },
 )
 
+test('keyboard controls with x-teleport',
+    [html`
+        <div x-data x-menu>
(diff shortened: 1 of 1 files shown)


calebporzio:
Thanks for this @Ahrengot. It was a problem with `$refs` not working with teleports. It's been fixed.

In the future, when you submit a pull request, can you check out a new branch (not `main`) on your fork. That way I can push commits directly to this PR.

Thanks again!

Ahrengot:
Sure thing. I rarely make pull requests to public repos, and assumed you had control over the branch once it was submitted.

PR #4032, opened 2024-02-13
Update intersect.md

simonferndriger:
Clarified x-intersect:leave and added `x-intersect:leave.full`

### DIFF
--- a/packages/docs/src/en/plugins/intersect.md
+++ b/packages/docs/src/en/plugins/intersect.md
@@ -89,7 +89,7 @@ You may choose to use this for clarity when also using the `:leave` suffix.
 <a name="x-intersect-leave"></a>
 ### x-intersect:leave
 
-Appending `:leave` runs your expression when the element leaves the viewport:
+Appending `:leave` runs your expression when the element leaves the viewport. By default, this means the element has left the viewport *completely*. Use `x-intersect:leave.full` to run your expression when the element *starts leaving* the viewport.
 
 ```alpine
 <div x-intersect:leave="shown = true">...</div>


calebporzio:
Interesting, I would expect `x-intersect:leave.full` to be the opposite of you're description.

I expect that it would only trigger when the entire element has left the viewport, not a part of it.

Is that the case? If so, we might want to consider changing the default behavior if decide it's a bug and not just a breaking change

simonferndriger:
Yes, it is very interesting indeed 😅 But unfortunately true, see https://github.com/alpinejs/alpine/discussions/4027#discussioncomment-8453883 for more details about it

ekwoka:
Yes, the threshold used in intersection observer is not concerned with direction. It is simply "this much of the element is in or not".

Leave is then just about "is it currently intersecting when the observer is fired or not".

could have a `leave + full` do `0` instead of `.99`. Then should `leave` without `full` do `.99`? Or is `full` more of a noop on `leave`, like just does nothing?

Publically, `leave + full` doesn't have a ton of use: https://github.com/search?q=x-intersect%3Aleave.full&type=code

So those would break, or they don't realize it's not doing what they expected.

Leave without `full` has a bit more usage: https://github.com/search?q=x-intersect%3Aleave&type=code

packages/docs/src/en/plugins/intersect.md
```diff
@@ -89,7 +89,7 @@ You may choose to use this for clarity when also using the `:leave` suffix.
 <a name="x-intersect-leave"></a>
 ### x-intersect:leave
 
-Appending `:leave` runs your expression when the element leaves the viewport:
+Appending `:leave` runs your expression when the element leaves the viewport. By default, this means the element has left the viewport *completely*. Use `x-intersect:leave.full` to run your expression when the element *starts leaving* the viewport.
```
ekwoka, line 92: I'd probably like including a link to IntersectionObserver documentation somewhere to help provide more context for people groking what thresholds and margins actually are.

simonferndriger:
Sorry, not sure how to change my pull request... can somebody else make the changes suggested by @ekwoka?

ekwoka:
You just make new commits to the branch in your fork.

calebporzio:
Thank you for the clarification @ekwoka, this is indeed unintuitive in combination: "leave + full", but it makes more sense with "enter + full", so yeah, let's keep it. I'll merge your other PR. Thanks!

PR #4038, opened 2024-02-15
Allows access to methods when class instance used for x-data

ekwoka:
Currently, if a class instance is used as the data scope, instance methods (and inherited properties *shudders*) are not accessible in expressions.

This is due to only checking for own properties in `mergeProxies`.

This changes this to use `Reflect.has` to better *reflect* the actual operation being performed, and uses that same behavior in other places doing this check for consistency. (Of note, I found in some testing that using `Reflect.ownKeys` in the `ownKeys` trap caused some very strange call stack issues).

(Note on using Class instances in general: actually private instance properties/methods will not be accessible from other methods on that class instance due to how the proxy reflection actually works).

### DIFF
--- a/packages/alpinejs/src/scope.js
+++ b/packages/alpinejs/src/scope.js
@@ -48,7 +48,7 @@ let mergeProxyTrap = {
         if (name == Symbol.unscopables) return false;
 
         return objects.some((obj) =>
-            Object.prototype.hasOwnProperty.call(obj, name)
+            Reflect.has(obj, name)
         );
     },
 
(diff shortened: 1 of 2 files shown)


calebporzio:
Thanks for this contribution @ekwoka - can you give me more context into where the need arose for this? A GH discussion or something? Has there been issues faced by multiple people relating to this?

ekwoka:
It was in the Discord. I think it came up twice there.

I believe there was also a discussion here but not related directly to this as an issue, so I'm not positive how we'd find it.

I think the first commit of this PR is sensible even without it as a major concern, since it is less code and more appropriate use of Reflection.

But yes, the second and third are just about addressing some potential future issues, that may just not be very important since the likelihood of such issues is very very very low.

calebporzio:
Mmmm yeah very interesting possibilities @ekwoka.

Do you mind removing the second commit then and we can tackle that as an independent task at some point? (sorry I'm OOO and will merge this if it's merge-ready) Thanks!

ekwoka:
Done.

calebporzio:
Thanks!

PR #4039, opened 2024-02-17
Handles Object Prototype Methods in Proxy

ekwoka:
Continuation of #4038 

This addresses a potential issue that may arise (and then another issue that fix may lead to).

These "fixes" are more for posterities sake, as these becoming issues hinges on #4038 being merged, but also the likelihood of them being issues is quite small.

This can just be in draft (or even closed) for history sake if those issues do come to fruition.

Basically, #4038 would expose prototype methods to the Proxy used for expressions, so expressions like `valueOf` would point to the data object method `valueOf` while the previous behavior would be the `window.valueOf`.

This checks if the things being accessed is an Object.prototype method and skips it, but to allow someones own object to use those same names, we check if the object has that name itself.

These seem like low likelihood/low impact issues, but are logical results of #4038

### DIFF
--- a/packages/alpinejs/src/scope.js
+++ b/packages/alpinejs/src/scope.js
@@ -46,10 +46,12 @@ let mergeProxyTrap = {
 
     has({ objects }, name) {
         if (name == Symbol.unscopables) return false;
-        if (name in Object.prototype) return false;
+        const isPrototypeMethod = name in Object.prototype;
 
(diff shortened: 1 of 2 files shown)


calebporzio:
Thanks for this. I'll close this for now and we can return to it if issues arise. Thanks @ekwoka for the great explanation.

PR #4040, opened 2024-02-18
update livewire url to v3

aftandilmmd:


joshhanley:
@aftandilmmd This actually works in v3 so the warning can probably be removed.

calebporzio:
Thanks! Good call Josh - just removed it

PR #4061, opened 2024-02-29
Improve x-transition docs with more examples and potential pitfall

jslvtr:
Added examples in opacity transition similar to those in scale for modifying duration and delay.

Added new section for what happens if you add multiple conflicting transition directives with `x-transition` and `x-transition:enter` and related.

Related to #4059 .

### DIFF
--- a/packages/docs/src/en/directives/transition.md
+++ b/packages/docs/src/en/directives/transition.md
@@ -81,10 +81,20 @@ The above example will delay the transition and in and out of the element by 50 
 
 By default, Alpine's `x-transition` applies both a scale and opacity transition to achieve a "fade" effect.
 
-If you wish to only apply the opacity transition (no scale), you can accomplish that like so:
+If you wish to only apply an opacity transition (no scale), you can accomplish that like so:
 
 ```alpine
 <div ... x-transition.opacity>
+```
+
+To customize the opacity transition properties, you'd use the same modifiers as above, but preceded with `.opacity`, like so:
+
+```alpine
+<div ...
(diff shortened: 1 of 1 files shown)


packages/docs/src/en/directives/transition.md
```diff
@@ -81,12 +81,22 @@ The above example will delay the transition and in and out of the element by 50
...
 <div ... x-transition.opacity>
 ```
 
+To customize the opacity transition properties, you'd use the same modifiers as above, but preceded with `.opacity`, like so:
+
+```alpine
+<div ...
+    x-transition.opacity
+    x-transition.opacity.duration.500ms
+    x-transition.opacity.delay.50ms
```
ekwoka, line 96: I'd worry this implies multiple x-transition is allowed.

jslvtr: I might be missing some understanding, as I thought that you could do this:

```
x-transition
x-transition.scale.80
```

To keep the scale + opacity transition, but customize the scale of the scale transition.

Having tried this, now I see you can't.

How could that be done, keep the opacity+scale, but customize one or both of them?

calebporzio:
I'm not sure this PR adds clarity. I'd be open to a smaller adjustment or simple warning to the docs about a potential pitfall otherwise. Thanks for your time, but closing this.

packages/docs/src/en/directives/transition.md
ekwoka: `x-transition.opacity.scale.80` probably...

PR #4063, opened 2024-03-01
bug fix __noscroll param

theGC:
fix to allow preventScroll to be set to true when element passed onScroll() method


ekwoka (approved):
Oh wow!

Obligatory: TypeScript would have caught this.

calebporzio:
lol @ekwoka, it had to be said!

Thanks @theGC

PR #4091, opened 2024-03-12
Fixed persist crash when the stored value is undefined

wahibimoh:
Fixed persist crash when the stored value is undefined. This bug crashes Alpine

### DIFF
--- a/packages/persist/src/index.js
+++ b/packages/persist/src/index.js
@@ -65,7 +65,12 @@ function storageHas(key, storage) {
 }
 
 function storageGet(key, storage) {
-    return JSON.parse(storage.getItem(key, storage))
+    try{
+      return JSON.parse(storage.getItem(key, storage));
+    }catch(e){
+      console.warn("Alpine: $persist "+key+" is not valid JSON");
+      return undefined; 
+    }
 }
 
 function storageSet(key, value, storage) {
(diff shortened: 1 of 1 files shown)


calebporzio:
What do you think about instead of adding try/catch, we just get the value, check for undefined and if so, return that, otherwise return `JSON.parse()`?

This way it works as expected and doesn't have a console.warn in there (which I try to avoid as they take up an abnormal amount of size in the minified bundle)

ekwoka:
That was my thought as well.

Whenever we check if the item even exists int he storage, we also check if it's undefined.

The one thing I think might need to be determined is the expected behavior of setting a persisted value to undefined. Should it be treated as non existent and reinitialize, or be treated as the value undefined.

I would think setting to null should be used for explicit, and undefined used to resetting.

calebporzio:
@ekwoka, I'm not sure I understand. I updated the PR and merged. Please re-submit with changes if you think there's a better way. Thanks!

wahibimoh:
thanks for the refactor and merge, I am just thinking is undefined the only value that could cause an error JSON.parse? If yes, then it the refactor is perfect, if there are other values, then we may have to use try/catch or check for these values as well.

So far, I only faced this crash with undefined value in production.

ekwoka:
Anything else would have had to be malformed in the first place...which would most likely only be possible if someone directly modified the storage. It wouldn't happen from within persist.

PR #4093, opened 2024-03-14
Throw error when passing anything other than a function to `Alpine.plugin`

mattvague:
Currently if you accidentally pass anything other than a function to `Alpine.plugin` you get this cryptic error:

image

This has happened to me a few times when I misspell my import e.g.

```
import myPlugin from './wrong-name'

Alpine.plugin(myPlugin)
```

This PR changes the plugin function so that it throws nice errors when this happens:

image

I also added integration tests for plugins and modified the test suite setup so it's possible to assert a specific error is thrown.

### DIFF
--- a/packages/alpinejs/src/plugin.js
+++ b/packages/alpinejs/src/plugin.js
@@ -3,5 +3,9 @@ import Alpine from "./alpine";
 export function plugin(callback) {
     let callbacks = Array.isArray(callback) ? callback : [callback]
 
-    callbacks.forEach(i => i(Alpine))
+    callbacks.forEach(i => {
+      if (typeof i === undefined) { throw new Error(`Alpine.plugin(...) expects a callback function but got undefined. Did you misspell your plugin import?`)}
+      if (typeof i !== 'function') { throw new Error(`Alpine.plugin(...) expects a callback function but got ${i}`)}
+      i(Alpine)
+    })
 }
--- a/packages/alpinejs/src/plugin.js
+++ b/packages/alpinejs/src/plugin.js
@@ -4,7 +4,6 @@ export function plugin(callback) {
     let callbacks = Array.isArray(callback) ? callback : [callback]
 
     callbacks.forEach(i => {
-      if (typeof i === undefined) { throw new Error(`Alpine.plugin(...) expects a callback function but got undefined. Did you misspell your plugin import?`)}
       if (typeof i !== 'function') { throw new Error(`Alpine.plugin(...) expects a callback function but got ${i}`)}
       i(Alpine)
     })
--- a/tests/cypress/integration/custom-plugins.spec.js
+++ b/tests/cypress/integration/custom-plugins.spec.js
@@ -1,0 +1,48 @@
+import { haveText, haveAttribute, html, test } from '../utils'
+
+test('can register custom directive from plugin',
+    [html`
+        <div x-data>
+            <span x-foo:bar.baz="bob"></span>
+        </div>
+    `,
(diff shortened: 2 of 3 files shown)


ekwoka:
Why throw at all?

Log the error and skip it.

You can avoid this by using typescript, or eslint...or any other assorted tools as well.

packages/alpinejs/src/plugin.js
```diff
@@ -3,5 +3,9 @@ import Alpine from "./alpine";
 export function plugin(callback) {
     let callbacks = Array.isArray(callback) ? callback : [callback]
 
-    callbacks.forEach(i => i(Alpine))
+    callbacks.forEach(i => {
+      if (typeof i === undefined) { throw new Error(`Alpine.plugin(...) expects a callback function but got undefined. Did you misspell your plugin import?`)}
```
ekwoka, line 7: I don't really think this check is needed at all. What does this really tell someone that the other doesn't?

packages/alpinejs/src/plugin.js
```diff
@@ -3,5 +3,9 @@ import Alpine from "./alpine";
...
+      if (typeof i !== 'function') { throw new Error(`Alpine.plugin(...) expects a callback function but got ${i}`)}
```
ekwoka, line 8:
```suggestion
      if (typeof i !== 'function') { 
         return console.error(`Alpine.plugin(...) expects a callback function but got ${i}`)
      }
```

ekwoka (changes requested):

packages/alpinejs/src/plugin.js
mattvague: Yeah maybe not, fair enough

mattvague:
>Why throw at all?

>Log the error and skip it.

I guess it depends on your philosophy around errors, but for my situation I think I'd definitely prefer an error to be thrown so that its fail my automated tests and whatnot. I can't see why I'd want the rest of my page to continue to work as normal if a plugin that I'm adding isn't actually working? A helpful console.error would be better than what happens now though so if you think that'd be better I'm good with it

>You can avoid this by using typescript, or eslint...or any other assorted tools as well.

Very true. I am using typescript and it does tell me about this, but in my setup TS only warns me in VSCode about and doesn't actually stop my build which leads me to finding this cryptic error in my browser console.

packages/alpinejs/src/plugin.js
ekwoka: But you should probably solve that in your tooling. Like mispelling should cause issues the moment you type it. You shouldn't be making it to running in the first place.

You can use strict for JS to yell at you instantly, and you ca n use eslint or typescript to point out you are accessing something that doesn't exist as well.

ekwoka:
> I guess it depends on your philosophy around errors, but for my situation I think I'd definitely prefer an error to be thrown so that its fail my automated tests and whatnot.

It just feels like not a good "stop the world" case that exceptions really are for.

Just like we don't want Alpine to stop running because one expression on one element is malformed.

calebporzio:
I hesitate to add "log" messages like this for 2 reasons:

1) hardcoded strings (like log messages) can't be minified and therefore disproportionately increase the bundle size
2) it feels like if we warn about callbacks in this case, the argument could be made to warn about type differences all over the codebase which is obviously not ideal.

I think because "I am using typescript and it does tell me about this" we are covered well enough and can close this.

Thanks for being willing to contribute, especially with tests.

Thanks, but closing for now. If there's more community demand, we can re-open.

PR #4099, opened 2024-03-18
Remove hidden from booleanAttributes

IHIutch:
Fixes: https://github.com/alpinejs/alpine/discussions/4086

This fix will allow you to pass strings to `:hidden` which will allow you to use `hidden="until-found"`. 

More information on [hidden until found](https://developer.chrome.com/docs/css-ui/hidden-until-found).

### DIFF
--- a/packages/alpinejs/src/utils/bind.js
+++ b/packages/alpinejs/src/utils/bind.js
@@ -150,7 +150,7 @@ function isBooleanAttr(attrName) {
     // As per HTML spec table https://html.spec.whatwg.org/multipage/indices.html#attributes-3:boolean-attribute
     // Array roughly ordered by estimated usage
     const booleanAttributes = [
-        'disabled','checked','required','readonly','hidden','open', 'selected',
+        'disabled','checked','required','readonly','open', 'selected',
         'autofocus', 'itemscope', 'multiple', 'novalidate','allowfullscreen',
         'allowpaymentrequest', 'formnovalidate', 'autoplay', 'controls', 'loop',
         'muted', 'playsinline', 'default', 'ismap', 'reversed', 'async', 'defer',
--- a/tests/cypress/integration/directives/x-bind.spec.js
+++ b/tests/cypress/integration/directives/x-bind.spec.js
@@ -46,6 +46,10 @@ test('non-boolean attributes set to null/undefined/false are removed from the el
             <span visible="true" x-bind:visible="null">null</span>
             <span visible="true" x-bind:visible="false">false</span>
             <span visible="true" x-bind:visible="undefined">undefined</span>
+
+            <span hidden="true" x-bind:hidden="null">null</span>
+            <span hidden="true" x-bind:hidden="false">false</span>
+            <span hidden="true" x-bind:hidden="undefined">undefined</span>
         </div>
     `,
     ({ get }) => {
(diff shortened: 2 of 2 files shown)


SimoTod:
The feature is experimental and not supported by some of the major browsers. It might be too early. https://caniuse.com/mdn-html_global_attributes_hidden_until-found_value

IHIutch:
True, but this also isn't a breaking change. Boolean values will still work as expected

SimoTod:
I think it's a breaking change. currently when the binding resolve to false, Alpine removes the attribute. With your change, I think it will add `hidden="false"` which browsers will just render as if the element had the hidden attribute set.

SimoTod:
I think it could be a breaking change. Not sure how it behaves with a falsy values. I believe Alpine currently removes the attribute.

SimoTod:
Maybe I'm wrong, with a quick codeen, it seems to work in the same way. You probably need to add some tests, though

IHIutch:
Correct, it is removed when falsy because its not apart of `attributeShouldntBePreservedIfFalsy `

https://github.com/alpinejs/alpine/blob/9c796179f696706ca9091e1b8608c759be71f025/packages/alpinejs/src/utils/bind.js#L163-L165

Happy to add some tests, if you like.

SimoTod:
It's more for the maintainer when he'll review the PR. It's more likely to be merged if there are tests showing that there are no breaking changes.

ekwoka:
Definitely second including tests just to be  *more* sure :)

IHIutch:
Okay, I added to 2 existing tests. The first checks the hidden attribute specifically to ensure backward compatibility. Then I updated to the `aria-pressed/checked/expanded/selected` test, which seemed like the inverse concern, but only tested 1 of the 4 attributes in `attributeShouldntBePreservedIfFalsy`. (Which, thinking about now, is a weird name for attributes that _are_ preserved 😉)

calebporzio:
Thanks for adding tests.

I'm sure people rely on the existence of `<div hidden>` in their elements (toggled using Alpine).

Specifically for CSS like this:

```css
div[hidden] {
    display: none;
}
```

So my biggest question is: does this PR break the above scenario?

First, to be crystal clear about that changes in this PR:

**Before this PR change:**
```html
<div x-bind:hidden="true">
<!-- Renders to: -->
<div x-bind:hidden="true" hidden="hidden">

<div x-bind:hidden="false">
<!-- Renders to: -->
<div x-bind:hidden="false">

<div x-bind:hidden="'until-found'">
<!-- Renders to: -->
<div x-bind:hidden="'until-found'" hidden="hidden">
```

**After this PR change:**
```html
<div x-bind:hidden="true">
<!-- Renders to: -->
<div x-data="" x-bind:hidden="true" hidden="true">

<div x-bind:hidden="false">
<!-- Renders to: -->
<div x-data="" x-bind:hidden="false">

<div x-bind:hidden="'until-found'">
<!-- Renders to: -->
<div x-bind:hidden="'until-found'" hidden="until-found">
```

After evaluating each, I think this is a safe change.

I'll leave this here for @SimoTod and @ekwoka to weigh in on before merging.

Thanks!

ekwoka:
Is that the result?

It should still render `hidden` for `true` since `true` is not a canonical *spec-compliant* value.

It's same result, though, for now.

SimoTod:
I think it's "safe enough". Minor thing but it's worth knowing it, 'hidden="true"' won't pass the W3C validator and I remember some people complaining about that for other "non-standard" things in the past.

calebporzio:
Good points. I agree, it's "safe enough" @SimoTod. Interesting that hidden="true" won't pass a validator. So that's a bummer, but I suppose that's a further problem after this one (because we should at least make a change like this to allow custom values for "hidden").

I'm going to merge this.

PR #4101, opened 2024-03-19
Fix fill modifier for radio buttons

willrowe:
This fixes filling model data based on which radio button is checked when the page loads. A very simple example can be found [here](https://codepen.io/willrowe/pen/BaEpRyJ) and this PR also includes a test.

In the Codepen, I expected that `type` in the Alpine data would be set to 'second' after page load and that the 'Second' radio button would be checked. However, `type` is set to 'first' and the 'First' radio button is checked instead.

I did a little bit of debugging and figured out that a `change` event is being triggered when the `.fill` modifier is present. Removing it from the first radio button (second example in the Codepen) makes it work as expected, but it should still work when the modifier is present by filling based on which radio button is currently checked.

The `getInputValue` function has been updated to take into account whether a radio button is checked, if not it will return the `currentValue`. I loosely based the changes on how checkboxes are handled.

I tested it with checkboxes (third example in the Codepen) and that appears to be working correctly.

### DIFF
--- a/packages/alpinejs/src/directives/x-model.js
+++ b/packages/alpinejs/src/directives/x-model.js
@@ -168,14 +168,16 @@ function getInputValue(el, modifiers, event, currentValue) {
                 return option.value || option.text
             })
         } else {
+            const newValue = event.target.value
+
             if (modifiers.includes('number')) {
-                return safeParseNumber(event.target.value)
+                return safeParseNumber(newValue)
             } else if (modifiers.includes('boolean')) {
-                return safeParseBoolean(event.target.value)
+                return safeParseBoolean(newValue)
             } else if (modifiers.includes('trim')) {
-                return event.target.value.trim()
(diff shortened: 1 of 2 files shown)


ekwoka:
This is a good bug find!

Would it not be simpler to only dispatch the fill/change event on a checked radio?

Or is there another context in which a radio could dispatch a change event when unchecked?

willrowe:
@ekwoka I made the least intrusive change possible. I thought the same thing about the change event, but when debugging I discovered the `getInputValue` function and saw that it considered checkbox groups there, so it made the most sense to consider radio button groups in the same way. You can see in that function that it checks whether checkboxes are checked when returning the value.

ekwoka:
This is true. The change you made is definitely inline with the existing decisions.

calebporzio:
Perfect, thanks!

PR #4104, opened 2024-03-20
Add missing border to fill example

bb:
Example input for `.fill` is missing a border:
CleanShot 2024-03-20 at 15 28 47@2x

This PR adds the border and makes it look like throttle above.


calebporzio:
Makes sense, thanks!

PR #4112, opened 2024-03-27
Update dispatch.md, add missing x-data to first two examples

buddhaCode:
When copy & pasting one of the first two example, they didn't work.

### DIFF
--- a/packages/docs/src/en/magics/dispatch.md
+++ b/packages/docs/src/en/magics/dispatch.md
@@ -8,7 +8,7 @@ title: dispatch
 `$dispatch` is a helpful shortcut for dispatching browser events.
 
 ```alpine
-<div @notify="alert('Hello World!')">
+<div x-data @notify="alert('Hello World!')">
     <button @click="$dispatch('notify')">
         Notify
     </button>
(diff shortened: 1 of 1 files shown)


SimoTod:
I think it was discussed already. X-data is required for every directive to work but it doesn't have to be on the element, it could be on any of the parent elements. 

It's not on any of the examples in the documentation because people would just copy paste without understanding and ending up with a page where every single html tag has an x-data attribute.

If we update this example, we should update all of them.

ekwoka:
Yeah, both ways produce opportunity for misunderstanding.

If it's on every example then people may think it is required all over the place.

I'd it's not there people may think it's not required.

But if they read the docs, which aren't that long or complicated, it should be quite clear.

I'd be in favor of evaluating whether the docs that cover that aspect (of component definition) being reviewed to see if it's unclear.

If anything, not having it in the example can make the examples better, since the misunderstanding will be made clear to them, as opposed to being wrong and never noticing.

buddhaCode:
Oh sorry. I wasn't aware, that this was already discussed.

I already used alpine a bit. But I always forget about the `x-data`. Reading the docs, I'm always not sure if a need `x-data`.

I think it would be a good idea to highlight the x-data item in the menu with a badge like "read first" or "you need this". And but the section about the data less components first.

PR #4137, opened 2024-04-09
Add "sort" plugin

calebporzio:


ekwoka:
Is x-sort preferable to x-sortable?

SimoTod:
My 2 cents. People will naturally try to use it with x-for. You may want to provide an example of how the 2 can work together.

PR #4141, opened 2024-04-12
Documentation issue: Event name in code snippet and description doesn't match in dispatch.md

AkosLukacs:
In the documentation about dispatch there is a typo. The text says:

> The first example won't work because when **`custom-event`** is dispatched, it'll propagate to its common ancestor, the `div`, not its sibling, the `<span>`. The second example will work because the sibling is listening for `notify` at the `window` level, which the custom event will eventually bubble up to.

But the code snippet doesn't dispatch `custom-event`, it dispatches a `notify` event: `@click="$dispatch('notify')"`

### DIFF
--- a/packages/docs/src/en/magics/dispatch.md
+++ b/packages/docs/src/en/magics/dispatch.md
@@ -66,7 +66,7 @@ Notice that, because of [event bubbling](https://en.wikipedia.org/wiki/Event_bub
 </div>
 ```
 
-> The first example won't work because when `custom-event` is dispatched, it'll propagate to its common ancestor, the `div`, not its sibling, the `<span>`. The second example will work because the sibling is listening for `notify` at the `window` level, which the custom event will eventually bubble up to.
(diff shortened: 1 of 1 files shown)


ekwoka:
Please include a description of what this is doing.

The PR title doesn't really describe what's happening

AkosLukacs:
In the documentation about dispatch there is a typo. The text says:

> The first example won't work because when **`custom-event`** is dispatched, it'll propagate to its common ancestor, the `div`, not its sibling, the `<span>`. The second example will work because the sibling is listening for `notify` at the `window` level, which the custom event will eventually bubble up to.

But the code snippet doesn't dispatch `custom-event`, it dispatches a `notify` event: `@click="$dispatch('notify')"`

ekwoka:
Can you put that in the PR description?

So that the PR description describes what the PR does and why?

calebporzio:
Good catch thanks!

calebporzio:
Also @ekwoka , you're right most PRs need thorough descriptions, but a small/obvious docs error doesn't really IMO.

PR #4151, opened 2024-04-16
Remove extra destroyTree call

c4710n:
`destroyTree` is hooked into `mutation.js` by [lifecycle.js#L22](https://github.com/alpinejs/alpine/blob/main/packages/alpinejs/src/lifecycle.js#L22), and will be called [when an element is removed](https://github.com/alpinejs/alpine/blob/cd171328bffdfc805256d1ada84522898ef5b107/packages/alpinejs/src/mutation.js#L179).

If we [call it again](https://github.com/alpinejs/alpine/blob/cd171328bffdfc805256d1ada84522898ef5b107/packages/alpinejs/src/mutation.js#L181), then we call it twice.

This PR removes the extra call, and it also eliminates the dependency-cycle between `lifecycle.js` and `mutation.js`.

### DIFF
--- a/packages/alpinejs/src/mutation.js
+++ b/packages/alpinejs/src/mutation.js
@@ -1,5 +1,3 @@
-import { destroyTree } from "./lifecycle"
-
 let onAttributeAddeds = []
 let onElRemoveds = []
 let onElAddeds = []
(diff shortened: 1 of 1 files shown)


c4710n:
I didn't find this issue through actual usage. I discovered it while reading the source code, as I was very interested in how Alpine is built.

calebporzio:
Gotcha, good eye! I tested and you're right, it was getting called twice! Thanks

PR #4155, opened 2024-04-18
:bug: Removes value attribute when bound value is null

ekwoka:
Better handles Progress elements #4153 

Not entirely positive if this is the best way to approach this, but it seems more likely to handle cases for other elements with `value` attributes than the main inputs, where `null` is unlikely to be a legitimate value anyway.

Passes all current tests (aside from the flaky ones that seem to change every run), will add new test for this specific case soon.


calebporzio:
Thanks @ekwoka - however, I'm going to close this for this reason: https://github.com/alpinejs/alpine/discussions/4153#discussioncomment-9157493

PR #4159, opened 2024-04-19
Fix form reset for x-model radio inputs

bb:
- **Fix x-on with both self and once**
- **Fix x-on with both outside and once**
- **Fix form reset for x-model radio inputs**


bb:
I started out adding more and more conditions in the reset listener but settled on simply always using the `getInputValue` approach because I basically ended up duplicating all the conditions from there anyway. In between it looked like this:

```javascript
nextTick(() => el._x_model && (el.type !== 'radio' || el.checked) && (((el.type === 'checkbox' && Array.isArray(getValue())) || el.type === 'radio' || (el.tagName.toLowerCase() === 'select')) ? setValue(getInputValue(el, modifiers, { target: el }, getValue())) : el._x_model.set(el.value)))
```

now it's just
```javascript
nextTick(() => el._x_model && el._x_model.set(getInputValue(el, modifiers, { target: el }, getValue())))
```

I kept using the public setter (which was prepared to allow programmatic overriding of x-model as described in a comment below that line). However, there are some places in the code which do not use the overridable setter: cloning and when using `.fill`. So I wonder if there are more bugs hidden in this area. I recommend adding a test for overriding the setter to make sure there are none.

packages/alpinejs/src/directives/x-model.js
```diff
@@ -71,7 +71,8 @@ directive('model', (el, { modifiers, expression }, { effect, cleanup }) => {
 
     if (modifiers.includes('fill'))
         if ([undefined, null, ''].includes(getValue())
-            || (el.type === 'checkbox' && Array.isArray(getValue()))) {
+            || (el.type === 'checkbox' && Array.isArray(getValue()))
+            || (el.tagName.toLowerCase() === 'select' && el.multiple)) {
```
ekwoka, line 75:
```suggestion
            || (el.tagName === 'SELECT' && el.multiple)) {
```

ekwoka (commented):
Lots of improvements for some simple changes.

packages/alpinejs/src/directives/x-model.js
bb: While this is definitely fine for HTML4 and HTML5, I'm afraid it would break usage in XHTML documents. Not sure about SVG, never used Alpine in an SVG context (yet).

Also, there are two other places checking for this tagName using toLowerCase comparison. If it is changed, it should be changed in all places.

bb: You made me curious, so I verified. It does not seem to break XHTML documents, at least not in current Firefox, Chrome, nor Safari versions (on Desktop). It gets weird when embedding svg inline in XHTML, though.... but honestly, who has not switched to HTML5 within the last 15 or so years 😃

I don't have any old stuff or IE available for testing currently... so I'm quite sure either approach is fine here.

#### Example (validated using https://validator.w3.org/check) prints every element uppercased:

```xhtml
<?xml version="1.0" encoding="UTF-8" ?>
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.1//EN" "http://www.w3.org/TR/xhtml11/DTD/xhtml11.dtd">
<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
<head>
<title>x</title>
</head><body>
<div>
<select name="s">
<option>foo</option>
</select>
</div>
<script type="text/javascript">
    document.querySelectorAll("*").forEach(e => console.log(e.tagName))
</script>
</body></html>
```
prints every element uppercased.


Interestingly, when embedding SVG without explicit namespaces, tag names for svg are lowercase. When embedding SVG with explicit namespace, they're uppercase.

#### Tagnames svg and path are lowercase, while HTML elements stay uppercase, e.g. `SELECT`, `svg`

```xhtml
<?xml version="1.0" encoding="UTF-8" ?>
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.1//EN" "http://www.w3.org/TR/xhtml11/DTD/xhtml11.dtd">
<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
<head>
<title>x</title>
</head><body>
<div>
<select name="s">
<option>foo</option>
</select>
<svg viewBox="0 0 10 10" style="background-color:#fff" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" xml:space="preserve" x="0px" y="0px" width="10" height="10">
    <path d="M 3 3 L 3 7 L 7 7 L 7 3 L 3 3 Z" fill="#f00"/>
</svg>
</div>
<script type="text/javascript">
    document.querySelectorAll("*").forEach(e => console.log(e.tagName))
</script>
</body></html>
```

#### Tagnames uppercase with namespace prefix (e.g. `SVG:SVG`):

```xhtml
<?xml version="1.0" encoding="UTF-8" ?>
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.1//EN" "http://www.w3.org/TR/xhtml11/DTD/xhtml11.dtd">
<html xmlns="http://www.w3.org/1999/xhtml" xmlns:svg="http://www.w3.org/2000/svg" xml:lang="en" lang="en">
<head>
<title>x</title>
</head><body>
<div>
<select name="s">
<option>foo</option>
</select>
<svg:svg viewBox="0 0 10 10" style="background-color:#fff" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" xml:space="preserve" x="0px" y="0px" width="10" height="10">
    <svg:path d="M 3 3 L 3 7 L 7 7 L 7 3 L 3 3 Z" fill="#f00"/>
</svg:svg>
</div>
<script type="text/javascript">
    document.querySelectorAll("*").forEach(e => console.log(e.tagName))
</script>
</body></html>
```

calebporzio:
Thanks @bb! Thanks for all your work on this

PR #4170, opened 2024-04-23
Comma modifier to keyup and keydown events

mikko-ahonen:
I have a need to use comma as a separator in autocomplete I am implementing with Alpine. I guess this is an occasional need, as comma is a thousands separator in some countries and decimal separator in others.


bb (approved):

calebporzio:
Sounds good!

PR #4175, opened 2024-04-26
:bug: Masks model updates

ekwoka:
Solves #3691 

Not exactly pretty. I can do a more significant refactoring of the directive to make this cleaner, but this basically shows what steps aren't being handled.

Just taps into and overrides how x-model performs it's input value update to format the value and subsequently update to the models underlying datasource.


calebporzio:
Hey @ekwoka - can you provide a clearer explanation of what this PR is doing? Thanks. (also, clever approach: decorating the forceModelUpdate method)

ekwoka:
You got it! 👍

Tim-Wils:
@ekwoka Hey Eric, does this fix also fixes the bug with the native user-valid/user-invalid ? As per this discussion: https://github.com/alpinejs/alpine/discussions/4046

Tim-Wils:
Alright, that's a still a problem then. Thanks for this PR though!

godismyjudge95:
Ran into this issue just now.  Going to have to figure out an alternate solution until this PR is merged as I need the mask to apply both with manual and programmatic updates.

Harkirat444:
When will this PR be merged?

tests/cypress/integration/plugins/mask.spec.js
```diff
@@ -239,3 +239,18 @@ test('$money with custom decimal precision',
         get('#3').type('1234.5678').should(haveValue('1,234.567'))
     }
 )
+
+test('$mask should process the value when updated by x-model',
+    [html`
+        <div x-data="{value:55555}">
+            <input type="text" x-model=value x-mask:dynamic="$money($input)">
```
CleitonDeLima, line 246:
```suggestion
            <input type="text" x-model="value" x-mask:dynamic="$money($input)">
```

ariaieboy:
@calebporzio 
Can you look at this PR and the issue it solves?

The problem is that when we change the Alpine data using JS the mask won't apply on the changed value.

in the demo link that the author provided is obvious:
```html
<div
  x-data="masked"
  class="w-full flex gap-2 justify-center flex-col items-center"
>
  <input
    type="text"
    x-model="value"
    x-mask:dynamic="$money($input)"
    class="p-2 text-center"
  />
  <button
    @click="value = 123000"
    type="button"
    class="py-2 px-4 bg-indigo-800 text-indigo-200 rounded-lg"
  >
    Update data value
  </button>
</div>
```
we have an input that has an x-mask on it and we have a button that changes the input value when we clicn on it. 
the problem here is that when we click on the button the value sets to `123000` and it wont get masked. 
the value should get masked and return `123,000`

matthias-margin:
@CleitonDeLima anything we can do to help push this along? we're running into this issue as well with some money inputs that can be changed by the user or by code. The test case in this PR targets this issue exactly and seems to solve it.

tests/cypress/integration/plugins/mask.spec.js
ekwoka: For a test, I'm not sure this matters, since it's still to spec.

But okay

willrowe:
I've seen some quirky things happen because of this as well.

avatar2033:
When will this PR be merged? This is a serious problem

PR #4182, opened 2024-04-28
Add minified ESM distribution for browsers

jonkoops:
Adds a minified version of the ESM distribution that can be included directly in browsers. See #4181 for a background on this topic.

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -55,8 +55,8 @@ function bundleFile(package, file) {
             })
 
         },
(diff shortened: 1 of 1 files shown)


calebporzio:
Thanks for this PR @jonkoops - because this isn't something we can easily revert. I'm going to hold off until there's more demand.

If others agree this is needed, please re-open this PR

In the meantime, @ekwoka has some good suggestions in the referenced discussion

Thanks again!

PR #4191, opened 2024-05-03
Update sort.md

mynona:
Extended description with usage example for x-sort:key and how to use string values.


calebporzio:
I am supporting `x-sort:key` for backwards compatibility right now so I don't think I want to document it. Thanks!

PR #4193, opened 2024-05-03
Add 'panelWithoutClose' to Dialogs

kzipp:
The hope here is to keep a dialog/modal open when clicking outside of the panel. 

Not a default or breaking change, just another option.

I added an additional directive of panelWithoutClose, cloned the handlePanel function, and removed the @click.outside line from it.


ekwoka:
This can be implemented as a modifier like `.noclose` would probably be more inline with how Alpine works elsewhere

PR #4194, opened 2024-05-04
Higher visibility in docs about needing x-data for things to work

cdmistman:
The fact that `x-data` is needed in order to enable any Alpine.js behavior is pretty hidden in the docs. It'd be nice to make it more visible via the getting started guide.


calebporzio:
Great, this is a helpful addition. Thanks!

PR #4209, opened 2024-05-14
:sparkles: Enhances clicks with key modifiers

ekwoka:
Handles #4208 

Allows for the use of `ctrl` `meta`...etc keys on click and mouse event listeners.

Overall simple change to include clicks

Would a more significant refactor also enable the use of mouse-button modifiers? Maybe...

This also solved some *potential* bugs related to modifiers that could exist on a keydown that wouldn't be properly ignored.

Includes:
- [x] Tests
- [x] Doc Updates


calebporzio:
Love this! Thanks @ekwoka!

PR #4215, opened 2024-05-15
Update transition.md - changed span to div

bobclewell:
With a span, the transition effect is not obvious because the element doesn't move so the documentation looks broken.

I confirmed this in MacOS for Firefox, Chrome, and Safari.


calebporzio:
Good call - thanks!

PR #4238, opened 2024-05-25
[CSP] Add support for nested properties to CSP build

SimoTod:
This PR adds a minimal enhancement to the CSP build to support nested properties via dot notation (property accessor).

```html
<div x-data="counter">
    <button @click="foo.increment">Increment</button>
    <span x-text="foo.count"></span>
</div>
```

```js
Alpine.data('counter', () => ({
    foo: {
        count: 1,

        increment() {
            this.count++
        },
    },
}))
```

The build size is pretty much the same and it seems a reasonable change otherwise dev using the csp build are forced to flatten the component and put all properties at the top.

I only know about one user mentioning it though so feel free to close it if you think it's not worth it for now.

### DIFF
--- a/packages/csp/src/evaluator.js
+++ b/packages/csp/src/evaluator.js
@@ -28,11 +28,18 @@ function generateEvaluator(el, expression, dataStack) {
     return (receiver = () => {}, { scope = {}, params = [] } = {}) => {
         let completeScope = mergeProxies([scope, ...dataStack])
 
-        if (completeScope[expression] === undefined) {
-            throwExpressionError(el, expression)
-        }
+        let evaluatedExpression = expression.split('.').reduce(
+            (currentScope, currentExpression) => {
+                if (currentScope[currentExpression] === undefined) {
+                    throwExpressionError(el, expression)
+                }
 
-        runIfTypeOfFunction(receiver, completeScope[expression], completeScope, params)
+                return currentScope[currentExpression]
+            },
+            completeScope,
+        );
+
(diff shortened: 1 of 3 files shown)


packages/csp/src/evaluator.js
```diff
@@ -28,11 +28,18 @@ function generateEvaluator(el, expression, dataStack) {
...
         let completeScope = mergeProxies([scope, ...dataStack])
 
-        if (completeScope[expression] === undefined) {
-            throwExpressionError(el, expression)
-        }
-
-        runIfTypeOfFunction(receiver, completeScope[expression], completeScope, params)
+        let evaluatedExpression = expression.split('.').reduce(
+            (currentScope, currentExpression) => {
+                if (currentScope[currentExpression] === undefined) {
```
ekwoka, line 33:
```suggestion
                if (!currentExpression in currentScope) {
```

should help in scenarios where the thing is expected to be undefined.

But also, this will now exhibit a different behavior than non-csp in regards to this, where it would only error when trying to access a key on the thing that is undefined, but would be happy to return undefined.

ekwoka (changes requested):

packages/csp/src/evaluator.js
SimoTod: Good point, I kept the same check as the original version. Happy to change it if Caleb is cool with that.

calebporzio:
This is great. Thanks Simone!

PR #4253, opened 2024-06-04
Fix bug in X-Mask where user input that matched the non wildcard value were being stripped

robertmarney:
Closes https://github.com/alpinejs/alpine/discussions/4252

Found where a non wildcard value matched the user input the value was being stripped, I added an additional check before stripping and added a couple test cases to the jest tests.

### DIFF
--- a/packages/mask/src/index.js
+++ b/packages/mask/src/index.js
@@ -137,7 +137,7 @@ export function stripDown(template, input) {
         }
 
         for (let j = 0; j < inputToBeStripped.length; j++) {
-            if (inputToBeStripped[j] === template[i]) {
+            if (inputToBeStripped[j] === template[i] && template[i] !== inputToBeStripped[i]) {
                 inputToBeStripped = inputToBeStripped.slice(0, j) + inputToBeStripped.slice(j+1)
 
                 break;
--- a/tests/jest/mask.spec.js
+++ b/tests/jest/mask.spec.js
@@ -11,6 +11,8 @@ test('strip-down functionality', async () => {
     expect(stripDown('(999) 999-9999', '716 2256108')).toEqual('7162256108')
     expect(stripDown('(999) 999-9999', '(716) 2256108')).toEqual('7162256108')
     expect(stripDown('(999) 999-9999', '(716) 2-25--6108')).toEqual('7162256108')
+    expect(stripDown('+1 (999) 999-9999', '7162256108')).toEqual('7162256108')
+    expect(stripDown('ABC (999) 999-9999', '7162256108')).toEqual('7162256108')
 })
 
 test('formatMoney functionality', async () => {


packages/mask/src/index.js
```diff
@@ -137,7 +137,7 @@ export function stripDown(template, input) {
         }
 
         for (let j = 0; j < inputToBeStripped.length; j++) {
-            if (inputToBeStripped[j] === template[i]) {
+            if (inputToBeStripped[j] === template[i] && template[i] !== inputToBeStripped[i]) {
```
ekwoka, line 140: This can never evaluate to true.

It's saying "if these match and if they also don't match" which is impossible.

ekwoka:
There isn't actually a bug here.

tests/jest/mask.spec.js
```diff
@@ -11,6 +11,8 @@ test('strip-down functionality', async () => {
     expect(stripDown('(999) 999-9999', '716 2256108')).toEqual('7162256108')
     expect(stripDown('(999) 999-9999', '(716) 2256108')).toEqual('7162256108')
     expect(stripDown('(999) 999-9999', '(716) 2-25--6108')).toEqual('7162256108')
+    expect(stripDown('+1 (999) 999-9999', '7162256108')).toEqual('7162256108')
```
ekwoka, line 14:
```suggestion
    expect(stripDown('+1 (999) 999-9999', '+1 (716) 225-6108')).toEqual('7162256108')
```
this would be a more accurate test to what you wanted to "fix". Which will likely fail.

packages/mask/src/index.js
robertmarney: Hi @ekwoka - Thanks for the quick feedback.

I disagree, note that the new conditional is not using the same iterator for inputToBeStripped (we are using i not j)

From `mask.spec.js:L11`
```js

    expect(stripDown('(999) 999-9999', '716 2256108')).toEqual('7162256108')

template = "(999) 999-9999"
input = "716 2256108"

// when i = 5 and j = 3 we see inputToBeStripped[j] = " " and template[i] = " " (so first condition match), 
// then we see inputToBeStripped[i] = 2, 

// This matches the condition the space is removed.

```

robertmarney:
> There isn't actually a bug. It was behaving as expected.

> But this actually breaks behavior, but seemingly there weren't appropriate tests to catch that breakage.

To confirm provided a user input of `3133133113'` the expectation is the output is `+1(333)133-113`?

Perhaps its the special nature of the phone number input but we definitely see a large number of users that have the more 'raw' format stored in autocomplete (and we would like to leverage if we can and do with other masking tools) 

image

image

tests/jest/mask.spec.js
SimoTod: Just to clarify, which bug qre you trying to fix here? The example you posted (maska) behaves exactly as x-mask.
If you type 1, this generates `+1` and not `+1 (1`.

robertmarney: @SimoTod - The bug is the application of autocomplete values

SimoTod: Ok, that is the first time you mention it.
I think you need to describe the problem clearly otherwise when the maintainer looks at the issue, he will not understand what you are trying to solve.

Are we saying that the other plugins work differently depending on you entering the number vs copying vs using autoconplete? In that case, it looks like x-mask does a better job at staying consistent. It also sounds like it would be a lot of work to make it work the way you like.

If it's causing pains in your application, you can maybe disable the autocomplete?

robertmarney: Hey @SimoTod - Your feedback is fair, apologies for the lack of clarity.

If I can get a working solution would you accept a PR? 

As @ekwoka mentioned, as tests are passed, we might want to align on the spec (preference for Jest) of what we expect the mask to do for some of these more complex cases?

packages/mask/src/index.js
ekwoka: Ah yes, you're right. That's hard to read.

robertmarney:
Closing in favor of: https://github.com/alpinejs/alpine/pull/4260

PR #4278, opened 2024-06-28
:bug: Initializes Interceptors before store

ekwoka:
Solves #4248 

Stores were running their `init` method before interceptors were handled.

This handles interceptors first, just like how `data` does.


calebporzio:
This is great, I think I actually ran into this the other day making a dark mode store and wanting to persist the value. Legend ❤️

PR #4304, opened 2024-07-15
Add resize plugin: `x-resize`

calebporzio:
This PR adds a new Alpine plugin called "Resize" with the following API:

CleanShot 2024-07-15 at 12 22 37@2x

### DIFF
--- a/packages/docs/src/en/plugins/anchor.md
+++ b/packages/docs/src/en/plugins/anchor.md
@@ -1,5 +1,5 @@
 ---
-order: 5
+order: 7
 title: Anchor
 description: Anchor an element's positioning to another element on the page
 graph_image: https://alpinejs.dev/social_anchor.jpg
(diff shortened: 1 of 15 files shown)


ekwoka:
Damn, I just made this plugin myself.

I think you can safely use a similar strategy for the event delegation on the element as well with a weakmap.

The code I made

```ts
import type { PluginCallback } from 'alpinejs'

export const Resize: PluginCallback = (Alpine) => {
  const observerMap = new WeakMap<Element, (width: number) => void>()
  const observer = new ResizeObserver((entries) =>
    entries.forEach((entry) =>
      observerMap.get(entry.target)?.(entry.contentRect.width),
    ),
  )
  Alpine.directive(
    'resize',
    (el, { expression }, { evaluateLater, cleanup }) => {
      let evaluate = evaluateLater(expression)
      observerMap.set(el, (width: number) =>
        // biome-ignore lint/suspicious/noEmptyBlockStatements: Does not need to do anything
        evaluate(() => {}, { scope: { $width: width }, params: [width] }),
      )
      observer.observe(el)
      cleanup(() => {
        observerMap.delete(el)
        observer.unobserve(el)
      })
    },
  )
}

export default Resize
```

PR #4307, opened 2024-07-17
:memo: Updates docs with notes for x-data usage

ekwoka:
This comes up quite often, that someone is stumped by a directive not working, despite copying it from the docs because it doesn't have an `x-data` parent.

So this adds an extra note in the start here, and a note under the first example that doesn't use `x-data` for each directive (as needed).

It's a sad day.


calebporzio:
This is great, sorry for the lag on it. Thanks @ekwoka

PR #4308, opened 2024-07-19
UI x-radio: fix keyboard navigation when value of first/last radio option is null

gdebrauwer:
When the first or last `x-radio:option` has null as value, then keyboard navigation does not work correctly. It skips that option because it thinks that is already reached the end or beginning of the options array (caused by `next || all[0]` or `prev || all.slice(-1)[0]`). See screen recordings as a visual example.

https://github.com/user-attachments/assets/aa7cdeed-cf67-415b-b346-ab63b68296bc

https://github.com/user-attachments/assets/22d762b3-a1d0-4e6b-8a9a-43a4249bca3f

### DIFF
--- a/packages/ui/src/radio.js
+++ b/packages/ui/src/radio.js
@@ -106,8 +106,8 @@ function handleRoot(el, Alpine) {
                 __focusOptionNext() {
                     let option = this.__active
                     let all = this.__optionValues.filter(i => !this.__disabledOptions.has(i))
-                    let next = all[this.__optionValues.indexOf(option) + 1]
-                    next = next || all[0]
+                    let index = all.indexOf(option)
+                    let next = all.length > index + 1 ? all[index + 1] : all[0]
 
                     this.__optionElsByValue.get(next).focus()
                     this.__change(next)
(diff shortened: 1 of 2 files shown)


packages/ui/src/radio.js
```diff
@@ -106,17 +106,17 @@ function handleRoot(el, Alpine) {
                 __focusOptionNext() {
                     let option = this.__active
                     let all = this.__optionValues.filter(i => !this.__disabledOptions.has(i))
-                    let next = all[this.__optionValues.indexOf(option) + 1]
-                    next = next || all[0]
+                    let index = all.indexOf(option)
+                    let next = all.length > index + 1 ? all[index + 1] : all[0]
```
ekwoka, line 110:
```suggestion
                    let index = all.indexOf(option)
                    let next = all[(index + 1 + all.length) % all.length]
```

packages/ui/src/radio.js
```diff
@@ -106,17 +106,17 @@ function handleRoot(el, Alpine) {
...
                     this.__optionElsByValue.get(next).focus()
                     this.__change(next)
                 },
                 __focusOptionPrev() {
                     let option = this.__active
                     let all = this.__optionValues.filter(i => !this.__disabledOptions.has(i))
-                    let prev = all[all.indexOf(option) - 1]
-                    prev = prev || all.slice(-1)[0]
+                    let index = all.indexOf(option)
+                    let prev = index >= 1 ? all[all.indexOf(option) - 1] : all.slice(-1)[0]
```
ekwoka, line 119:
```suggestion
                    let index = all.indexOf(option)
                    let prev = all[(index - 1 + all.length) % all.length]
```

ekwoka (changes requested):
Interesting case!

PR #4310, opened 2024-07-22
make docs more generic to package managers

browner12:
for a long time NPM has basically been the de facto front end package manager. however, BunJS is making a strong push, and a good case for itself, so I think it'd be good for us to include it where appropriate.

we all know specific tooling comes and goes, so I've made the language *more generic* where appropriate, and refer to "a package manager" instead of specifically "NPM".

I've also then included the analogous `bun` command next to `npm` commands.

### DIFF
--- a/packages/docs/src/en/advanced/csp.md
+++ b/packages/docs/src/en/advanced/csp.md
@@ -14,7 +14,7 @@ In order to accommodate environments where this CSP is necessary, Alpine offer's
 <a name="installation"></a>
 ## Installation
 
-You can use this build by either including it from a `<script>` tag or installing it via NPM:
+You can use this build by either including it from a `<script>` tag or installing it via a package manager:
 
 ### Via CDN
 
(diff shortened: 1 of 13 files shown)


ekwoka:
`pnpm` would probably be a more important add, since it's the current most common standard for basically everything (and directly supported everywhere). Bun is new but much less common and not supported many places without extra code (cloudflare auto detects)

browner12:
I'm definitely okay with adding others, if you want to do a separate PR after/if this one gets merged.

I still think it's important to make the appropriate parts of the documentation language "tool agnostic".

What do you mean by "bun is not supported many places"? It's a binary that can be installed with a simple bash script.

ekwoka:
> What do you mean by "bun is not supported many places"? It's a binary that can be installed with a simple bash script.

I mean naturally. Like auto-detected, preinstalled.

> I'm definitely okay with adding others, if you want to do a separate PR after/if this one gets merged.

Probably better just to do it all in one, tbh.

browner12:
If someone wants to implement the tabbed approach, that'd be great. Just felt it was overkill.

I definitely think the generic rewording to "a package manager" should happen. Open to other suggestions on the specific that we list.

PR #4325, opened 2024-07-29
:construction: Updates List of boolean attributes to follow spec

ekwoka:
This PR simply updates the list of Boolean Attributes to follow updates in the spec, as well as moves the list out of the function and into a set for improved performance (and sorting my alphabet to make updates easier).

The actual attribute changes are

```diff
- allowpaymentrequest
+ inert
+ shadowrootclonable
+ shadowrootdelegatesfocus
+ shadowrootserializable
```

`allowpaymentrequest` was deprecated a long time ago, so I vote for removal.

Not included is `credentialess` which is a draft attribute, but is supported by everything except FireFox.


calebporzio:
great, thanks!

PR #4326, opened 2024-07-29
Adds `x-data` to `x-mask` documentation

tjbarber:
Spent more time than I'd like to admit trying to add `x-mask` to a field just to have it do nothing. It wasn't until I found [this discussion](https://github.com/alpinejs/alpine/discussions/2920) that I realized that I was missing `x-data` which wasn't mentioned in the docs for this plugin.

I think it makes sense to highlight to the user that this is required, either with something like this or an alert near the top of the page.

### DIFF
--- a/packages/docs/src/en/plugins/mask.md
+++ b/packages/docs/src/en/plugins/mask.md
@@ -69,7 +69,7 @@ The primary API for using this plugin is the `x-mask` directive.
 Let's start by looking at the following simple example of a date field:
 
 ```alpine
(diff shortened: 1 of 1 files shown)


ekwoka:
The way this is done introduces the possibility of the opposite mis understanding now, where the reader thinks that x-data is always required.

Perhaps following how this pr https://github.com/alpinejs/alpine/pull/4307 handles it is acceptable?

tjbarber:
I would say that would be acceptable if your alert was also on `x-mask` documentation!

PR #4328, opened 2024-08-02
:bug: Cleans teleported clone

ekwoka:
As a follow-up to #4300 

When `x-teleport` was nested inside `x-if`, the use of `mutateDom` was preventing x-teleport clones from being cleaned up.

This just applies the same fix to immediately clean up the `x-teleport` clone.

Thanks @mgschoen for finding this regression.


calebporzio:
Thanks!

PR #4350, opened 2024-08-23
Added `.viewport` modifier to `x-on` directive to add listeners to `window.visualViewport` instead

signorekai:
Problem
---

On iOS Safari (v17.3), `resize` events on `window` does not fire immediately to viewport resizes (caused by browser chrome hiding address bars). However, it does fire almost immediately when binded to `window.visualViewport`.

Solution
---
Added `.viewport` modifier to `x-on` directive to add listeners to `window.visualViewport`

### DIFF
--- a/packages/alpinejs/src/utils/on.js
+++ b/packages/alpinejs/src/utils/on.js
@@ -18,6 +18,7 @@ export default function on (el, event, modifiers, callback) {
     if (modifiers.includes('capture')) options.capture = true
     if (modifiers.includes('window')) listenerTarget = window
     if (modifiers.includes('document')) listenerTarget = document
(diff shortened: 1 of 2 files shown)


SimoTod:
Does the resize plugin (x-resize) have the same issue?

signorekai:
Yes it does.

Screenshot 2024-08-23 at 4 56 38 PM

As the user scrolls and the browser starts to hide the address bar, the `resize` event for `window.visualViewport` gets called almost immediately while the `x-resize` only gets called at the end of the scroll (after easing).

ekwoka:
Interesting.

Could this be an issue where the listener should be passive or you risk causing issues?

PR #4361, opened 2024-09-03
:zap: Improves x-for performance

ekwoka:
This also solves #4192 #4157 related to using `x-sort` with `x-for`.

The core issue is that x-for never checks if any elements have moved, due to how it is optimized.

I looked at just solving that directly, but it would definitely just increase work and increase code.

I identified that, if we want to just always ensure we put the elements in the right place, then we could also massively simplify the logic while maintaining the vast majority of the logic and optimizations, and solve this problem.

Looks like my test for the x-sort issue is missing so I'll get that in place.

Note: As I worked through this, I was quite impressed with how optimized Alpine actually was in the first place. It wasn't a joke when you said it was complexity with a purpose. I was quite worried I'd end up in a situation where the code logic was uglier, longer, slower, and there were times when the old version was doing way better at things I wasn't sure how to handle effectively. Had to do a lot of profiling to find the simplest fixes. 

I think I got there. Smaller, faster, and I think many would agree clearer logic.

### Synthetic Benchmarks

Some synthetic benchmarks that isolated just the x-for logic. This number includes the browser repaint/reflow, not purely logical gains. For most, the repaint/reflow is the same/similar for both.

| Action | Refactored | Diff from Original |
| --- | --- | --- |
| Add 10,000 Items | **124.235** | -20 (-14%)</span> |
| Add Additional 1 Item | **32.845** | -27 (-45%)</span> |
| Shuffle 10,001 Items | **165.770** | -68 (-29%)</span> |
| Remove 1 Item | **54.895** | -5 (-8%)</span> |
| Swapping 2 Items | **61.250** | 14 (30%)</span> |
| Sort 10,000 Items | **142.555** | -71 (-33%)</span> |
| Reverse 10,000 Items | **146.460** | -306 (-68%)</span> |
| Remove Every Other Item | **42.130** | -9 (-17%)</span> |
| Remove Remaining 5,000 Items | **16.530** | -7 (-31%)</span> |

### JS Framework Benchmarks

|Name|With Refactor|Before Refactor|
|---|---|---|
|create rows|140.1|144.3|
|replace all rows|154.5|165.6|
|updating every 10th row|29.5|29.9|
|select row|51.5|56.4|
|swap rows|42.1|45.1|
|remove row|32.1|33.7|
|create many rows|1,169.8|1,236.7|
|append rows to large table|146.4|157.7|
|clear rows|69.1|62.9|
|weighted mean|2.45|2.54|

### Extra Huge Lists

For fun.

| Action | Original | Refactor | Difference (%) |
| --- | --- | --- | --- |
| Adding 10,000 + 10,000 Items | **258** | **157** | -39.15% |
| Add 100,000 Items | **2,784** | **1,491** | -46.44% |
| Rerender Same Items | **2,863** | **125** | -95.63% |
| Shuffling 100,000 Items | **8,163** | **1,839** | -77.47% |
| Removing Every Other Item | **1,855** | **338** | -81.78% |
| Removing 100,000 Items | **827** | **115** | -86.09% |

### DIFF
--- a/packages/alpinejs/src/directives/x-for.js
+++ b/packages/alpinejs/src/directives/x-for.js
@@ -11,226 +11,152 @@ directive('for', (el, { expression }, { effect, cleanup }) => {
     let iteratorNames = parseForExpression(expression)
 
     let evaluateItems = evaluateLater(el, iteratorNames.items)
-    let evaluateKey = evaluateLater(el,
+    let evaluateKey = evaluateLater(
+        el,
         // the x-bind:key expression is stored for our use instead of evaluated.
         el._x_keyExpression || 'index'
     )
 
-    el._x_prevKeys = []
-    el._x_lookup = {}
+    el._x_lookup = new Map()
 
     effect(() => loop(el, iteratorNames, evaluateItems, evaluateKey))
 
     cleanup(() => {
-        Object.values(el._x_lookup).forEach(el =>
+        Object.values(el._x_lookup).forEach((el) =>
             mutateDom(() => {
                 destroyTree(el)
 
                 el.remove()
-            }
-        ))
+            })
+        )
 
         delete el._x_prevKeys
         delete el._x_lookup
     })
 })
 
-let shouldFastRender = true
 
-function loop(el, iteratorNames, evaluateItems, evaluateKey) {
-    let isObject = i => typeof i === 'object' && ! Array.isArray(i)
-    let templateEl = el
+const makeRefresher = (scope) => (newScope) => {
+    Object.entries(newScope).forEach(([key, value]) => {
+        scope[key] = value
+    })
+}
 
+function loop(templateEl, iteratorNames, evaluateItems, evaluateKey) {
     evaluateItems(items => {
         // Prepare yourself. There's a lot going on here. Take heart,
         // every bit of complexity in this function was added for
         // the purpose of making Alpine fast with large datas.
 
-        // Support number literals. Ex: x-for="i in 100"
-        if (isNumeric(items) && items >= 0) {
-            items = Array.from(Array(items).keys(), i => i + 1)
-        }
+        // Support number literals. Ex: x-for='i in 100'
+        if (isNumeric(items))
+            items = Array.from({ length: items }, (_, i) => i + 1)
 
         if (items === undefined) items = []
 
-        let lookup = el._x_lookup
-        let prevKeys = el._x_prevKeys
-        let scopes = []
-        let keys = []
+        const oldLookup = templateEl._x_lookup
+        const lookup = new Map()
+        templateEl._x_lookup = lookup
+        const scopeEntries = []
 
-        // In order to preserve DOM elements (move instead of replace)
-        // we need to generate all the keys for every iteration up
-        // front. These will be our source of truth for diffing.
         if (isObject(items)) {
-            items = Object.entries(items).map(([key, value]) => {
-                let scope = getIterationScopeVariables(iteratorNames, value, key, items)
+            // In order to preserve DOM elements (move instead of replace)
+            // we need to generate all the keys for every iteration up
+            // front. These will be our source of truth for diffing.
+            Object.entries(items).forEach(([prop, value]) => {
+                const scope = getIterationScopeVariables(
+                    iteratorNames,
+                    value,
+                    prop,
+                    items
+                )
 
-                evaluateKey(value => {
-                    if (keys.includes(value)) warn('Duplicate key on x-for', el)
-
-                    keys.push(value)
-                }, { scope: { index: key, ...scope} })
-
-                scopes.push(scope)
+                evaluateKey(
+                    key => {
+                        if (oldLookup.has(key)) {
+                            lookup.set(key, oldLookup.get(key))
+                            oldLookup.delete(key)
+                        }
+                        scopeEntries.push([key, scope])
+                    },
+                    {
+                        scope: { index: prop, ...scope },
+                    }
+                )
             })
         } else {
-            for (let i = 0; i < items.length; i++) {
-                let scope = getIterationScopeVariables(iteratorNames, items[i], i, items)
+            items.forEach((item, index) => {
+                const scope = getIterationScopeVariables(
+                    iteratorNames,
+                    item,
+                    index,
+                    items
+                )
 
-                evaluateKey(value => {
-                    if (keys.includes(value)) warn('Duplicate key on x-for', el)
+                evaluateKey(
+                    key => {
+                        if (typeof key === 'object')
+                            warn(
+                                'x-for key cannot be an object, it must be a string or an integer',
+                                templateEl
+                            )
 
-                    keys.push(value)
-                }, { scope: { index: i, ...scope} })
-
-                scopes.push(scope)
-            }
+                        if (oldLookup.has(key)) {
+                            lookup.set(key, oldLookup.get(key))
+                            oldLookup.delete(key)
+                        }
+                        scopeEntries.push([key, scope])
+                    },
+                    {
+                        scope: { index, ...scope },
+                    }
+                )
+            })
         }
 
-        // Rather than making DOM manipulations inside one large loop, we'll
-        // instead track which mutations need to be made in the following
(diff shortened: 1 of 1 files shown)


ekwoka:
Sure. I think there's like 2 or 3 that just got caught up in the mix.

packages/alpinejs/src/directives/x-for.js
```diff
@@ -11,226 +11,133 @@ directive('for', (el, { expression }, { effect, cleanup }) => {
...
+         * In order to remove Elements early we need to generate the key/scope pairs
+         * up front, moving existing elements from the old lookup and to the new.
+         * This leaves only the Elements to be removed in the old lookup.
+         */
+        const oldLookup = templateEl._x_lookup
+        const lookup = new Map()
+        templateEl._x_lookup = lookup
+        const scopeEntries = []
+        const hasStringKeys = isObject(items)
+        Object.entries(items).forEach(([index, item]) => {
```
jackvanderbilt-visma, line 67: Using `Object.keys` and then doing `items[key]` is more verbose but may be faster and more memory efficient, though I am not sure.

Also the forEach could be replaced with a standard `for` loop which is faster too.

craigballinger:
Thanks for all your work on this! Any estimate when it might be merged in?

simonkincaidkintronix:
I think this will fix a problem that has stumped me for a good few hours now. Just bumping to see if we can merge it in?

brianfeister:
I'm getting hit pretty hard by this bug. Also discussed here https://github.com/alpinejs/alpine/discussions/4192

Can we merge this PR please @jackvanderbilt-visma?

brianfeister:
Also, btw really great work @ekwoka in terms of being so thorough with the benchmarking. Agreed that `for` loops as an optimization aren't generally relevant in modern JS runtimes

brianfeister (approved):
LGTM

brianfeister:
@calebporzio can you or someone on core team review / merge this? The sortable problem described here is a really tragic oversight in terms of merging that plugin without `x-for` support. Also discussed here with super basic `x-sort` implementation bug: #4192

brianfeister:
@ekwoka if you can find someone to merge this, I know it's not anything massive to retire on, but I'll pay you and that person $100 via github contributions

brianfeister:
(I worked around this via https://github.com/alpinejs/alpine/discussions/4157#discussioncomment-11773743)

Senitram666:
I hope this gets merged soon.. really needing some of this fixes here..

FutsalShuffle:
merge this please

BrandonSurowiec:
I just tried Alpine's sort plugin with `x-for` and been running into so many issues with DOM updating only, and not the internal variable deriving the items. Would love to have this fix!

SkyDeckAGoGo:
I am currently working on an integration with x-sort that would benefit from one of the fixes implemented here. Would love to see this merged.

Jipok:
bump

duemti:
What's up, when will this get resolved??

SlothLady:
Any update on when this will be merged?

Using `:key="Math.floor( Math.random() * 999999 )"` atm from https://github.com/alpinejs/alpine/discussions/4157#discussioncomment-13537722 to get around the x-for x-sort interaction issue which is not ideal

JuroOravec:
@ekwoka Do you still have the code for running the benchmarks?

Last night I was looking into performance hotspots (`walk()`, `flushHandlers()`), and it looked like I got 10-25% improvement of `Alpine.start()`. But today morning I'm unable to reproduce it 😄

calebporzio:
## PR Review: #4361 — Improves x-for performance

**Type**: Bug fix + Refactor
**Verdict**: Merge

1. When `x-sort` (or any external code) physically moves DOM elements that were rendered by `x-for`, Alpine's `x-for` doesn't notice. It tracks element positions via a `prevKeys` array that records where things *should* be — but never checks where they *actually are* in the DOM.

2. So when `x-sort` moves element B before element A in the DOM, and then the reactive data updates to match that new order, `x-for`'s diff says "nothing changed" (because `prevKeys` already matches `keys`) — so it doesn't move anything. The DOM stays wrong.

3. This PR rewrites the `x-for` loop to use a simpler, more direct approach: instead of computing adds/moves/removes arrays and then applying them in separate batches, it does a single linear pass. For each key in order, it either finds the existing element in a `Map` lookup and moves it into position (if needed), or creates a new one. Elements not in the new key list get removed up front.

4. The key insight is that by always ensuring elements are in the right position (via `prev.nextElementSibling !== el` check), externally-moved elements get corrected automatically. No separate "did things move?" tracking needed.

1. **Patch the existing diffing algo to check actual DOM positions** — could compare `prevKeys` against actual DOM order before diffing. Would preserve the existing architecture but adds complexity to an already-complex algorithm. The PR's approach is cleaner because it eliminates the need for the tracking entirely.

2. **Reset `prevKeys` on every loop iteration** — force a full re-diff every time. Simple but would lose the performance benefits of incremental diffing. The PR's approach is better because it *is* effectively a full positioning pass, but one that's cheaper due to the `Map` lookup and `nextElementSibling` short-circuit.

3. **Only fix the x-sort interaction** by detecting external DOM mutations — would be surgical but fragile and would miss other cases where external code moves elements.

The PR's approach is the right one: simpler code, faster execution, and the bug fix falls out naturally.

### Changes Made

I pushed a commit fixing several issues:

1. **Bug fix: `Object.values()` on a `Map`** — The cleanup function (line 25) used `Object.values(el._x_lookup)` but `_x_lookup` was changed from `{}` to `new Map()`. `Object.values()` on a Map returns `[]`, so cleanup would silently fail to destroy child trees. Fixed to use `el._x_lookup.forEach()`.

2. **Dead code removal** — `delete el._x_prevKeys` in cleanup referenced a property that no longer exists in the new code.

3. **Style fixes** — Changed all `const` to `let` (Alpine convention), removed semicolons, restored `! ` spacing convention, converted `const makeRefresher = scope => ...` arrow to `function refreshScope(scope)` for consistency with the rest of the file.

4. **Restored original formatting** in `parseForExpression`, `getIterationScopeVariables`, `isNumeric` to match main branch exactly (the PR had gratuitous whitespace/style changes to these untouched functions).

### Test Results

- **x-for spec**: 32/32 passing (including the new "handles moved elements correctly" test)
- **morph spec**: 34/34 passing (no regressions)
- **Regression verified**: Stashed the fix, rebuilt, confirmed the new test fails on main with the original code (`expected '2' but got '1'`). Restored the fix, test passes.
- **CI**: No CI checks ran on this branch (contributor fork).

### Code Review

**The good:**
- The new algorithm is genuinely simpler: ~90 lines of loop logic vs ~130 in the original, with fewer intermediate data structures (`adds`, `moves`, `removes`, `sames`, `prevKeys` all eliminated).
- The `Map` swap from plain object `{}` is a legitimate improvement — `Map.has()`/`Map.get()` is faster than `in`/bracket access for this use case.
- The `replaceWith` trick for element swapping (line 96-98) is clever: when moving element A into position, if there's already something at that position, swap it to where A was. This avoids the marker-div dance from the original.
- Performance benchmarks are thorough and show meaningful improvements across the board.
- The new test correctly isolates the root cause: external DOM manipulation followed by reactive data sync.

**Minor concerns (not blocking):**
- The `replaceWith` swap at line 96-98 is a novel pattern. It works because every element in the lookup will be visited in order, so displaced elements get corrected on their own iteration. But it's worth understanding that this depends on visiting every key — if any key were skipped, an element could end up orphaned. This is fine given the current code structure.
- `Object.entries(items)` for arrays gives string indices that need `parseInt()` — slightly wasteful compared to a simple `for` loop, but the readability gain is worth it for Alpine.

### Security

No security concerns identified. No changes touch expression evaluation, `x-html`, or user-provided input handling.

### Verdict

**Merge.** This is a well-executed refactor that fixes a real bug (x-for ignoring externally-moved DOM elements, breaking x-sort integration), simplifies the code, and improves performance. 14 thumbs-up reactions, multiple community members blocked by the bug, one external approval. The cleanup bug I found and fixed was the only real issue — the core algorithm is sound. Community has been waiting 16+ months for this.

---
*Reviewed by Claude*

PR #4363, opened 2024-09-05
Remove unnecessary parentheses from bind.md example

danielsbird:
When I saw this example using `x-data="dropdown()"` and [other examples](https://alpinejs.dev/globals/alpine-data#encapsulating-directives-with-x-bind) using ``x-data="dropdown"`` I wondered what caused this example to require parentheses. It seems like adding parentheses is redundant since the example code works both with and without them.

### DIFF
--- a/packages/docs/src/en/directives/bind.md
+++ b/packages/docs/src/en/directives/bind.md
@@ -162,7 +162,7 @@ And like most expressions in Alpine, you can always use the result of a JavaScri
 The object keys can be anything you would normally write as an attribute name in Alpine. This includes Alpine directives and modifiers, but also plain HTML attributes. The object values are either plain strings, or in the case of dynamic Alpine directives, callbacks to be evaluated by Alpine.
 
 ```alpine
-<div x-data="dropdown()">
+<div x-data="dropdown">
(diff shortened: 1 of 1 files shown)


browner12:
I think it's fine as is, and probably more accurate. `dropdown()` is a function you are executing to return an object.

danielsbird:
@browner12 I agree with your reasoning. 

How do you think Alpine's docs can help beginners like me understand that there's no difference between `x-data="dropdown()"` and `x-data="dropdown"`? Should all examples use `x-data="dropdown()"` instead of `x-data="dropdown"`?

browner12:
is kind of less of an Alpine question, and more of a JS question. Alpine *does* do some magic that allows you to [omit the parentheses](https://alpinejs.dev/directives/data#methods), and you're kind of not *technically* writing Javascript in the attribute, but IMO I treat it as valid Javascript, which is why I always use the parentheses.

consistency would be nice, and I would personally lean towards using the parentheses, but that's something a maintainer will have to decide.

ekwoka:
To clarify on the difference, if the function is nested in an object, and accesses `this`, the `this` will be different between with and without the parentheses.

PR #4371, opened 2024-09-18
Added missing classes to anchor docs

PluTiper:
Noticed that the last example in the anchor plugin docs doesn't have the same style as the previous ones.

### DIFF
--- a/packages/docs/src/en/plugins/anchor.md
+++ b/packages/docs/src/en/plugins/anchor.md
@@ -205,7 +205,7 @@ Because `x-anchor` accepts a reference to any DOM element, you can use utilities
     </div>
 
 
-    <div x-show="open" x-anchor="document.querySelector('.trigger')">
+    <div x-show="open" x-anchor="document.querySelector('.trigger')" class="bg-white rounded p-4 border shadow z-10">
         Dropdown content
     </div>
 </div>


PluTiper:
I was experimenting with the plugin and also noticed that there might be an issue with the previous position caching. The variable gets re-declared each time the compute function is executed which unnecessarily updates the alpine variables.

calebporzio:
## PR Review: #4371 — Added missing classes to anchor docs and fixed anchor position caching

**Type**: Bug fix + Docs
**Verdict**: Close

This PR has two changes:

1. **Docs fix**: The last demo in the anchor plugin docs (the "Anchor to an ID" section) was missing the styling classes (`bg-white rounded p-4 border shadow z-10`) that all the other demos have. This made it look unstyled when rendered on the docs site.

2. **Position caching fix**: The `x-anchor` directive has a performance optimization that's supposed to prevent unnecessary Alpine reactivity updates when the anchor position hasn't changed. It works by comparing the current `{x, y}` to a `previousValue`. But `previousValue` was declared *inside* the `compute()` function:

```js
let compute = () => {
    let previousValue  // <-- resets to undefined every call
    computePosition(...).then(({ x, y }) => {
        if (JSON.stringify({ x, y }) !== previousValue) { // always true
            ...
        }
        previousValue = JSON.stringify({ x, y }) // stored but lost next call
    })
}
```

So every time `autoUpdate` fires (on scroll, resize, etc.), `previousValue` starts as `undefined` and the comparison always passes. The reactive properties get set even when nothing changed, triggering unnecessary Alpine reactivity cascades. The PR moves `let previousValue` outside `compute()` so it persists across calls.

1. **Move `previousValue` outside `compute()`** — This is what the PR does. It's the correct minimal fix.
2. **Compare `x`/`y` directly instead of JSON.stringify** — Could compare `el._x_anchor.x !== x || el._x_anchor.y !== y` instead. Avoids JSON serialization overhead. Slightly cleaner, but the current approach works fine for this use case.
3. **Do nothing** — The bug is real but functionally benign. Positions are always correct; you just get extra reactive updates. However, on pages with many anchored elements, unnecessary reactive churn on every scroll event could add up.

### Changes Made

No changes pushed — this PR cannot be merged as-is (see below).

### Test Results

- Existing `anchor.spec.js`: 1/1 passing
- No new tests added by the PR
- CI: No checks reported

### Code Review

**The core problem: this PR is stale and conflicts with main.**

PR #4735 (`feat(x-anchor): allow dynamic reference to be used with x-anchor`) was recently merged and **substantially restructured** the anchor directive — wrapping the compute logic inside an `effect()` callback with reference tracking. This PR's diff conflicts with that new code.

However, **the `previousValue` bug still exists in the new code from #4735** — `let previousValue` is still declared inside `compute()` in the refactored version. So the underlying issue this PR identified is legitimate and still present.

The docs fix (missing classes) is also still needed — it was not addressed by #4735.

### Security

### Verdict

**Close this PR.** The contributor correctly identified a real bug, but the PR is now stale (opened Sep 2024) and conflicts with the recently merged #4735 refactor. The fix cannot be cleanly applied to the current codebase.

However, both issues should still be addressed:
- The `previousValue` scoping bug persists in the new code from #4735
- The docs styling fix is still needed

I'd recommend applying both fixes directly on main since the changes are trivial — just move `let previousValue` outside `compute()` in the new code structure, and add the missing classes to the docs. Thank you @PluTiper for catching this!

---
*Reviewed by Claude*

PR #4382, opened 2024-09-29
Add `x-anchor.noflip` modifier

aguingand:
This PR adds ability to disable auto flipping for anchor. In some cases you want to force the anchored element to stay at his position and not flipping to the top if there is no enough room below. For example with large menus or popovers teleported to body. 

Usage : 
```html
<div x-anchor.noflip>
</div>

<!-- with position -->
<div x-anchor.bottom-start.noflip>
</div>
```


ekwoka:
Is there a way to add a test for this?

PR #4406, opened 2024-10-19
Fix getting the old value in deep watching

Suall1969:
This addresses the issue with old value in [deep watching](https://alpinejs.dev/magics/watch#deep-watching) as described in #3293. In contrast to #3046 and #3617, this PR relies exclusively on the [strcuturedClone](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone) API and falls back to the current behaviour if not available, drastically simplifying things.


ekwoka (approved):
looks good! But could you add tests for it?

For reference, `structuredClone` is Baseline (as of 2022).

PR #4419, opened 2024-10-28
Handle all types of white space in class lists, not just spaces

mattvague:
Fixes bug [mentioned here](https://github.com/alpinejs/alpine/discussions/4418) where using a newline in e.g. x-transition:enter would throw an error

### DIFF
--- a/tests/cypress/integration/directives/x-bind-class.spec.js
+++ b/tests/cypress/integration/directives/x-bind-class.spec.js
@@ -82,13 +82,13 @@ test('classes are removed before being added',
     }
 )
 
-test('extra whitespace in class binding string syntax is ignored',
+test.only('extra whitespace in class binding string syntax is ignored',
     html`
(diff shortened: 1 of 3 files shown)


mattvague:
@ekwoka Just FYI while working on this I was finding Cypress painfully slow to work with on my machine so I upgraded to the latest cypress version in [this other PR](https://github.com/alpinejs/alpine/pull/4420)

tests/cypress/integration/directives/x-bind-class.spec.js
```diff
@@ -82,13 +82,13 @@ test('classes are removed before being added',
     }
 )
 
-test('extra whitespace in class binding string syntax is ignored',
+test.only('extra whitespace in class binding string syntax is ignored',
```
txels: I presume you left this unintentionally.

PR #4420, opened 2024-10-28
Task/update cypress

mattvague:
Current version of Cypress is 7.0.0 which is tremendously out of date and was running unbelievably slow at least on my computer. This updates to Cypress 13.15.1 which paves the way for us to also see if we can migrate from e2e specs to component specs which (IMO) are more suited to this case.


tests/cypress/integration/directives/x-on.spec.js
```diff
@@ -770,11 +770,11 @@ test(
...
-        get("@ctrl").as("ctrl").should(beChecked());
-        get("@meta").as("meta").should(beChecked());
-        get("@alt").as("alt").should(beChecked());
-        get("@cmd").as("cmd").should(beChecked());
+        get("@shift").should(beChecked());
+        get("@ctrl").should(beChecked());
+        get("@meta").should(beChecked());
+        get("@alt").should(beChecked());
+        get("@cmd").should(beChecked());
     }
```
mattvague, line 778: Not sure why this wasn't causing an issue before (or maybe it was and it was hidden), but in the latest cypress this cypress this throws a max stack depth error (as you would expect it to I think?)

ekwoka: funny.

I think that's my mistake. I added those in one spot, and maybe copy pasted it without care and it happened to work on the old version.

calebporzio:
## PR Review: #4420 — Update to latest cypress

**Type**: Refactor (tooling/infrastructure)
**Verdict**: Close

1. Alpine's test suite uses Cypress 7.0.0, which is from 2021. The contributor wants to upgrade to Cypress 13.15.1.
2. Cypress 10+ moved from `cypress.json` to `cypress.config.js` — this PR handles that migration.
3. The new Chrome bundled with Cypress 13 serializes CSS custom properties differently: `--var:value` becomes `--var: value` (space after colon). Three style assertions in `x-bind-style.spec.js` were updated to match.
4. Cypress 13 is stricter about `.as()` aliasing — using `.as("shift")` then querying `@shift` on the same element caused infinite recursion. The fix removes the `.as()` calls since the aliases weren't needed (the elements were already queried by `@shift` etc. via `x-ref`).

1. **Do the upgrade now with a complete config migration** — carry over all the old `cypress.json` settings (`screenshotOnRunFailure: false`, `video: false`, `fixturesFolder`, `screenshotsFolder`, `videosFolder`, `excludeSpecPattern: ['*.html']`). The PR's config is incomplete — it drops several settings.
2. **Skip straight to Cypress 14/15** — the PR targets 13.15.1 but Cypress is now at 15.10.0. If we're going to eat the migration cost, we should go to the latest.
3. **Don't merge, do the upgrade ourselves on main** — given the merge conflicts with `package-lock.json` and the staleness of this PR (Oct 2024, 16+ months old), it would be less work to do a fresh upgrade on main.

### Problems with this PR

**1. Stale and conflicts.** The PR is from Oct 2024. It has merge conflicts in `package-lock.json` that can't be trivially resolved. Main has diverged significantly.

**2. Incomplete cypress.config.js migration.** The new config is missing several settings from the old `cypress.json`:
- `screenshotOnRunFailure: false` — defaults to `true` in Cypress 13+, so screenshots will now be generated on failures
- `video: false` — actually fine (defaults to false in 13+), but worth being explicit
- `fixturesFolder`, `screenshotsFolder`, `videosFolder` — all default to `cypress/` paths instead of `tests/cypress/` paths
- `excludeSpecPattern: ['*.html']` — the old `ignoreTestFiles` equivalent. The `specPattern: '*.spec.js'` happens to exclude `.html` files, but the broader `tests/cypress/**/*.spec.js` pattern is wider than the old `integrationFolder` path

**3. `specPattern` is too broad.** `tests/cypress/**/*.spec.js` matches specs under `support/` or any other subdirectory. The old config only looked in `tests/cypress/integration/`.

**4. Already outdated.** Targets Cypress 13.15.1, but Cypress is at 15.10.0 now. If we're going to break the lock file anyway, let's go to current.

**5. Style issues.** `cypress.config.js` uses semicolons and `const` — Alpine convention is no semicolons and `let`. Minor, but worth noting.

**6. No CI checks.** The PR has zero CI check results, so we can't verify tests pass.

### Changes Made

No changes made.

### Test Results

No tests were run — the PR has merge conflicts and would require a `package-lock.json` regeneration to be installable.

### Code Review

The test assertion changes are correct:
- `x-bind-style.spec.js`: CSS custom property serialization changed in newer Chrome (space after colon). This is a browser behavior change, not an Alpine bug.
- `x-on.spec.js:770-777`: Removing `.as()` before `.should()` is the right fix — the aliases create circular references.

The `cypress.config.js` is functional but incomplete (see problems above).

### Security

### Verdict

**Close.** The intent is good — Cypress 7 is ancient and should be upgraded. But this PR is 16 months stale with merge conflicts, targets an already-outdated Cypress version, and has an incomplete config migration. The cleanest path is to do a fresh Cypress upgrade on main (targeting 14 or 15), write a complete `cypress.config.js`, and resolve any test adjustments at that time. Happy to do that upgrade if you want.

---
*Reviewed by Claude*

PR #4428, opened 2024-11-03
Don't init trees within ignore elements (x-ignore)

calebporzio:


lonnylot:
@calebporzio Can we get this tagged / pulled in to livewire? 🙇

angus-mcritchie:
@calebporzio This seems to have caused breaking changes in [Filament](https://github.com/filamentphp/filament), I attempted to make a [pull request](https://github.com/filamentphp/filament/pull/15044#event-15597711151) and fix it there, but there is more code relying on this behaviour so it was declined.

As it's a breaking change from Alpine JS, maybe it should be fixed in Alpine as [pointed out here](https://github.com/filamentphp/filament/pull/15044#issuecomment-2531015022)?

PS. big fan of your work, Livewire & Flux Pro subscriber ❤️

SimoTod:
@angus-mcritchie the fix is in code review with async alpine v2 (not sure if filament is already using it though, i think they are still on v1).
This is not technically a breaking change but it was a bug for previous versions of alpine, a directive that has a parent element with x-ignore should not initialise.

I can see if I can fix it for async alpine v1 as well

angus-mcritchie:
@SimoTod Filament is using Alpine JS 3 as a dependency from Livewire.

In Filament v3.2.125, Livewire had to be locked to v3.5.12 because of [3.14.5](https://github.com/livewire/livewire/pull/9050) breaking change/bug fix.

Although I can see that this is meant to be a bug fix, this change should be reverted as it has caused a terrible breaking change (making the whole app unusable).

SimoTod:
@angus-mcritchie I can see your point, let's see what the maintainer decides.

In theory filament should get rid of those x-ignores but I think they have limited time/resources for implementing it (they also locked async alpine because they already had issues with those x-ignores with the latest v1 release).

danharrin:
I'm going to stop using Async Alpine for this table component in Filament so that the nested components inside of it don't break. We can't upgrade Async Alpine to any new versions because there are breaking changes which we haven't had time to investigate. We may attempt an upgrade to v2 in the next major version of Filament but this is currently low priority as it does not really deliver benefits for our users.

PR #4430, opened 2024-11-07
Clearer example for x-bind

sa-:
Hey, first of all thanks for making alpine - it's a breath of fresh air to use in the modern web stack for someone who is not a "frontend guy"

While I was reading through the docs I was quite confused because `placeholder` was both a variable name and an attribute for the input field. It took me unreasonably long to figure it out and I believe changing the variable name slightly would have helped me grok it faster.

Cheers

### DIFF
--- a/packages/docs/src/en/directives/bind.md
+++ b/packages/docs/src/en/directives/bind.md
@@ -10,8 +10,8 @@ title: bind
 For example, here's a component where we will use `x-bind` to set the placeholder value of an input.
 
 ```alpine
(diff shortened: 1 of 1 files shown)


sa-:
I don't think my changes are related to the failing test `does not initialise components twice when contained in multiple mutations`

ekwoka:
The failed tests are related to some things Caleb is working on, not yours.

ekwoka (approved):
Makes sense! Just remove the ambiguity

PR #4441, opened 2024-11-22
Modify the CSP Evaluator to handle negated expressions

cadolphs:
This commit adds support to use straightforward negated expressions in Alpine directives in the CSP build.

While we (probably) do not want to rewrite a full CSP-save expression evaluator to handle expressions of arbitrary complexity, directives with negation are such a common use case that it makes sense to support them.

Consider, for example,

* `x-show="!loading"`
* `:disabled="!valid"`

In the CSP, we would have to write custom getters for each boolean attribute whose negation we might want to use, adding needless verbosity.

On the other hand, being a unary operator, hand-rolling negation is easy enough.

To test that it all works, we add relevant cypress tests to the csp spec to verify that we can negate both a plain attribute as well as the result of a function call.

### DIFF
--- a/packages/csp/src/evaluator.js
+++ b/packages/csp/src/evaluator.js
@@ -25,10 +25,13 @@ function generateDataStack(el) {
 }
 
 function generateEvaluator(el, expression, dataStack) {
-    return (receiver = () => {}, { scope = {}, params = [] } = {}) => {
+    return (receiver = () => { }, { scope = {}, params = [] } = {}) => {
         let completeScope = mergeProxies([scope, ...dataStack])
 
-        let evaluatedExpression = expression.split('.').reduce(
+        let isNegated = expression.trim().charAt(0) === '!'
+        let expressionToEvaluate = isNegated ? expression.trim().slice(1) : expression.trim()
+
+        let evaluatedExpression = expressionToEvaluate.split('.').reduce(
             (currentScope, currentExpression) => {
                 if (currentScope[currentExpression] === undefined) {
                     throwExpressionError(el, expression)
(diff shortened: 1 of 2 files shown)


ekwoka:
The part for "negation of a function call" I think is problematic. As this would now introduce unique behavior to the CSP build that doesn't work in Alpine standard, or in javascript.

🤔 

It introduces a bit of unnecessary magic I think...

packages/csp/src/evaluator.js
```diff
@@ -25,10 +25,13 @@ function generateDataStack(el) {
 }
 
 function generateEvaluator(el, expression, dataStack) {
-    return (receiver = () => {}, { scope = {}, params = [] } = {}) => {
+    return (receiver = () => { }, { scope = {}, params = [] } = {}) => {
         let completeScope = mergeProxies([scope, ...dataStack])
 
-        let evaluatedExpression = expression.split('.').reduce(
+        let isNegated = expression.trim().charAt(0) === '!'
+        let expressionToEvaluate = isNegated ? expression.trim().slice(1) : expression.trim()
```
levu42, line 32: shouldn't it be `...expression.trim().slice(1).trim()`?

SimoTod:
I think you can apply the negation to evaluatedExpression before passing it into runIfTypeOfFunction to preserve the same behaviour as the non-CSP built

cadolphs:
> I think you can apply the negation to evaluatedExpression before passing it into runIfTypeOfFunction to preserve the same behaviour as the non-CSP built

That's what I did initially, and then I started overthinking it, assuming that I needed to support the function-calling thing.

cadolphs:
Thanks for your comments, everyone. I made the changes as recommended and pushed a new commit.

ekwoka:
Yeah, I could see the motivation for it, and it could make sense in a strict "built from nothing" situation.

But in this case, divergent behaviors would be an issue.

cadolphs:
@ekwoka I've taken out the extra function evaluation magic; wondering if anyone wants to take a second look.

cadolphs:
@calebporzio Hi Caleb, thought I'd ping you here. I love Alpine and have pushed hard for it to be used in a commercial project, but due to CSP requirements we have to use the CSP build, and I just wanted to add this small ergonomic enhancement so that in addition to "plain" attributes, we can also use negations.

It's now been several months with no additional input on where this should go, so I was hoping to get your eyeballs on this.

Cheers!

cadolphs:
@ekwoka I see you being quite active in the discussions around here, so was hoping to get your thoughts on the whole CSP build side of things; am I missing something about how I should go about moving this PR forward?

ekwoka (approved):
Looks good to me. Simple and clean

cadolphs:
Thanks; now what can we do to get it into a release? :D

tanczosm:
It's definitely frustrating to have to write custom negation methods for everything.  (notIsOpen)

It also seems more like a hacky workaround.  I'm looking forward to seeing an update with this alteration.

PR #4450, opened 2024-11-28
:bug: Fixes Regression in Mutation handling

ekwoka:
Fixes #4449 

Due to the location of the check for "has this element already been initialized", elements that were moved within an alpine component, without a new wrapper element, were being improperly marked as only having been removed, and we summarily cleaned up and not reinitialized.

this just moves that check to just before the element would be reinitialized (Should this check maybe just be part of the lifecycle instead of mutation entirely?)

Includes test for the case that fails if you revert the changes to the mutation code.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -63,6 +63,10 @@
             "resolved": "packages/persist",
             "link": true
         },
+        "node_modules/@alpinejs/resize": {
+            "resolved": "packages/resize",
+            "link": true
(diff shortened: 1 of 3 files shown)


calebporzio:
Legend, thank you

syrtsev-aleksandr:
Spended entire day to fix bug in my app related to this problem. Absolute legend

PR #4451, opened 2024-11-28
Optimize mutation observer to better handle move operations - ref #4450

calebporzio:


packages/alpinejs/src/mutation.js
```diff
@@ -112,14 +112,16 @@ export function flushAndStopDeferringMutations() {
 }
 
 function onMutate(mutations) {
+    console.log(mutations);
```
levu42, line 115: @calebporzio you forgot a console.log in the code here ;)

joshhanley: @levu42 thanks! He removed it after in a WIP commit 😆

PR #4458, opened 2024-11-29
Fix regression on x-ignore when removed

SimoTod:
Release v3.14.4 introduce an _x_marker property to flag elements that were already initialised so they don't get initialised again when moved around (I think it was causing some issues in Livewire). However the code was also marking as initialised elements that were skipped by x-ignore.

This causes issues if x-ignore is removed and people calls initTree on the element: the subtree is initialised but the parent is skipped even if it wasn't previously initialised so, if it has x-data, it will trigger errors.

See https://github.com/alpinejs/alpine/discussions/4457

This PR changes the logic so _x_marker is not added when the element has x-ignore

### DIFF
--- a/packages/alpinejs/src/lifecycle.js
+++ b/packages/alpinejs/src/lifecycle.js
@@ -93,16 +93,16 @@ export function initTree(el, walker = walk, intercept = () => {}) {
             // If the element has a marker, it's already been initialized...
             if (el._x_marker) return
 
-            // Add a marker to the element so we can tell if it's been initialized...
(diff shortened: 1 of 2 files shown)


ekwoka (approved):
Quite sensible

SimoTod:
@s-kitade that code doesn't use x-ignore so no, it won't. 
However, the last version is 3.14.5 which resolved another bug so it may be fixed already

calebporzio:
Legend, thanks @SimoTod

PR #4469, opened 2024-12-06
Fix x-teleport by removing clone x-ignore

joshhanley:
As per the discussion here https://github.com/alpinejs/alpine/discussions/4467#discussioncomment-11461276 I've removed the Teleport clone x-ignore.

I've run the tests locally and they pass. I've also made a build of Livewire with it, run it's tests and all pass.

Fixes #4461, fixes #4467, fixes livewire/livewire#9067


calebporzio:
Thanks Josh!

PR #4483, opened 2024-12-22
Add support for [x-sort:handle] attributes defined in template tags

SimoTod:
Closes #4482

Many developer uses x-sort with x-for. In this cases, when x-sort gets processed there's technically no `[x-sort:handle]` in any of the children even when the x-for template would create element with said attribute.

The MR adds some additional logic (in the real word there will probably just be 1 single tag as a direct children but it takes a more generic approach) to support [x-sort:handle] defined in template tags (likely x-for).


jan-dh:
Would really be happy to see this get merged

PR #4496, opened 2025-01-03
Fix focus documentation example code

iamshcc:
Fix the issue in the x-trap.noscroll example code in documentation.


joshhanley:
@iamshcc thanks!

PR #4265, opened 2024-06-14
:bug: Fixes issue with setters accessing deeply nested data

ekwoka:
Solves an issue presented in Discord

Apparently there is an [upstream bug](https://github.com/vuejs/core/blob/31abdc8adad569d83b476c340e678c4daa901545/packages/reactivity/src/baseHandlers.ts#L148) in @vue/reactivity that, when using a setter to access deeply nested data on a part of `this` that was not in the same object, would error out. As vue was not respecting the `this` context when using reflection.

I intend to make a PR there to address this as well, but Alpine is broken by even updating one minor version, so this still needs to be addressed here.

The issue can be demonstrated here:

[3.13.0](https://awesomealpine.com/play/?coreplugins=0&html=PGRpdgogIHgtZGF0YT0ieyBmb286ICB7IGJhcjogJ2ZpenpidXp6JyB9LCBlcnJvcjogJycgfSIKICBjbGFzcz0icC0yIGZsZXggZmxleC1jb2wgZ2FwLTIgaXRlbXMtY2VudGVyIgogIEBlcnJvci53aW5kb3c9ImVycm9yID0gJGV2ZW50LmVycm9yIgo-CiAgPGRpdgogICAgeC1kYXRhPSJ7IGdldCBiYXIoKSB7IHJldHVybiB0aGlzLmZvby5iYXIgfSwgc2V0IGJhcih2YWx1ZSkgeyB0aGlzLmZvby5iYXIgPSB2YWx1ZSB9IH0iCiAgPgogICAgPGJ1dHRvbgogICAgICBjbGFzcz0icC0yIGJnLWdyYXktMjAwIHJvdW5kZWQiCiAgICAgIGlkPSJvbmUiCiAgICAgIHgtdGV4dD0iYmFyIgogICAgICBAY2xpY2s9ImJhciA9ICdmb29iYXInIgogICAgPjwvYnV0dG9uPgogIDwvZGl2PgogIDxzcGFuCiAgICBpZD0idHdvIgogICAgeC10ZXh0PSJmb28uYmFyIgogID48L3NwYW4-CiAgPHNwYW4KICAgIHgtdGV4dD0iYEVycm9yOiAke2Vycm9yfWAiCiAgICBjbGFzcz0idGV4dC1yZWQtNjAwIgogID48L3NwYW4-CjwvZGl2Pgo&script=Y29uc29sZS5sb2coJ2hlbGxvJyk7Cg&tw=1&v=3.13.0&panel=config)

[...]

PR #4528, opened 2025-02-12
Fix class setters invocation on nested `x-data`

helio3197:
The current implementation of the `set` trap in `mergeProxies` restricts key lookup to the target's own property keys. This prevents class setters defined in the target's prototype from being called when there is more than one object in the `x-data` stack.

This allows components like this to work:

```html
<script>
  class BaseHandler {
    _prop = 'original'
  
    get prop() {
      return this._prop
    }
  
    set prop(v)  {
      this._prop = v
    }
  }
  window.baseHandler = () => new BaseHandler()
</script>

<div x-data>
  <div x-data="baseHandler()">
    <button x-text="prop" @click="prop = 'edited'"></button>
  </div>
</div>
```

### DIFF
--- a/packages/alpinejs/src/scope.js
+++ b/packages/alpinejs/src/scope.js
@@ -38,10 +38,10 @@ export function mergeProxies (objects) {
 }
 
 function keyInPrototypeChain(obj, key) {
+    if (obj === null || obj === Object.prototype) return null
     if (Object.prototype.hasOwnProperty.call(obj, key)) return obj
 
-    const proto = Object.getPrototypeOf(obj)
-    return proto && keyInPrototypeChain(proto, key)
+    return keyInPrototypeChain(Object.getPrototypeOf(obj), key)
 }
 
 let mergeProxyTrap = {


ekwoka:
I'm not sure how the test demonstrates the issue.

You have code related to prototype chain but the test doesn't use the prototype chain at all.

Was the previous PR that enabled classes working not work for this test?

What exactly was the issue?

helio3197:
It does, because the setter function is defined in the class instance's prototype not in the instance object itself, then when the setter is invoked e.g. `value = 'somevalue'` the current proxy's set trap won't be able to set the property value, because there is a check asserting that the invoked key must be an own property of the target (`Object.prototype.hasOwnProperty.call(obj, name)`) which evaluates to false in this case because the property itself is a getter and setter defined in the class instance prototype.

You can check the live issue in this [codepen](https://codepen.io/helio3197/pen/dPyPMvy), and [here](https://codepen.io/helio3197/pen/XJWJdMo) is the same example with the implemented change.

ekwoka:
Okay, and it bypasses the vue reactivity bug since it's a class setter and not own setter?

helio3197:
No, I had not considered the scenario described by this [PR](https://github.com/alpinejs/alpine/pull/4265), therefore referencing other parent data-stack objects will result in error. However, I've modified the code to use the prototype, where the setter is defined as the 'target'. Thus, this check can pass, and the setter method can be called with the mergeProxy as the `this` value.
```
const descriptor = Object.getOwnPropertyDescriptor(target, name);
        if (descriptor?.set && descriptor?.get)
```

calebporzio:
## PR Review: #4528 — Fix class setters invocation on nested `x-data`

**Type**: Bug fix
**Verdict**: Merge

When you use a class as your `x-data` component and that class defines getter/setter pairs (like `get value()` / `set value(v)`), those accessors live on the **class prototype**, not on the instance itself.

Here's the sequence that triggers the bug:

1. You have nested `x-data` elements: an outer `<div x-data>` and an inner `<div x-data="handler()">` where `handler()` returns a class instance.
2. The data stack has two objects: `[classInstance, outerScopeObject]`.
3. When you write `value = 'foo'` in an Alpine expression, the proxy's `set` trap fires.
4. The `set` trap loops through the data stack looking for which object owns `value`.
5. **The bug**: it used `hasOwnProperty` to check — which only looks at own properties. Since `get value()`/`set value()` live on `BaseHandler.prototype`, not on the instance, `hasOwnProperty` returns `false`.
6. No match is found, so it falls through to the **last** object in the stack (the outer scope).
7. A plain `value` property gets created on the outer scope instead of invoking the class setter. The setter never runs.

This only breaks when there are **multiple** objects in the data stack (nested `x-data`). With a single `x-data`, the fallback `objects[objects.length - 1]` happens to be the class instance itself, so it accidentally works.

1. **Use `Reflect.has()` instead of `hasOwnProperty` in the set trap** — Would detect prototype properties but only returns a boolean, not the actual prototype object where the descriptor lives. We need the specific object to call `Object.getOwnPropertyDescriptor(target, name)` and find the setter.
2. **Use `name in obj` + separate prototype walk** — Functionally equivalent to the PR's approach but split across two steps. No simpler.
3. **The PR's approach: `keyInPrototypeChain` recursive helper** — Walks up the prototype chain and returns the object that owns the key. This is the cleanest solution because it combines "does this object have the key?" and "give me the right target for the descriptor" in one step.

The PR's approach is the right one.

### Changes Made

I pushed one fix: **guard `keyInPrototypeChain` against `Object.prototype`**. The original implementation would walk all the way up to `Object.prototype`, which means properties like `toString`, `valueOf`, `hasOwnProperty`, etc. would be found there instead of falling through to the intended fallback (`objects[objects.length - 1]`). In theory, `Reflect.set(Object.prototype, 'toString', value)` could pollute the global prototype. The fix adds an early return when reaching `Object.prototype`:

```js
// Before (from PR)
function keyInPrototypeChain(obj, key) {
    if (Object.prototype.hasOwnProperty.call(obj, key)) return obj
    const proto = Object.getPrototypeOf(obj)
    return proto && keyInPrototypeChain(proto, key)
}

// After (with guard)
function keyInPrototypeChain(obj, key) {
    if (obj === null || obj === Object.prototype) return null
    if (Object.prototype.hasOwnProperty.call(obj, key)) return obj
    return keyInPrototypeChain(Object.getPrototypeOf(obj), key)
}
```

### Test Results

- `scope.spec.js`: **9/9 passing** (7 existing + 2 new)
- Regression verified: both new tests **fail without the fix**, **pass with it**
- All 7 existing tests continue to pass (no regressions)
- CI: build passing

### Code Review

- `packages/alpinejs/src/scope.js:40-45`: The `keyInPrototypeChain` helper is clean and minimal. Recursion depth is bounded by prototype chain length (typically 2-3 levels for class instances). No performance concern.
- `packages/alpinejs/src/scope.js:75-81`: The `set` trap change is surgical — only the target lookup logic changed. The existing `descriptor?.set && descriptor?.get` check and the Vue reactivity workaround are untouched.
- `tests/cypress/integration/scope.spec.js:161-233`: Both tests are well-structured. The first tests the core bug (class setter + nested x-data). The second tests the interaction with parent scope access through `this` in the setter (covering the Vue reactivity workaround path). Naming is clear and descriptive.
- The `get` trap already uses `Reflect.has()` which walks prototypes — this fix brings the `set` trap into parity. The asymmetry was the root cause.

### Security

The `Object.prototype` guard I added prevents a theoretical prototype pollution vector. No other security concerns identified.

### Verdict

This is a legitimate bug with a clean, minimal fix. The contributor correctly identified the root cause (get/set asymmetry in prototype chain handling), provided two well-targeted tests, and handled the Vue reactivity workaround interaction. The discussion in the comments shows they iterated thoughtfully after @ekwoka's feedback. I've added one safety guard to prevent prototype pollution via `Object.prototype`. **Recommend merge.**

---
*Reviewed by Claude*

PR #4548, opened 2025-02-21
Don't assign the normalEvaluator per default

JeroenBoersma:
Assigning the normal evaluator to theEvaluatorFunction means that it always needs to be exported.

CSP doesn't use the normalEvaluator and should not include it in any build. alpinsjs/src/index.js always assigns the normalEvaluator manually and works as expected, and it isn't included anymore in the csp builds.

* normalEvaluator also needed dependencies for unsafe functions which we don't want to include in the csp build


ekwoka (approved):
I like it!

PR #4579, opened 2025-03-17
Add `ignore-nullish` modifier to `x-html` and `x-text`

willrowe:
This add an `ignore-nullish` modifier to `x-html` and `x-text`. If the value is set to `null` or `undefined`, the currently set HTML/text will remain unchanged.

The main use case for this is to have the initial HTML/text set in the markup on page load.  See the updates to the documentation for examples.

### DIFF
--- a/packages/alpinejs/src/directives/x-html.js
+++ b/packages/alpinejs/src/directives/x-html.js
@@ -2,18 +2,22 @@ import { directive } from '../directives'
 import { initTree } from '../lifecycle'
 import { mutateDom } from '../mutation'
 
-directive('html', (el, { expression }, { effect, evaluateLater }) => {
+directive('html', (el, { expression, modifiers }, { effect, evaluateLater }) => {
     let evaluate = evaluateLater(expression)
 
     effect(() => {
         evaluate(value => {
-            mutateDom(() => {
-                el.innerHTML = value
+            value = value ?? null;
 
-                el._x_ignoreSelf = true
-                initTree(el)
-                delete el._x_ignoreSelf
-            })
+            if (!modifiers.includes('ignore-nullish') || value !== null) {
+                mutateDom(() => {
+                    el.innerHTML = value
+    
+                    el._x_ignoreSelf = true
+                    initTree(el)
+                    delete el._x_ignoreSelf
+                })    
+            }
         })
     })
 })
--- a/packages/alpinejs/src/directives/x-text.js
+++ b/packages/alpinejs/src/directives/x-text.js
@@ -1,14 +1,18 @@
 import { directive } from '../directives'
 import { mutateDom } from '../mutation'
 
-directive('text', (el, { expression }, { effect, evaluateLater }) => {
+directive('text', (el, { expression, modifiers }, { effect, evaluateLater }) => {
     let evaluate = evaluateLater(expression)
 
     effect(() => {
         evaluate(value => {
-            mutateDom(() => {
-                el.textContent = value
-            })
+            value = value ?? null;
+
+            if (!modifiers.includes('ignore-nullish') || value !== null) {
+                mutateDom(() => {
+                    el.textContent = value
+                })    
+            }
         })
     })
 })
--- a/packages/docs/src/en/directives/html.md
+++ b/packages/docs/src/en/directives/html.md
@@ -27,3 +27,25 @@ Here's a basic example of using `x-html` to display a user's username.
 <!-- END_VERBATIM -->
 
 Now the `<span>` tag's inner HTML will be set to "<strong>calebporzio</strong>".
+
+<a name="modifiers"></a>
+## Modifiers
+
+<a name="ignore-nullish"></a>
+### .ignore-nullish
+
+`.ignore-nullish` will ignore the value if it resolves to `null` or `undefined` and retain the currently set HTML.
+
+```alpine
+<section x-data="{ items: null }">
+    <ul x-html.ignore-nullish="items">
+        <li>
+            <em>Not loaded yet...</em>
+        </li>
+    </ul>
+
+    <button x-on:click="items = '<li>First item</li><li>Second item</li>'">Load Items</button>
+</section>
+```
+
+In the above example, with the `.ignore-nullish`, when the page finished loading it will retain the "Not loaded yet" HTML. Clicking the button will then replace that HTML with the list items.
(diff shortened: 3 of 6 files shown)


ekwoka:
I like the modifier approach.

I think I'd be more partial to `preserve-nullish` as opposed to `ignore` but that's just my opinion.

packages/alpinejs/src/directives/x-html.js
```diff
@@ -2,18 +2,22 @@ import { directive } from '../directives'
...
         evaluate(value => {
-            mutateDom(() => {
-                el.innerHTML = value
+            value = value ?? null;
 
-                el._x_ignoreSelf = true
-                initTree(el)
-                delete el._x_ignoreSelf
-            })
+            if (!modifiers.includes('ignore-nullish') || value !== null) {
```
ekwoka, line 12:
```suggestion
            if (!modifiers.includes('ignore-nullish') || value == null) {
```

Then you can remove the above `value = value ?? null`

ekwoka, line 12:
```suggestion
            if (!modifiers.includes('ignore-nullish') || value != null) {
```

NIT: Then you can remove the above `value = value ?? null`

packages/alpinejs/src/directives/x-text.js
```diff
@@ -1,14 +1,18 @@
...
     let evaluate = evaluateLater(expression)
 
     effect(() => {
         evaluate(value => {
-            mutateDom(() => {
-                el.textContent = value
-            })
+            value = value ?? null;
+
+            if (!modifiers.includes('ignore-nullish') || value !== null) {
```
ekwoka, line 11:
```suggestion
            if (!modifiers.includes('ignore-nullish') || value != null) {
```

NIT

packages/alpinejs/src/directives/x-html.js
willrowe: I would rather explicitly coerce to `null` and be able to do a strict check. It makes it more clear what the intent is here.

willrowe: I would rather explicitly coerce to `null` and be able to do a strict check. It makes it more clear what the intent is here.

packages/alpinejs/src/directives/x-text.js
willrowe: I would rather explicitly coerce to `null` and be able to do a strict check. It makes it more clear what the intent is here.

ekwoka:
Yeah....but `preserve-on-nullish` is too long 😭

Tim-Wils:
What is the value becomes `null` again after being set to a value? Shouldn't the value become the original innerHTML again with this modifier?

calebporzio:
## PR Review: #4579 — Add `ignore-nullish` modifier to `x-html` and `x-text`

**Type**: Feature
**Verdict**: Needs discussion

This PR adds a new `.ignore-nullish` modifier to `x-text` and `x-html`. The idea:

1. You have server-rendered HTML like `<h1 x-text.ignore-nullish="title">Loading...</h1>`
2. Your data starts as `null` — maybe it hasn't loaded from an API yet
3. Without the modifier, Alpine immediately sets the text to `null` (the string), wiping your placeholder
4. With `.ignore-nullish`, Alpine skips the update when the value is `null` or `undefined`, preserving whatever text/HTML was already there

**1. Use the nullish coalescing operator in the expression (already works today):**
```html
<h1 x-text="title ?? 'Not loaded yet'">Not loaded yet</h1>
```
For `x-html`:
```html
<section x-html="items ?? $el.innerHTML">
    <em>Loading...</em>
</section>
```
This is zero new API surface and already handles the use case. The only downside is the `$el.innerHTML` approach for x-html is slightly verbose for complex fallback HTML — but it works.

**2. Use `x-show` or `x-if` to conditionally display:**
```html
<h1 x-show="title" x-text="title"></h1>
<h1 x-show="!title">Not loaded yet</h1>
```
More verbose but also zero new API.

**3. This PR's approach — a dedicated modifier.** Saves a few characters in the expression but adds permanent public API surface to two directives.

### Changes Made

No changes made. This is a feature scope question for Caleb.

### Test Results

- `x-text.spec.js`: 4/4 passing (with fix), 3/4 passing (without fix — new test correctly fails)
- `x-html.spec.js`: 5/5 passing (with fix), 4/5 passing (without fix — new test correctly fails)
- CI: passing

The tests are well-written and correctly isolate the feature behavior. They fail without the implementation, confirming they're testing the right thing.

### Code Review

**If this feature were to be accepted, these issues need fixing first:**

1. **Style: semicolons** — `x-html.js:10` and `x-text.js:9` both have `value = value ?? null;` with semicolons. Alpine doesn't use semicolons.

2. **Style: trailing whitespace** — `x-html.js:19` and `x-text.js:14` have trailing spaces after `})`.

3. **Naming: `ignore-nullish` breaks convention** — Nearly every Alpine modifier is a single word (`fill`, `lazy`, `trim`, `camel`, `important`, `immediate`). `ignore-nullish` is a hyphenated two-word modifier. The word "nullish" is also JavaScript jargon — most Alpine users think in terms of "null" or "empty", not the TC39 spec term "nullish."

4. **Malformed HTML in test** — `x-html.spec.js:59`: the `<button>` tag is never closed (missing `>`). It happens to work because browsers are forgiving, but it's sloppy.

5. **The `value ?? null` coercion is indirect** — The code converts `undefined` to `null`, then only checks `!== null`. This works but is roundabout. `value == null` (loose equality) already catches both `null` and `undefined` in one check, no coercion needed.

### Security

No security concerns identified. The modifier only gates whether `innerHTML`/`textContent` is set — it doesn't change what values can be set.

### Verdict

**Lean toward closing.** The use case is real but already solvable with existing Alpine expressions (`x-text="title ?? 'fallback'"`). Adding a modifier for this means:

- Permanent API surface on two directives to maintain forever
- A naming precedent (`ignore-nullish`) that breaks Alpine's single-word modifier convention
- JavaScript jargon ("nullish") in the public API that non-JS-expert users won't immediately grok
- Zero community demand (no reactions on the PR or linked issues)

The "laziest correct solution" here is the one Alpine already has: `??` in expressions. A modifier saves maybe 10 characters per usage but costs ongoing API complexity.

If Caleb still wants this feature, the naming should be reconsidered — something single-word like `.optional` or `.nullable` would fit Alpine conventions better. And the style issues above need fixing.

---
*Reviewed by Claude*

PR #4588, opened 2025-03-21
use absolute path.

jimmywarting:
use explicit path in imports (See #4587)

### DIFF
--- a/packages/alpinejs/builds/cdn.js
+++ b/packages/alpinejs/builds/cdn.js
@@ -1,4 +1,4 @@
-import Alpine from './../src/index'
+import Alpine from './../src/index.js'
 
 window.Alpine = Alpine
 


calebporzio:
## PR Review: #4588 — use absolute path.

**Type**: Refactor (cosmetic)
**Verdict**: Close

This PR adds `.js` to one import path in `packages/alpinejs/builds/cdn.js`:

```js
// Before
import Alpine from './../src/index'

// After
import Alpine from './../src/index.js'
```

The idea is that explicit file extensions in import specifiers are "more correct." The linked issue #4587 presumably argues for this.

1. **Do nothing** — esbuild (Alpine's bundler) resolves extensionless imports without issue. The build output is byte-for-byte identical with or without `.js`. This is the right approach.
2. **Fix all inconsistent files** — If we did want consistency, this PR misses 3 other files with the same extensionless pattern: `packages/alpinejs/builds/module.js`, `packages/csp/builds/cdn.js`, and `packages/csp/builds/module.js`. A consistency fix that only fixes 1 of 4 files isn't consistent.
3. **Adopt a lint rule** — If explicit extensions mattered, enforce it project-wide. But they don't matter here because these files are build entry points consumed exclusively by esbuild.

### Changes Made
No changes made.

### Test Results
No tests to run — this is a cosmetic change to a build entry point. CI build check passed.

### Code Review
- The change itself is technically fine — adding `.js` to an import doesn't break anything.
- But it also doesn't fix anything. esbuild resolves `./src/index` to `./src/index.js` automatically.
- The PR only touches 1 of 4 files with this same pattern, so it doesn't even achieve consistency.
- Zero community engagement (no reactions, comments, or reviews).

### Security

### Verdict
**Close.** This is a zero-impact cosmetic change that doesn't fully achieve its own stated goal (consistency). Alpine's build system uses esbuild, which handles extensionless imports without issue — the build output is identical either way. The PR touches 1 of 4 files with this pattern, so merging it would actually make the codebase *less* consistent (1 file with `.js`, 3 without) until the others are fixed too. Not worth the merge noise. Thanks for the contribution though — the instinct toward explicit imports is reasonable in general, it just doesn't apply to bundled entry points.

---
*Reviewed by Claude*

PR #4589, opened 2025-03-21
use explicit import everywhere

jimmywarting:
superseeds #4588 

solved automatically using help from deno:
```bash
deno lint --include-rule=no-sloppy-imports --fix
```


ekwoka (approved):
I like it

PR #4600, opened 2025-04-09
[docs] Update mask.md: adding "x-data" to the mask plugin examples

erickcomp:
Hello, Caleb and Alpine crew.

After trying for some time to make the mask plugin work on my setup (I use it with both the Livewire bundled version and the standalone one), I could only make it work after reading this discussion, specifically this comment:

https://github.com/alpinejs/alpine/discussions/2920#discussioncomment-4831861
> hello bro i found the fix problem by adding x-data attribute
> 
> according to this article
> 
> ```
> <div class="w-full" >
> <input wire:model.defer="budget" x-data x-mask:dynamic="$money($input)" class="input input-bordered w-full" type="text" name="budget" id="budget">
> </div>
> ```
> 
> https://justinbyrne.dev/blog/2022/5/22/masking-input-fields-with-alpinejs
> 
> hope it help

After putting the x-data, everything started to work as it should. I then wondered why the docs examples did not use the x-data in the HTML, so I decided to create this PR.

Fun-fact: The docs themselves use alpine and the "live-code" examples in the md file *do use* the x-data attribute alongside x-mask:

image

So, this is the motivation for this PR. =)

### DIFF
--- a/packages/docs/src/en/plugins/mask.md
+++ b/packages/docs/src/en/plugins/mask.md
@@ -69,7 +69,7 @@ The primary API for using this plugin is the `x-mask` directive.
 Let's start by looking at the following simple example of a date field:
 
 ```alpine
-<input x-mask="99/99/9999" placeholder="MM/DD/YYYY">
+<input x-data x-mask="99/99/9999" placeholder="MM/DD/YYYY">
 ```
 
 <!-- START_VERBATIM -->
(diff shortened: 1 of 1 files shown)


erickcomp:
Just realized there are other PRs about this already. 😖
Sorry!

ekwoka:
Yeah, I did open a PR adding notes about this to more pages.

It's one of those issues of how to communicate the idea that `x-data` is necessary somewhere in the tree above this, but not on this specific element.

There is not showing it and people get confused when they don't have it, or showing it and people putting x-data on every little thing, tanking performance.

PR #4602, opened 2025-04-10
Fix x-for directive not working inside x-teleport

kevnk:
This PR fixes an issue where x-for directives are not processed correctly when they appear inside content that's been teleported using x-teleport.

### Problem

The directive processing order in Alpine is defined in directives.js. Currently, the for directive is processed before the teleport directive. This causes issues when using x-for inside an element that's being teleported, as the x-for directive attempts to process the content before it's moved by x-teleport.

This results in scenarios where the array data is accessible (as shown by x-effect logs), but the x-for loop doesn't process the items when the content is teleported.

### Solution

Two changes were made to fix this issue:
1. Changed the directive order in directives.js to place teleport at the beginning of the processing order.
2. Improved event forwarding in x-teleport.js to handle @event directives on template elements.

### Tests

Added specific tests for x-for inside x-teleport that verify:
- Basic x-for functionality inside teleported content.
- Reactivity when data changes.
- Support for complex data types (objects) in the array.

All existing tests for both x-teleport and x-for continue to pass.

### Breaking Changes

None. This change preserves all existing functionality while adding support for a currently broken use case.

### Disclaimer

100% Written by Claude Code

### DIFF
--- a/packages/alpinejs/src/directives.js
+++ b/packages/alpinejs/src/directives.js
@@ -200,6 +200,7 @@ function toParsedDirectives(transformedAttributeMap, originalAttributeOverride) 
 const DEFAULT = 'DEFAULT'
 
 let directiveOrder = [
+    'teleport',
     'ignore',
     'ref',
     'data',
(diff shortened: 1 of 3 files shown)


kevnk:
Simple examples show x-for inside an x-template working fine, so I'm not sure if this is the actual problem and solution for my project. It will require more investigation

PR #4640, opened 2025-07-02
Fix Chrome warning when using `x-trap.inert`

nikitaourazbaev:
Related discussion: https://github.com/alpinejs/alpine/discussions/4634

In recent versions of Chromium, using the `inert` modifier with `x-trap` results in this warning:

> Blocked aria-hidden on an element because its descendant retained focus. The focus must not be hidden from assistive technology users. Avoid using aria-hidden on a focused element or its ancestor. Consider using the inert attribute instead, which will also prevent focus. For more details, see the aria-hidden section of the WAI-ARIA specification at https://w3c.github.io/aria/#aria-hidden.

CleanShot 2025-06-12 at 18 08 39@2x

Here’s a minimal reproduction:
https://codepen.io/magiclantern/pen/vEOjaVg/9827b04ae16fffe57feb1f762870aeaa

This seems to happen because `aria-hidden` is applied by the Focus plugin _before_ focus is trapped in the new container. 

This PR moves the call to `setInert()` into `focus-trap`’s [`onPostActivate` callback](https://github.com/focus-trap/focus-trap?tab=readme-ov-file#createoptions), where it runs after a tick.

Another solution would be to replace `aria-hidden="true"` with `inert`, but [Safari compatibility for `inert` is only >= 15.5](https://caniuse.com/?search=inert), which may not be sufficient? Perhaps it could be an option, but I couldn’t come up with a good / short enough name for a modifier.

### DIFF
--- a/packages/focus/src/index.js
+++ b/packages/focus/src/index.js
@@ -108,6 +108,8 @@ export default function (Alpine) {
                 fallbackFocus: () => el,
             }
 
+            let undoInert = () => {}
+
             if (modifiers.includes('noautofocus')) {
                 options.initialFocus = false
             } else {
@@ -116,9 +118,16 @@ export default function (Alpine) {
                 if (autofocusEl) options.initialFocus = autofocusEl
             }
 
+            if (modifiers.includes('inert')) {
+                options.onPostActivate = () => {
+                    Alpine.nextTick(() => {
+                        undoInert = setInert(el);
+                    });
+                }
+            }
+
             let trap = createFocusTrap(el, options)
 
-            let undoInert = () => {}
             let undoDisableScrolling = () => {}
 
             const releaseFocus = () => {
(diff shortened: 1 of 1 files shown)


ekwoka (approved):
Simple and clean

calebporzio (commented):
This PR looks good to merge! 

## Summary:
- Fixes a Chrome accessibility warning when using `x-trap.inert`
- The warning occurred because `aria-hidden` was being applied before focus was properly trapped
- Solution moves the `setInert()` call into focus-trap's `onPostActivate` callback, ensuring it runs after focus is trapped

## Key points:
✅ Has approval from ekwoka
✅ All CI checks passing
✅ Minimal, targeted fix that addresses the specific issue
✅ Includes clear explanation and reproduction case
✅ Maintains backward compatibility

The fix is clean and follows Alpine's patterns. It properly addresses the timing issue that was causing the Chrome warning without breaking existing functionality.

PR #4644, opened 2025-07-09
style: use let/const instead of var in debounce.js

GeoDaoyu:


ekwoka:
Simple and clean

ekwoka:
I'm not the one that makes those decisions.

PR #4671, opened 2025-08-30
Improve CSP build

calebporzio:


utsav-infynno:
```
If you wish to only run Jest tests, you can run npm run jest like normal and target specific tests. You can specify command line config options to forward to the jest command with -- like so: npm run jest -- --watch
```

Forgot to update the `readme.md`

@calebporzio

PR #4712, opened 2025-12-16
update `x-cloak` CSS to allow conditional disabling

browner12:
change the suggested `x-cloak` CSS to specifically only target the boolean `x-cloak` attribute or the attribute with an empty string value (`x-cloak=""`).

this means we will no longer target the attribute when it has a non-empty string value (`x-cloak="off"`).

the benefit this gives is we can conditionally disable the cloaking for elements we know will be visible immediately. this prevents a "blip in" from the hidden to visible state.

Here is a Blade example to show how it would work:

```blade
<div
    x-show="jsCondition"
    x-cloak="{{ $condition ? 'off' : '' }}"
>Alpine Rocks</div>
```

this works because, given these elements:

```html
<div x-cloak></div>
<div x-cloak=""></div>
<div x-cloak="off"></div>
```

we get this behavior:

| CSS Selector    | Matches        |
|-----------------|----------------|
| [x-cloak]       | all three      |
| [x-cloak=""]    | first two only |
| [x-cloak="off"] | third only     |

---

This ___is___ already possible by conditionally showing the `x-cloak`, but I think this solution provides a better looking code.

```blade
<div
    x-show="jsCondition"
    @if($condition)
        x-cloak
    @endif
>Alpine Rocks</div>
```

### DIFF
--- a/packages/docs/src/en/directives/cloak.md
+++ b/packages/docs/src/en/directives/cloak.md
@@ -12,13 +12,19 @@ Sometimes, when you're using AlpineJS for a part of your template, there is a "b
 For `x-cloak` to work however, you must add the following CSS to the page.
 
 ```css
-[x-cloak] { display: none !important; }
+[x-cloak=""] { display: none !important; }
 ```
 
 The following example will hide the `<span>` tag until its `x-show` is specifically set to true, preventing any "blip" of the hidden element onto screen as Alpine loads.
 
 ```alpine
 <span x-cloak x-show="false">This will not 'blip' onto screen at any point</span>
+```
+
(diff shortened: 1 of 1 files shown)


browner12:
Gotta say I disagree with this verdict. It's not a CSS "trick". It's us recommending a more restrictive selector that makes it easier for implementing code to toggle behavior.

The bigger problem is 3rd parties that include the original suggestion in their own code, which can leak into user code, and prevent the end user from making this decision for themself.

https://github.com/laravel/framework/blob/bcaa68a6587ecf26818922583a9070b0bf4e240c/src/Illuminate/Foundation/resources/exceptions/renderer/styles.css#L6

Issue #450, opened 2020-05-02
[Bug] x-for should generate valid/compliant html when used on elements such as lists, tables, selects etc


OwenMelbz:
Currently the `x-for` directive requires you to wrap the element in a `<template>` tag, however, this produces invalid HTML when used with `<ul>`, `<ol>`,`<table>`, `<select>` etc etc.

As the HTML spec say only `<li>` elements can live within a ol, or ul - they cannot accept `<template>` (which also messes up css like first-child)

E.g. you're required to do..

```html
<ul>
    <template x-for="item in items">
        <li x-text="item"></li>
    </template>
</ul>
```

This obviously isn't valid, although Vue/React/etc can get around this by their virtual dom as they remove the `<template>` tags during rendering I guess? 

So it actually produces

```html
<ul>
    <template x-for="item in items">
         <li x-text="item">d</li>
    </template>
    <li x-text="item">1</li>
    <li x-text="item">2</li>
    <li x-text="item">3</li>
    <li x-text="item">4</li>
</ul>
```

Although I understand this is probably quite a tricky problem, lists are the fundamental reason for loops such as `x-for` and you'd expect to be able to do this without side effects, maybe even accessibility issues?

The same problem is present when building HTML tables, another common use case for `x-for`

```html
<table>
   <template x-for="item in items">
        <tr>
            <td x-text="item"></td>
        <tr>
    </template>
</table>
```

which outputs

```html
<table>
    <template x-for="item in items">
        <tr>
            <td x-text="item"></td>
        </tr>
    </template>
    <tr>
        <td x-text="item">1</td>
    </tr>
    <tr>
        <td x-text="item">2</td>
    </tr>
    <tr>
        <td x-text="item">3</td>
    </tr>
    <tr>
        <td x-text="item">4</td>
    </tr>
</table>
```

Same for `selects` etc...

```html
<select>
	<template x-for="item in items">
		<option x-text="item"></option>
	</template>
	<option x-text="item">1</option>
	<option x-text="item">2</option>
	<option x-text="item">3</option>
	<option x-text="item">4</option>
</select>
```

Mozilla actually provide an `<template>`, with tables where the template is defined outside of the markup to make sure this doesn't happen: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/template - maybe something like this could be investigated to solve this issue?

-----------

[...]

PR #4750, opened 2026-02-18
Support multiple root elements in x-for templates

calebporzio:
## Situation

`x-for` renders one element per iteration. Works great — until you need siblings:

```html
<tr>
    <template x-for="(color, idx) in colors">
        <td x-text="color"></td>
        <td x-text="idx"></td>
    </template>
</tr>
```

### What happens:

1. You put two `<td>`s in the template
2. Only the **first** `<td>` renders per iteration — the second is silently dropped
3. You try wrapping in a `<div>` — browser strips it (invalid inside `<tr>`)
4. No warning. No error. Just missing elements.

Known pain point since 2020 (#450, discussion #935).

## Problem

`x-for`'s clone logic hardcodes single-element extraction:

```js
let clone = document.importNode(templateEl.content, true).firstElementChild
```

- `.firstElementChild` grabs one element, discards the rest
- The lookup map, reordering, cleanup, and scope all assume one element per iteration
- No warning when children are dropped

## Solutions considered

| # | Approach | Verdict |
|---|----------|---------|
| 1 | Docs-only (original PR #4730) | ❌ Doesn't fix it. Several workarounds empirically broken. |
| 2 | Always use comment markers | ❌ Unnecessary overhead for single-element templates. |
| 3 | Auto-wrap in table elements | ❌ No valid HTML wrapper for `<td>` siblings inside `<tr>`. |
| 4 | New directive (`x-for-each`) | ❌ Two directives for one concept. Confusing. |
| 5 | Console warning | ⚡ Shipped separately in #4752. Visible failure, not a fix. |
| 7 | Fragment class abstraction | ❌ Overengineered. Comment markers are simpler. |
| **11** | **Auto-detect fragment mode** | ✅ **Chosen** |

## Chosen approach: auto-detect fragment mode

When `templateEl.content.children.length > 1` → switch to fragment mode with comment-node boundaries.
When `=== 1` → existing code, unchanged, zero overhead.

**Why:**
- 🟢 Vue 3 (fragment VNodes) and Svelte (comment anchors) both solved it this way
- 🟢 Alpine's morph plugin already uses this pattern (`Block` class with `startComment`/`endComment`)
- 🟢 `_x_lookup` and `_x_refreshXForScope` are fully internal — no public API changes
- 🟢 Comment nodes invisible to CSS, don't affect `:nth-child`, survive inside `<tr>`

Based on the problem identified in #4730 by @NightFurySL2001.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/docs/src/en/directives/for.md
+++ b/packages/docs/src/en/directives/for.md
@@ -49,10 +49,9 @@ You may also pass objects to `x-for`.
 </div>
 <!-- END_VERBATIM -->
 
-There are two rules worth noting about `x-for`:
+There is one rule worth noting about `x-for`:
 
 > `x-for` MUST be declared on a `<template>` element.
-> That `<template>` element MUST contain only one root element
 
 <a name="keys"></a>
 ## Keys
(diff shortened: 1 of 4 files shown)


packages/docs/src/en/directives/for.md
```diff
@@ -115,21 +114,40 @@ If you need to simply loop `n` number of times, rather than iterate through an a
...
 <template x-for="color in colors">
     <p>
         <span>The next color is </span><span x-text="color">
     </p>
 </template>
 ```
+
+### Limitations with `<table>` elements
+
+HTML tables have strict parsing rules — a `<div>` is not valid inside `<tr>`, so the browser will strip it before Alpine runs. If you need to iterate table cells, either place multiple `<td>` elements directly in the template (as shown above) or use `<tr>` as the root element:
```
NightFurySL2001: Shouldn't it be "td is not valid inside div" ? And should be "strict template parsing rule", not "strict table parsing rule"

calebporzio:
closing because i don't trust this AI generated code lol

NightFurySL2001:
What is the plan forward? Will x-for be refactored in the future or should I reopen the docs to give more information on multi-child x-for?

PR #4769, opened 2026-03-08
fix: sync x-model select value after x-for renders options

ljk0071:
https://github.com/alpinejs/alpine/discussions/2697
https://github.com/alpinejs/alpine/discussions/4546

Fix x-model binding for select options rendered via x-for


calebporzio:
refactored to a different approach. THANKS!

PR #4773, opened 2026-03-17
Add .fixed positioning modifier to x-anchor

nicolagianelli:
## What

Adds a new `.fixed` modifier to the `x-anchor` directive that instructs Floating UI to use its `position: fixed` strategy instead of the default `position: absolute`.

```html
<div x-show="open" x-anchor.fixed="$refs.button">
    Dropdown content
</div>
```

## Why

`x-anchor` currently hardcodes `position: absolute` when applying styles. This works for most cases, but it breaks down when:

- The reference element lives inside a container with `overflow: hidden` / `overflow: clip` / `overflow: auto` — the popover gets clipped along with the parent.
- You want the popover to stay pinned to the viewport regardless of scroll.

Floating UI natively supports a `fixed` strategy for exactly this — Alpine just didn't have a way to pass it through.

## How

- `getOptions()` returns a new `strategy: 'fixed' | 'absolute'` field, derived from whether the `.fixed` modifier is present. Defaults to `'absolute'`, so existing `x-anchor` usage is byte-identical in behavior.
- `computePosition()` receives the `strategy` option.
- `setStyles()` takes the strategy as an argument (defaulting to `'absolute'`) and applies it as the CSS `position` value.
- The clone/morph branch also reads `strategy` from `getOptions(modifiers)` and forwards it to `setStyles()`. This is necessary for correctness: when `.fixed` is active, the `(x, y)` stored on `_x_anchor` are viewport-relative, so the pre-existing branch that called `setStyles(el, x, y)` would have applied viewport coordinates under `position: absolute` after a morph, placing the popover in the wrong spot. This PR implicitly fixes that.

## Backward compatibility

Fully additive. `strategy` defaults to `'absolute'` in every code path, and `setStyles` defaults its parameter to `'absolute'` too. Any existing `x-anchor` expression without the `.fixed` modifier produces identical output before and after this change.

## Documentation

- Tightens the Fixed positioning section's lede to describe when absolute falls short.
- Adds a prominent heads-up about the transformed-ancestor gotcha — if an ancestor has `transform`, `filter`, `perspective`, `backdrop-filter`, `will-change`, or `contain`, the CSS spec says it becomes a new containing block for `position: fixed` descendants, which is the #1 footgun when `.fixed` appears to do nothing.
- Adds a note under Manual styling explaining how to combine `.no-style` with `.fixed` correctly (the caller must apply `position: fixed` themselves, because `$anchor.x/y` are returned in the coordinate space of whichever strategy is active).

## Tests

New Cypress tests in `tests/cypress/integration/plugins/anchor.spec.js`:

- `.fixed` applies `position: fixed`
- absence of `.fixed` keeps the default `position: absolute` (backward-compat regression guard)
- modifier order is interchangeable (`.bottom.fixed` === `.fixed.bottom`)
- `.fixed` still correctly anchors when the reference lives inside a `position: relative; overflow: hidden` clipping container
- `.fixed` strategy survives an `Alpine.morph()` pass — verifies the clone-branch `setStyles` call now correctly re-applies the strategy

All existing anchor tests continue to pass unchanged.

## Scope

The `.match-width` modifier that was present in an earlier version of this PR has been removed. It is orthogonal to `.fixed` (works with either positioning strategy) and deserves its own PR where its inline-width override behavior and bundle cost can be reviewed on their own merits. A follow-up PR will reintroduce it with docs and tests.

---

Original work by @nicolagianelli. Scope cleanup, tests, and docs expansion by @calebporzio.

### DIFF
--- a/packages/anchor/src/index.js
+++ b/packages/anchor/src/index.js
@@ -1,4 +1,4 @@
-import { computePosition, autoUpdate, flip, offset, shift, size } from '@floating-ui/dom'
+import { computePosition, autoUpdate, flip, offset, shift } from '@floating-ui/dom'
 
 export default function (Alpine) {
     Alpine.magic('anchor', el => {
@@ -14,7 +14,7 @@ export default function (Alpine) {
     })
 
     Alpine.directive('anchor', Alpine.skipDuringClone((el, { expression, modifiers, value }, { evaluate, effect, cleanup }) => {
-        let { placement, offsetValue, unstyled, strategy, matchWidth, allowFlip } = getOptions(modifiers)
+        let { placement, offsetValue, unstyled, strategy, allowFlip } = getOptions(modifiers)
 
         el._x_anchor = Alpine.reactive({ x: 0, y: 0 })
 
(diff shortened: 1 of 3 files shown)


raquelpn1979-tech (commented):
a63fc30d28d41b50d8249306993c0f3818c2e047

PR #4776, opened 2026-03-19
Fix function assignments in x-init

khepin:
See: https://github.com/alpinejs/alpine/discussions/4775

During `x-init` if creating an arrow function to assign to a property (`x-init="$el.fn = () => { flag = true }"`), it will be executed immediately (interpreted as a block and not a function).
If a statement precedes that declaration, it is interpreted as a function and not executed immediately (`x-init="0; $el.fn = () => { flag = true }"`)

Minimal repro HTML:

```html
	<html>
		<head>
			<script defer src="https://cdn.jsdelivr.net/npm/alpinejs@3/dist/cdn.min.js"></script>
		</head>
		<body>
			<h2>Alpine x-init arrow function bug</h2>
			<p>Bug: flag should be "false" but shows "true"</p>
			<div x-data="{ flag: false }" x-init="$el.fn = () => { flag = true }">
				flag = <span x-text="flag"></span>
			</div>
			<p>Workaround: prepend a no-op statement</p>
			<div x-data="{ flag: false }" x-init="0; $el.fn = () => { flag = true }">
				flag = <span x-text="flag"></span>
			</div>
		</body>
	</html>
```

Result:

```
Alpine x-init arrow function bug
Bug: flag should be "false" but shows "true"

flag = true
Workaround: prepend a no-op statement

flag = false
```

I would expect both of those to give the same result, and I would expect both to function like the second case (prepended statement).

# resolution

Adds a regex in `generateFunctionFromString` to correctly recognize those situations.

### DIFF
--- a/packages/alpinejs/src/evaluator.js
+++ b/packages/alpinejs/src/evaluator.js
@@ -200,8 +200,8 @@ export function normalRawEvaluator(el, expression, extras = {}) {
             || /^[\n\s]*if.*\(.*\)/.test(expression.trim())
             // Support expressions starting with "let/const" like: "let foo = 'bar'"
             || /^(let|const)\s/.test(expression.trim())
-            ? `(async()=>{ ${expression} })()`
-            : expression
+                ? `(async()=>{ ${expression} })()`
+                : expression
 
         let func = new AsyncFunction(
             ["scope"],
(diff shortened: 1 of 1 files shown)


calebporzio:
idk, i see what you're saying but I don't love the implementation. I think I'm going to close until more report this and a better implemenation comes across. thanks!

PR #4809, opened 2026-04-10
docs: add alpine-turnout to plugins

rodezee:
This PR adds Alpine Turnout to the Plugins section.

Alpine Turnout is a lightweight, state-preserving router that treats your application routes as "Tracks." Unlike traditional routers that swap elements in and out of the DOM, Turnout utilizes an x-show approach to keep all routes persistent.
Why it's useful:

    Persistence: Because route elements are never destroyed, they maintain their internal state (scroll positions, form inputs, active intervals, and media playback) when switching between pages.

    Declarative API: Uses simple x-route and x-title directives that feel native to Alpine.

    Automatic 404s: Includes a built-in customizable fallback for unregistered paths.

    Dynamic Scoping: Automatically injects URL parameters (e.g., /user/:id) into the component's scope for use with x-text or x-model.

    SEO Friendly: All route content is present in the initial DOM, allowing for immediate indexing by search crawlers.

    Transition Compatibility: AlpineJS Transitions work out of the box!

Links:

    Demo: [https://alpine-turnout.netlify.app/](https://alpine-turnout.netlify.app/)

    Repository: [https://github.com/rodezee/alpine-turnout]

    Size: ~1.4kb (minified)

I believe this fills a specific gap in the Alpine ecosystem for developers building dashboards or audio/video-centric apps that require navigation without state loss. This PR adds Alpine Turnout to the Third-Party Plugins section.

### DIFF
--- a/packages/docs/src/en/plugins/turnout.md
+++ b/packages/docs/src/en/plugins/turnout.md
@@ -1,0 +1,250 @@
+---
+order: 10
+title: Turnout
+description: A lightweight, persistent tab-style switch for Alpine.js
+graph_image: https://alpine-turnout.netlify.app/alpine-turnout.png
+---
+
+# Alpine Turnout Plugin
+
+## A lightweight, persistent tab-style switch for Alpine.js
+
+Unlike traditional routers that destroy and recreate DOM elements, **Alpine Turnout** treats your routes like railroad tracks.
+
+Every section stays in the DOM, preserving its internal state, while the "Turnout" guides the view and URL to the correct destination.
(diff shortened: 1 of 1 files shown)


joshhanley:
@rodezee thanks for sharing Alpine Turnout, it looks like an interesting plugin! However, the plugins section of the Alpine docs is reserved for first-party plugins that ship as part of the Alpine.js project. We don't include third-party plugin documentation here. So going to close this, but appreciate you contributing to the ecosystem!

PR #4810, opened 2026-04-11
feat(csp): support optional chaining and nullish coalescing in evaluator

dasm:
Add obj?.prop, obj?.[key], fn?.(), and a ?? b to the CSP evaluator, avoiding verbose ternary workarounds in templates. Route all optional paths through the existing checkForDangerousKeywords and checkForDangerousValues guards.

Like && and ||, ?? eagerly evaluates both operands rather than short-circuiting per spec. This is consistent with the existing evaluator pattern.

a?.b.c where a is null will throw rather than short-circuit the entire chain. Use a?.b?.c instead. Fixing this would require an OptionalExpression wrapper node to propagate short-circuit context.

Ref: ECMA-262 11.7: OptionalChainingPunctuator digit lookahead
Ref: ECMA-262 12.3.9: Optional Chains
Ref: ECMA-262 12.13: CoalesceExpression

### DIFF
--- a/packages/csp/src/parser.js
+++ b/packages/csp/src/parser.js
@@ -197,6 +197,12 @@ class Tokenizer {
         } else if (char === '-' && next === '-') {
             this.position += 2;
             this.tokens.push(new Token('OPERATOR', '--', start, this.position));
+        } else if (char === '?' && next === '?') {
+            this.position += 2;
+            this.tokens.push(new Token('OPERATOR', '??', start, this.position));
+        } else if (char === '?' && next === '.' && !this.isDigit(nextNext)) {
+            this.position += 2;
+            this.tokens.push(new Token('OPERATOR', '?.', start, this.position));
         }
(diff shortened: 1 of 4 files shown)


calebporzio:
thanks for PRing these additions to the CSP evaluator. However, I think I want to just wait a bit while we at least just release the property assignment stuff. Then go off demand - as users have real demand for things to be supported we add just that. I don't want to the parser to become an attempt at a full JS parser (at least not right now). thanks!

dasm:
Thanks @calebporzio for consideration.
When I worked on my pet project, I noticed it was missing. I considered it would be nice addition. However, I understand that it might be something out of scope for alpine!
Thanks for the project, I was able to simplify a lot of my own code in the app.

calebporzio:
love to hear it! Yeah...I see the use for it, just want to make totally sure it's universally demanded before expanding the parser to stuff like that. all the best!

PR #4811, opened 2026-04-11
feat(csp): support template literals in evaluator

dasm:
The CSP evaluator does not recognize backtick strings, so `value: ${x}` throws while the equivalent 'value: ' + x works. This forces users into string concatenation that standard Alpine does not require.

Add template literal parsing with ${...} interpolation support. Expressions inside interpolations are fully parsed, so member access, arithmetic, and ternaries all work. null and undefined interpolations resolve to empty string, matching native JS behavior.

Ref: ECMA-262 11.8.6: Template Literal Lexical Components

### DIFF
--- a/packages/csp/src/parser.js
+++ b/packages/csp/src/parser.js
@@ -37,6 +37,8 @@ class Tokenizer {
                 this.readIdentifierOrKeyword();
             } else if (char === '"' || char === "'") {
                 this.readString();
+            } else if (char === '`') {
+                this.readTemplateLiteral();
             } else if (char === '/' && this.peek() === '/') {
                 this.skipLineComment();
             } else {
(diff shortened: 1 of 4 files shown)


danie-ramdhani:
This is cool sir. Maybe need to update the docs too https://github.com/alpinejs/alpine/blob/main/packages/docs/src/en/advanced/csp.md#whats-not-supported
and maybe update in livewire repo too after this merged? https://github.com/livewire/livewire/blob/main/docs/csp.md#whats-not-supported

calebporzio:
idk... i feel like we don't really need to support them in the CSP version right? like how often are people using template literals inside alpine expressions?

calebporzio:
closing - same reason as here: https://github.com/alpinejs/alpine/pull/4810

dasm:
I actually used it here and there, like below. There are other ways to solve the same problem.
```
            <div class="modal-header">
                <h2 class="modal-title" id="edit-user-modal-title"
                    x-text="'Edit User: ' + editUserForm.email">Edit User</h2>
                <button type="button" class="modal-close" x-on:click="closeEditUser()" aria-label="Close">&times;</button>
            </div>
 ```

PR #4812, opened 2026-04-11
Fix mask resurrecting key in object when it is cleared

danharrin:
When a reactive object is cleared (e.g. `form = {}`), Alpine's x-model effect fires `_x_forceModelUpdate(undefined)` on inputs still connected to the DOM. The mask plugin's wrapper around this function converts `undefined` to the string `"undefined"` via `String(value)`, formats it to `""`, then calls `el._x_model.set('')` — which evaluates `form.amount = ''`, resurrecting the deleted key with an empty value. In a Livewire context, this stale key gets sent back with the next server request, corrupting server-side state.

Steps to reproduce:
- Have an input with `x-model="form.amount"` and `x-mask:dynamic="$money($input)"`
- Set `form.amount` to a value (e.g. `'5000'`)
- Clear the parent object: `form = {}`
- Observe that `form.amount` reappears as an empty string

## Findings

The mask plugin wraps `_x_forceModelUpdate` (defined in x-model.js) to apply formatting before updating the DOM. The original function already handles `undefined` correctly for nested model paths (converts to `''` and updates the input). The issue is that the mask wrapper runs `String(value)` and `el._x_model.set(value)` unconditionally — even when the value is `undefined` because the model path no longer exists.

## Proposed solution

Add an early return in the mask wrapper when `value === undefined`. This skips the formatting/write-back path entirely and delegates to the original `_x_forceModelUpdate`, which already handles `undefined` properly. Crucially, `el._x_model.set()` is not called, so the deleted key is not recreated.

### DIFF
--- a/packages/mask/src/index.js
+++ b/packages/mask/src/index.js
@@ -52,6 +52,14 @@ export default function (Alpine) {
 
                 let updater = el._x_forceModelUpdate
                 el._x_forceModelUpdate = (value) => {
+                    // If the model value was cleared (e.g. the parent object was
+                    // removed), just clear the input — don't format and write back
+                    // as that would resurrect the model path with an empty value.
+                    if (value === undefined) {
+                        lastInputValue = ''
+                        return updater(value)
+                    }
+
                     value = String(value)
                     let template = templateFn(value)
                     if (template && template !== 'false') {
--- a/tests/cypress/integration/plugins/mask.spec.js
+++ b/tests/cypress/integration/plugins/mask.spec.js
@@ -1,4 +1,4 @@
-import { haveData, haveValue, html, test } from '../../utils'
+import { haveData, haveText, haveValue, html, test } from '../../utils'
 
 test('x-mask',
     [html`<input x-data x-mask="(999) 999-9999">`],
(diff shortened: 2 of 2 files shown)


ekwoka:
Why is this the way you want to clear the form state?

Can you not use `form.reset()`?

danharrin:
This is not for clearing form state, this is part of a larger system that manages nested modal stacks that can contain forms.

Tugzrida:
I was just working on a very similar patch then found this. I'd like to suggest adding `null` to the early return alongside `undefined`.

We've also experienced issues caused by the unconditional String cast introduced in #4175, where inputs we have masked for a MAC address style format are filled with `nu:ll` (standalone inputs) or `un:de:fi:ne:d` (inputs as part of a filament repeater, similar to where @danharrin came across it).

If a further failing test is wanted to demonstrate the issue, this is the one I was going to include in my patch:

```js
test('x-mask excludes null and undefined',
    [html`
        <div x-data="{ value: '00:00:00:00:00:00' }">
            <input type="text" x-model="value" x-mask="**:**:**:**:**:**">
            <button id="1" @click="value = null">Change</button>
            <button id="2" @click="value = undefined">Change</button>
        </div>
    `],
    ({ get }) => {
        get('#1').click()
        get('input').should(haveValue(''))
        get('div').should(haveData('value', null))
        get('#2').click()
        get('input').should(haveValue(''))
        get('div').should(haveData('value', undefined))
    }
)
```

danharrin:
In general, I just want masked inputs to behave exactly the same as those using `x-model` when the underlying parent object is cleared. What's your use case for setting the state to `null`?

Tugzrida:
We're not explicitly setting the state to `null`, that's just the default state for a filament TextInput. We're seeing that when other fields are updated server-side, alpine is applying the mask to `null` as a string and resulting in `nu:ll` in the field.

For unmasked inputs, setting the model to `null` results in an empty input, whereas currently masked inputs will cast `null` to a string then apply the mask to it. It would make sense to me if this behaviour were the same regardless of whether the input has a mask.

MaxencePaulin:
Hey, i think it's related to https://github.com/livewire/livewire/pull/10202 no ?

I don't know if Livewire release that on v3. But it's already the approach on v4.

calebporzio:
Perfect, thanks!

PR #4821, opened 2026-04-21
Fix `x-id` evaluation order so `$id` resolves consistently inside `x-data`

### LINKED ISSUE #4817: Order of execution on element with x-id and $id in x-data
### Alpine.js version

v3.15.11 (didn't work in v3.15.8 either)

Chrome v147 on Windows 11, probably browser unrelated

If an element has `x-id=['name']` and on the same element `x-data` with magic `$id('name')` is used, the magic `$id('name')` creates a new context on the fly. Only after that happened, `x-id` is evaluated: Alpine sees that the id-name has been used before, counts it up and sets this as the new id-context. Further calls to `$id('name')` are working as expected.

[Minimal reproduction on Codesandbox](https://codesandbox.io/p/sandbox/yz4sps)

```
  <body x-data="{ outerScopeOnlyForXIdToWork: true }">
    <h1>Example a</h1>
    <div
      x-id="['a']"
      x-data="{ idFromData: $id('a'), anotherIdFromData: $id('a') }"
    >
      <p>
        1st ID from magic inside x-data:
        <strong x-text="idFromData"></strong> (expected: a-1)
      </p>
      <p>
        2nd ID from magic inside x-data:
        <strong x-text="anotherIdFromData"></strong> (expected: a-1)
      </p>
      <p>
        1st ID from magic inside x-text:
        <strong x-text="$id('a')"></strong> (expected: a-1)
      </p>
      <p>
        2nd ID from magic inside x-text:
        <strong x-text="$id('a')"></strong> (expected: a-1)
      </p>
    </div>
    <h1>Example b</h1>
    <div x-id="['b']">
      <div x-data="{ idFromData: $id('b'), anotherIdFromData: $id('b') }">
        <p>
          ID from magic inside x-data:
          <strong x-text="idFromData"></strong> (expected: b-1)
        </p>
        <p>
          2nd ID from magic inside x-data:
          <strong x-text="anotherIdFromData"></strong> (expected: b-1)
        </p>
        <p>
          1st ID from magic inside x-text:
          <strong x-text="$id('b')"></strong> (expected: b-1)
        </p>
        <p>
          2nd ID from magic inside x-text:
          <strong x-text="$id('b')"></strong> (expected: b-1)
        </p>
      </div>
    </div>
  </body>
```

Example a shows the bug, example b shows a workaround by using `x-id` on a parent element.

_No response_

I would expect, that the usage of `$id` inside `x-data` would respect the `x-id` context. It looks to me as if the `x-id`-context is established only after `x-data` has been evaluated.

Maybe it is a viable solution to not count up the found index inside [packages/alpinejs/src/ids.js#L17](https://github.com/alpinejs/alpine/blob/main/packages/alpinejs/src/ids.js#L17) (via the call to `findAndIncrementId(name)`)

joshhanley:
# The Scenario

When `x-id` and `x-data` are on the same element, `$id('name')` inside the `x-data` expression returns a different id than `$id('name')` in descendants — breaking ARIA wiring that relies on them matching.

```html
<div
    x-id="['a']"
    x-data="{ idFromData: $id('a') }"
>
    <p>From <code>x-data</code>: <strong x-text="idFromData"></strong></p>   <!-- a-1 -->
    <p>From <code>x-text</code>: <strong x-text="$id('a')"></strong></p>     <!-- a-2 -->
</div>
```

# The Problem

`data` runs before `id` in the directive order. So `$id('a')` inside `x-data` finds no `_x_ids` on its ancestor yet and falls back to the global counter (returns `1`). Then `x-id` runs and calls `findAndIncrementId` again (returns `2`), which becomes the actual id root. Descendants resolve `a-2` while the `x-data` scope holds `a-1`.

# The Solution

Swap `id` ahead of `data` in `directiveOrder` so the id root is registered first.

`x-id` only takes a static array of name strings, so it doesn't need `x-data` scope to evaluate. Docs, tests, and real-world usage all use this pattern.

Fixes #4817

### DIFF
--- a/packages/alpinejs/src/directives.js
+++ b/packages/alpinejs/src/directives.js
@@ -206,8 +206,8 @@ const DEFAULT = 'DEFAULT'
 let directiveOrder = [
     'ignore',
     'ref',
-    'data',
     'id',
+    'data',
     'anchor',
     'bind',
     'init',
--- a/tests/cypress/integration/magics/$id.spec.js
+++ b/tests/cypress/integration/magics/$id.spec.js
@@ -128,6 +128,28 @@ test('$id scopes can be reset',
     }
 )
 
+test('$id resolves consistently when x-id and x-data are on the same element',
+    html`
+        <div x-data>
+            <div
+                x-id="['a']"
+                x-data="{ idFromData: $id('a'), anotherIdFromData: $id('a') }"
+            >
+                <p><span id="from-data" x-text="idFromData"></span></p>
+                <p><span id="from-data-2" x-text="anotherIdFromData"></span></p>
+                <p><span id="from-text" x-text="$id('a')"></span></p>
+                <p><span id="from-text-2" x-text="$id('a')"></span></p>
+            </div>
+        </div>
+    `,
+    ({ get }) => {
+        get('#from-data').should(haveText('a-1'))
+        get('#from-data-2').should(haveText('a-1'))
+        get('#from-text').should(haveText('a-1'))
+        get('#from-text-2').should(haveText('a-1'))
+    }
+)
+
 test('can be used with morph without losing track',
     [html`
         <div x-data>


calebporzio:
great!

PR #4822, opened 2026-04-22
Fix `@teleport` leaking detached nodes on every Livewire render

### LINKED ISSUE #10235: `@teleport` creates detached nodes on every render, causing memory leaks
### Livewire version

v4.2.4

### Laravel version

v13.4.0

### PHP version

v8.4.14

Brave on Windows

When using `@teleport` inside a Livewire component, every render cycle creates new detached DOM nodes that are never cleaned up.

It's worth noting that even on the very first render, a small number of detached nodes already exist — these are lightweight and likely tied to Livewire's initial hydration morph. They're not a concern by themselves.

## Environment:
- Livewire 3.x - 4.x
- Laravel 12.x - 13.x
- PHP 8.4.x - 8.5.x

## Why this matters:
For a simple <span>, the leak is manageable. But `@teleport` is typically used for modals, drawers, tooltips and other components with rich markup and Alpine state. In those cases, every re-render of the parent Livewire component leaks a full copy of that subtree, which can become significant over time in long-lived pages.

Adding `wire:ignore` around the `@teleport` block fixes the memory leak, but as mentioned above, some elements need to be both teleported and dynamic, making `wire:ignore` not a viable solution in those cases.

## Step to reproduce :
1. Create a livewire component with a button and a `@teleport` feature
2. Open DevTools and switch to Memory Tab, take a Heap Snapshot
3. Click a few times on the refresh button
4. Take another Heap Snapshot and compare, detached nodes will have accumulated

There is the same issue in CSP Safe mode and without it.

A live reproduction is available here : https://wirebox.app/b/4q0q2

```php
class Simple extends Component
{
    public function render()
    {
        return <<<'BLADE'
            <div>
                <div>
                    <button type="button" wire:click="action">
                        refresh
                    </button>
                </div>
                @teleport('body')
                    <div>
                        teleported content
                    </div>
                @endteleport
            </div>
        BLADE;
    }

    public function action(): void
    {
        //
    }
}
```

Initial state :
Image

After a few renders (17 detached nodes) :

Image

I believe there should be no memory leak in this feature, as `@teleport` can be very useful.

Possible approaches could be either a proper cleanup of teleported nodes before each re-render, or tracking teleported nodes per component so they can be correctly morphed/updated like the rest of the component. What do you think ?

joshhanley:
# The Scenario

A Livewire component containing a `@teleport` block leaks detached DOM nodes on every render. Each server round-trip accumulates another orphaned copy of the teleported subtree, so long-lived pages build up hundreds of retained clones. `wire:ignore` avoids the leak but isn't viable when the teleported content needs to stay reactive.

```php
<?php

use Livewire\Component;

new class extends Component {
    public function action(): void
    {
        //
    }
}; ?>

<div>
    <button wire:click="action">refresh</button>

    @teleport('body')
        <div>teleported content</div>
    @endteleport
</div>
```

# The Problem

When Alpine first boots a `<template x-teleport="#target">`, the directive clones the template content, places the clone inside the target, and stashes references on both sides so the morph plugin can find them later:

```js
let target = getTarget(expression)
let clone = el.content.cloneNode(true).firstElementChild

el._x_teleport = clone
clone._x_teleportBack = el
...
placeInDom(clone, target, modifiers)
```

Livewire updates the page by calling `Alpine.morph(from, newHtml)`. Morph parses the new HTML into a `to` tree and walks both trees in parallel, and for each pair of elements it calls `Alpine.cloneNode(from, to)` to re-run directives against `to` in "clone mode". Clone mode is Alpine's signal to directives to skip real side effects (DOM writes, subscriptions, cleanup registrations), the `to` tree is throwaway scaffolding for the morph, not something we actually want to wire up.

`x-teleport` didn't honour that. Even in clone mode it still called `placeInDom(clone, target, modifiers)`, and `getTarget()` fell back to a module-level singleton div:

```js
let teleportContainerDuringClone = document.createElement('div')

function getTarget(expression) {
    let target = skipDuringClone(() => {
        return document.querySelector(expression)
    }, () => {
        return teleportContainerDuringClone
    })()
    ...
}
```

That container is never cleared. Every morph of a teleport template appended a fresh clone into it, and because it's a live module-scoped `<div>` holding them as children, nothing could be garbage collected, one leaked subtree per render, forever.

# The Solution

The only thing the directive needed to stop doing in clone mode was `placeInDom`, the one call that attaches the clone to the singleton and creates the leak. Extending the existing `skipDuringClone` block to also cover it is enough:

```diff
     mutateDom(() => {
-        placeInDom(clone, target, modifiers)
-
         skipDuringClone(() => {
+            placeInDom(clone, target, modifiers)
+
             initTree(clone)
         })()
     })
```

The clone still exists on `to._x_teleport` so `@alpinejs/morph` can patch it via `context.patch(from._x_teleport, to._x_teleport)`, but it's no longer attached anywhere. The `_x_teleportPutBack` assignment and `cleanup` callback are still set on the `to` template in clone mode, but that's harmless, they live on the throwaway `to` and get garbage collected along with it.

Added a Cypress test that morphs a teleport template five times, then probes the container a fresh `Alpine.cloneNode` would use. Before the fix it found six leaked clones; after, the clone has no parent. Existing `x-teleport` and morph specs still pass.

Fixes livewire/livewire#10235

### DIFF
--- a/packages/alpinejs/src/directives/x-teleport.js
+++ b/packages/alpinejs/src/directives/x-teleport.js
@@ -8,6 +8,8 @@ import { warn } from "../utils/warn"
 directive('teleport', (el, { modifiers, expression }, { cleanup }) => {
     if (el.tagName.toLowerCase() !== 'template') warn('x-teleport can only be used on a <template> tag', el)
 
+    let target = getTarget(expression)
+
     let clone = el.content.cloneNode(true).firstElementChild
 
     // Add reference to element on <template x-teleport, and visa versa.
@@ -44,37 +46,38 @@ directive('teleport', (el, { modifiers, expression }, { cleanup }) => {
         }
     }
 
-    // During clone mode (used by Alpine.morph to seed new server HTML), the clone
-    // only needs to exist as `el._x_teleport` so morph can patch it. Attaching it
-    // to the DOM or registering its cleanup would leak a detached node every morph.
-    skipDuringClone(() => {
-        let target = getTarget(expression)
-
-        mutateDom(() => {
+    mutateDom(() => {
+        skipDuringClone(() => {
             placeInDom(clone, target, modifiers)
 
             initTree(clone)
+        })()
+    })
+
+    el._x_teleportPutBack = () => {
+        let target = getTarget(expression)
+
+        mutateDom(() => {
+            placeInDom(el._x_teleport, target, modifiers)
         })
+    }
 
-        el._x_teleportPutBack = () => {
-            let target = getTarget(expression)
-
-            mutateDom(() => {
-                placeInDom(el._x_teleport, target, modifiers)
-            })
-        }
-
-        cleanup(() =>
-            mutateDom(() => {
-                clone.remove()
-                destroyTree(clone)
-            })
-        )
-    })()
+    cleanup(() =>
+      mutateDom(() => {
+        clone.remove()
+        destroyTree(clone)
+      })
+    )
 })
 
+let teleportContainerDuringClone = document.createElement('div')
+
 function getTarget(expression) {
-    let target = document.querySelector(expression)
+    let target = skipDuringClone(() => {
+        return document.querySelector(expression)
+    }, () => {
+        return teleportContainerDuringClone
+    })()
 
     if (! target) warn(`Cannot find x-teleport element for selector: "${expression}"`)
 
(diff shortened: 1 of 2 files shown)


calebporzio:
thanks!

PR #4823, opened 2026-04-22
Fix `&&` and `||` not short-circuiting in CSP evaluator

### LINKED ISSUE #2571: Two toasts displayed for a single Flux::toast() call when using flux:toast.group with flux:toast inside
### Flux version

v2.13.2

### Livewire version

v4.2.4

### Tailwind version

v4.2.2

### Browser and Operating System

Browser: All browsers tested (Firefox, Chrome) - issue is not browser-specific | Operating System: Windows 11

### What is the problem?

When using <flux:toast.group position="top end"><flux:toast /></flux:toast.group> as shown in the official documentation, every Flux::toast() call displays two toasts simultaneously instead of one.
Investigation confirmed:

Only one ui-toast-group and one ui-toast exist in the DOM
Livewire response contains only one toast-show dispatch
showToast is called only once (confirmed via console.trace)
The issue occurs on all browsers (Firefox, Chrome) and all pages

The only workaround is to use <flux:toast /> alone without <flux:toast.group>, which correctly shows a single toast.

### Code snippets to replicate the problem

```
<flux:toast.group position="top end">
<flux:toast />
</flux:toast.group>
```

### Screenshots/ screen recordings of the problem

Image
<br>

Image

A single Flux::toast() call should display exactly one toast notification. When using <flux:toast.group><flux:toast /></flux:toast.group> as documented, only one toast should appear at a time.

- [x] I understand my bug report will be closed if I haven't met the criteria above.

joshhanley:
# The Scenario

In an app running Livewire's CSP build, a single `Flux::toast(...)` call renders two toasts instead of one when the page uses a grouped toast setup.

Screenshot 2026-04-23 at 07 46 01AM

```blade
<flux:toast.group position="top end">
    <flux:toast />
</flux:toast.group>
```

One toast appears correctly inside the group; a second appears separately, originating from the inner `<ui-toast>` element. Swapping the CSP build for the regular Livewire build makes it behave correctly.

# The Problem

Flux's inner `<ui-toast>` guards against handling the `toast-show` event when nested inside a group with a short-circuit expression:

```blade
<ui-toast x-on:toast-show.document="! $el.closest('ui-toast-group') && $el.showToast($event.detail)" ...>
```

Under the regular Alpine build this works correctly: when `$el.closest('ui-toast-group')` returns the group, the expression short-circuits and `showToast` is never called. Under the CSP build it doesn't, because the `BinaryExpression` handler in `packages/csp/src/parser.js` evaluates both operands up front before applying the operator:

```js
case 'BinaryExpression':
    const left = this.evaluate({ node: node.left, ... });
    const right = this.evaluate({ node: node.right, ... });

    switch (node.operator) {
        // ...
        case '&&': return left && right;
        case '||': return left || right;
    }
```

By the time `left && right` is computed, the right side has already run. Any `&&` or `||` expression with a side effect on the right leaks that side effect, regardless of the left value.

# The Solution

Wrap the right-hand evaluation in a function and only call it when short-circuiting doesn't apply:

```js
case 'BinaryExpression':
    const left = this.evaluate({ node: node.left, ... });

    // Wrapped in a function so && and || can skip it when short-circuiting.
    const evalRight = () => this.evaluate({ node: node.right, ... });

    // Short-circuit && and || so side-effects on the right aren't evaluated when the result is already determined.
    if (node.operator === '&&') return left && evalRight();
    if (node.operator === '||') return left || evalRight();

    const right = evalRight();

    switch (node.operator) { ... }
```

Screenshot 2026-04-23 at 07 47 38AM

Four new tests in `tests/vitest/csp-evaluator.spec.js` cover both the happy path (right side runs when needed) and the short-circuit path (right side does not run) for both `&&` and `||`.

Fixes livewire/flux#2571

### DIFF
--- a/packages/csp/src/parser.js
+++ b/packages/csp/src/parser.js
@@ -812,7 +812,15 @@ class Evaluator {
 
             case 'BinaryExpression':
                 const left = this.evaluate({ node: node.left, scope, context, forceBindingRootScopeToFunctions });
-                const right = this.evaluate({ node: node.right, scope, context, forceBindingRootScopeToFunctions });
+
+                // Wrapped in a function so && and || can skip it when short-circuiting.
+                const evalRight = () => this.evaluate({ node: node.right, scope, context, forceBindingRootScopeToFunctions });
+
+                // Short-circuit && and || so side-effects on the right aren't evaluated when the result is already determined.
+                if (node.operator === '&&') return left && evalRight();
+                if (node.operator === '||') return left || evalRight();
+
+                const right = evalRight();
 
                 switch (node.operator) {
                     case '+': return left + right;
@@ -828,8 +836,6 @@ class Evaluator {
                     case '>': return left > right;
                     case '<=': return left <= right;
                     case '>=': return left >= right;
-                    case '&&': return left && right;
-                    case '||': return left || right;
                     default:
                         throw new Error(`Unknown binary operator: ${node.operator}`);
                 }
--- a/tests/vitest/csp-evaluator.spec.js
+++ b/tests/vitest/csp-evaluator.spec.js
@@ -94,6 +94,54 @@ describe('cspRawEvaluator', () => {
         expect(cspRawEvaluator(element, 'true || false')).toBe(true)
         expect(cspRawEvaluator(element, '!false')).toBe(true)
     });
+
+    it('&& evaluates right side when left side is truthy', () => {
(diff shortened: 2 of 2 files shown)


codeldev:
Any ETA on when this may get merged? I have a client live project waiting on this fix :) Thanks!

PR #4825, opened 2026-04-24
Fix `x-for` crashing on null-prototype objects like `Object.groupBy()` results

### LINKED ISSUE #4824: Bug: Alpine x-for cannot handle objects returned by Object.groupBy
### Alpine.js version

v3.15.11

Linux Firefox

When using Object.groupBy() to group an array and passing the resulting object directly into an Alpine x-for, a runtime error occurs.

Object.groupBy() returns a plain object where keys are the group names and values are arrays. However, Alpine's x-for appears to have issues iterating over this structure when used in the form

```html
<script defer src="https://unpkg.com/alpinejs@3.x.x/dist/cdn.min.js"></script>

<div x-data="app()">
    <h2> gItems </h2>

    <template x-for="( group, groupName ) in gItems" :key="groupName">
        <div>
            <h3 x-text="groupName"></h3>

            <template x-for="item in group" :key="item.val">
                <div x-text="item.val"></span>
            </template>
        </div>
    </template>

    <h2> groupedItems() </h2>

    <template x-for="( group, groupName ) in groupedItems()" :key="groupName">
        <div>
            <h3 x-text="groupName"></h3>

            <template x-for="item in group" :key="item.val">
                <div x-text="item.val"></span>
            </template>
        </div>
    </template>

</div>
```

```js
function app() {
    return {
        items: [
            { type: 'a', val: 1 },
            { type: 'b', val: 2 },
            { type: 'a', val: 3 }
        ],
        gItems: {
            "A" : [ { type: "A", val: 10 }, { type: "A", val: 30 } ],
            "B" : [ { type: "B", val: 20 } ],
        },

        groupedItems() {
            return Object.groupBy( this.items, i => i.type )
        }
    }
}
```

[https://jsfiddle.net/d6rf0sxo/5/](https://jsfiddle.net/d6rf0sxo/5/)

x-for should be able to iterate over plain objects (as documented), where:

groupName represents the key
group represents the value (array of items)

```html
<template x-for="(group, groupName) in groupedItems">
```

joshhanley:
# The Scenario

Iterating over a null-prototype object with `x-for` (for example, the object returned by `Object.groupBy()`) throws `TypeError: Cannot convert object to primitive value` instead of iterating over its entries. A plain object literal with the same shape works fine.

```html
<div x-data="{
    items: [
        { type: 'a', val: 1 },
        { type: 'b', val: 2 },
        { type: 'a', val: 3 },
    ],
    groupedItems() {
        return Object.groupBy(this.items, i => i.type)
    },
}">
    <template x-for="(group, groupName) in groupedItems()" :key="groupName">
        <div>
            <h3 x-text="groupName"></h3>
            <template x-for="item in group" :key="item.val">
                <div x-text="item.val"></div>
            </template>
        </div>
    </template>
</div>
```

# The Problem

`x-for` uses `isNumeric(items)` to detect the `x-for="i in 100"` shortcut. That helper calls `isNaN(subject)`, which coerces its argument to a primitive via `valueOf`/`toString`. Null-prototype objects have no such methods, so the coercion throws. Plain object literals inherit from `Object.prototype` and coerce to `"[object Object]"` without issue, which is why the bug only shows up for things like `Object.groupBy()` (or any `Object.create(null)`-style object).

```js
function isNumeric(subject){
    return ! Array.isArray(subject) && ! isNaN(subject)
}
```

# The Solution

Filter out objects before calling `isNaN`. Objects are never numeric, so excluding them up front is both correct and sidesteps the coercion throw. `null` now falls out of the numeric path (since `typeof null === 'object'`), so the `undefined` guard is extended to cover it as well.

```js
function isNumeric(subject){
    // `isNaN` throws on null-prototype objects, so skip objects before calling it.
    return typeof subject !== 'object' && ! isNaN(subject)
}
```

Fixes #4824

### DIFF
--- a/packages/alpinejs/src/directives/x-for.js
+++ b/packages/alpinejs/src/directives/x-for.js
@@ -51,7 +51,7 @@ function loop(templateEl, iteratorNames, evaluateItems, evaluateKey) {
         if (isNumeric(items))
             items = Array.from({ length: items }, (_, i) => i + 1)
 
-        if (items === undefined) items = []
+        if (items === undefined || items === null) items = []
 
         // Support Set and Map objects by converting to arrays.
         if (items instanceof Set) items = Array.from(items)
@@ -188,7 +188,8 @@ function getIterationScopeVariables(iteratorNames, item, index, items) {
 }
 
 function isNumeric(subject){
-    return ! Array.isArray(subject) && ! isNaN(subject)
+    // `isNaN` throws on null-prototype objects, so skip objects before calling it.
+    return typeof subject !== 'object' && ! isNaN(subject)
 }
 
 function isObject(subject) {
--- a/tests/cypress/integration/directives/x-for.spec.js
+++ b/tests/cypress/integration/directives/x-for.spec.js
@@ -634,6 +634,20 @@ test('iterates over a Map',
     }
 )
 
+test('iterates over a null-prototype object (e.g. Object.groupBy result)',
+    html`
+        <div x-data="{ items: Object.assign(Object.create(null), { a: 'foo', b: 'bar' }) }">
+            <template x-for="(value, key) in items" :key="key">
+                <span x-text="key + ':' + value"></span>
+            </template>
+        </div>
+    `,
+    ({ get }) => {
+        get('span:nth-of-type(1)').should(haveText('a:foo'))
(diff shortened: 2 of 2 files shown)


calebporzio:
thanks!

PR #4833, opened 2026-05-11
Fix `:value` bindings not updating custom elements when bound object data mutates

### LINKED ISSUE #2414: Charts do not update when data is updated
### Flux version

v2.12 (pro)

### Livewire version

v4.1.4

### Tailwind version

>v4.2

### Browser and Operating System

Edge on Windows

### What is the problem?

Charts do not update when underlying data is changed - see also #1182 which was previously fixed. Reopening as a new issue as requested by @joshhanley 

In that issue, the problem was: _"The issue is that there is a wire:ignore on the root chart element, so any changes to the value attribute aren't being applied by Livewire and hence not being picked up by Flux. We'll have to see if we can move the wire:ignore further down the tree, to enable this to work."_

`wire:ignore` is definitely on the root chart element again. I have a temporary workaround of `wire:key="{{ Str::random(10) }}"` which means my chart will re-render when the date range is changed.

I've illustrated the issued in the code below where you can click change which updates the underlying data (changes the first term to 0) and you can see that is happening where the data is dumped below the chart.

### Code snippets to replicate the problem

```
<?php

use Livewire\Component;

new class extends Component {
    public array $data = [
        ['label' => 'Mon', 'revenue' => 1127],
        ['label' => 'Tue', 'revenue' => 249],
        ['label' => 'Wed', 'revenue' => 1365],
        ['label' => 'Thu', 'revenue' => 402],
        ['label' => 'Fri', 'revenue' => 690],
        ['label' => 'Sat', 'revenue' => 734],
        ['label' => 'Sun', 'revenue' => 828],
    ];
    public array $data2 = [
        ['label' => 'Mon', 'revenue' => 0],
        ['label' => 'Tue', 'revenue' => 249],
        ['label' => 'Wed', 'revenue' => 1365],
        ['label' => 'Thu', 'revenue' => 402],
        ['label' => 'Fri', 'revenue' => 690],
        ['label' => 'Sat', 'revenue' => 734],
        ['label' => 'Sun', 'revenue' => 828],
    ];

    public function change()
    {
        $this->data = $this->data2;
    }
};
?>

<div class="border border-gray-200 rounded-lg">
    <flux:button wire:click="change">Change</flux:button>
    <flux:chart wire:model="data" class="w-full aspect-2/1">
        <flux:chart.viewport class="size-full">
            <flux:chart.svg>
                <flux:chart.bar field="revenue" class="text-blue-500" radius="2 0" />

                <flux:chart.axis axis="x" field="label" position="bottom">
                    <flux:chart.axis.tick />
                    <flux:chart.axis.line />
                </flux:chart.axis>

                <flux:chart.axis axis="y" position="left">
                    <flux:chart.axis.grid />
                    <flux:chart.axis.tick />
                </flux:chart.axis>

                <flux:chart.cursor class="text-zinc-800" type="area" stroke-dasharray="4 4" />
            </flux:chart.svg>

            <flux:chart.tooltip>
                <flux:chart.tooltip.heading field="label" :format="['month' => 'long', 'day' => 'numeric']" />

                <flux:chart.tooltip.value field="revenue" label="Revenue">
                    <flux:chart.legend.indicator class="bg-blue-500" />
                </flux:chart.tooltip.value>
            </flux:chart.tooltip>
        </flux:chart.viewport>
    </flux:chart>
{{ var_dump($data) }}
</div>
```

### Screenshots/ screen recordings of the problem

Click the button in the component

Chart should update when the underlying data changes.

- [x] I understand my bug report will be closed if I haven't met the criteria above.

joshhanley:
# The Scenario

When `:value` (or `x-model`) binds an object to a custom element, mutations to nested properties don't propagate to the element:

```html
<script>
    customElements.define('value-display', class extends HTMLElement {
        get value() { return this._value }
        set value(v) {
            this._value = v
            this.textContent = JSON.stringify(v)
        }
    })
</script>

<div x-data="{ data: [{ label: 'Mon', value: 100 }] }">
    <value-display :value="data"></value-display>
    <button @click="data[0].value = 999">Mutate</button>
</div>
```

After clicking the button the display still shows `"value":100` instead of `999`.

# The Problem

`bindInputValue` short-circuits with a reference-equality check:

```js
if (el.value === value) return
```

The custom element's `value` getter returns the reactive proxy that was last set. When the proxy is the same reference between renders (its contents mutate but the top-level reference is preserved), the equality check is true and the setter is never re-invoked.

# The Solution

Skip the short-circuit when `value` is a non-null object:

```js
if (el.value === value && (typeof value !== 'object' || value === null)) return
```

Strings, numbers, booleans, and `null`/`undefined` still short-circuit as before.

Fixes livewire/flux#2414

### DIFF
--- a/packages/alpinejs/src/utils/bind.js
+++ b/packages/alpinejs/src/utils/bind.js
@@ -66,7 +66,12 @@ function bindInputValue(el, value) {
     } else if (el.tagName === 'OPTION') {
         bindAttribute(el, 'value', value)
     } else {
-        if (el.value === value) return
+        // For non-null objects, reference equality doesn't imply state
+        // equality: reactive proxies can mutate internally while the
+        // top-level reference stays the same. Without this guard, custom
+        // elements receiving an object value via x-bind:value or x-model
+        // would never see those nested updates.
+        if (el.value === value && (typeof value !== 'object' || value === null)) return
 
         el.value = value === undefined ? '' : value
     }
--- a/tests/cypress/integration/directives/x-bind.spec.js
+++ b/tests/cypress/integration/directives/x-bind.spec.js
@@ -580,3 +580,30 @@ test('option element value attribute is removed when bound value is null or unde
         get('option:nth-of-type(2)').should(haveValue('Undefined Label'))
     }
 )
+
+test(':value on a custom element re-syncs when bound object data mutates',
+    html`
+        <script>
+            if (! customElements.get('value-display')) {
+                customElements.define('value-display', class extends HTMLElement {
+                    get value() { return this._value }
+                    set value(v) {
+                        this._value = v
+                        this.textContent = JSON.stringify(v)
+                    }
+                })
+            }
+        </script>
+
+        <div x-data="{ data: [{ label: 'Mon', value: 100 }] }">
+            <value-display :value="data"></value-display>
+
+            <button @click="data[0].value = 999">Mutate</button>
+        </div>
+    `,
+    ({ get }) => {
+        get('value-display').should(haveText('[{"label":"Mon","value":100}]'))
+        get('button').click()
+        get('value-display').should(haveText('[{"label":"Mon","value":999}]'))
+    }
+)


ekwoka:
This is curious, since, of course, normal attributes wouldn't be allowed to be objects, but it is definitely what someone might expect for this kind of thing, especially coming from less "web-native" frameworks.

The alternative would be forcefully stringifying it into the attribute...which would then require the element to parse that...

But also the fact it's a reactive object being passed could cause other issues just passing the proxy as is...but it would play nice with mutatibility, and leave making it `raw` up to the end user, which would be simpler than the inverse of making it raw and trying to make the proxy pass be in user land...

Question: Does this same issue occur for attributes other than `value` or just for `value`? Does this change make `value` behave the same as other attributes?

I assume this change doesn't change the behavior of assigning to `value` on normal elements, since the assignment doesn't actually change at all, so it gets turned into a string tag...

joshhanley:
Yeah Flux has `value` defined as a property on the custom elements, hence why they are treated as data instead of an attribute value. I had initially implemented a `JSON.stringify` check in Livewire for this, but Caleb thought it might be able to be fixed upstream in Alpine without the stringification and this is what I came up with.

This only impacts `value` and nope, shouldn't change assigning on normal elements.

ekwoka (approved):
Makes sense

calebporzio:
thanks!

PR #4836, opened 2026-05-27
Add x-data to $dispatch documentation example

Manan2245:
The $dispatch example can be confusing when copied directly because Alpine directives and magic helpers require an Alpine component scope.

Without x-data, the example may not work as expected in a standalone setup.

This PR adds x-data to the example to make it function correctly when copied from the documentation.
Documentation link: https://alpinejs.dev/essentials/events#dispatching-custom-events

### DIFF
--- a/packages/docs/src/en/essentials/events.md
+++ b/packages/docs/src/en/essentials/events.md
@@ -71,7 +71,7 @@ In addition to listening for browser events, you can dispatch them as well. This
 Alpine exposes a magic helper called `$dispatch` for this:
 
 ```alpine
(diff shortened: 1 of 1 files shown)


joshhanley:
@Manan2245 thanks for the PR! I'm going to close this, because we don't like adding `x-data` to the examples directly. See this PR https://github.com/alpinejs/alpine/pull/4307 for more details. Feel free to submit a PR that adds a note like that PR, if we are missing one on that documentation page. Thanks for being willing to contribute!

PR #4849, opened 2026-06-24
Upgrade Cypress

joshhanley:
# The Scenario

On Apple Silicon Macs without Rosetta installed, Alpine's Cypress test suite cannot run with the current Cypress 7 dependency.

Cypress installs an x64 macOS binary, then fails during the smoke test with a bad CPU type error.

```sh
Command failed with Unknown system error -86: /Users/josh/Library/Caches/Cypress/7.7.0/Cypress.app/Contents/MacOS/Cypress --no-sandbox --smoke-test
spawn Unknown system error -86
```

# The Problem

Alpine currently depends on Cypress `^7.0.0`, which resolves to Cypress `7.7.0`.

Cypress 7 does not provide a native macOS ARM binary. On Apple Silicon machines without Rosetta, the installed x64 Cypress binary cannot be executed.

Cypress 8 and 9 have the same limitation, so the branch needs to continue through to Cypress 10.2 or newer to get native Apple Silicon support.

# The Solution

Upgrade Cypress from 7 to 10 on this branch.

The commits are split by major version so CI can validate each step independently. Cypress was upgraded from 7 to 8 first, then from 8 to 9, then from 9 to 10.

The Cypress 10 config migration is included with the Cypress 10 commit, keeping each step focused on the compatibility work required for that major.


calebporzio:
this ready to get out of draft and merged?

joshhanley:
@calebporzio not quite. That one test kept failing, so it's the only test enabled in CI at the moment. I'll finishing it today, then I will merge it.

PR #4852, opened 2026-07-01
Fix `x-mask` attribute changes on `x-model` inputs

### LINKED ISSUE #10377: Dynamic `x-mask` broken when used together with `wire:model`
### Livewire version

v4.3.3

### Laravel version

v13.18.0

### PHP version

8.5.7

Chrome on macOS

When dynamically determining `x-mask` for a text input, the mask fails to work properly when the field also has `wire:model` and `x-mask` changes inbetween renders.

Considering the example below:

1. Starting with country 'FR' and postal code '12345' in a field masked '99999'
2. Change the country to 'BE' and the postal code changes to '1234' adhering to the mask '9999'. The mask still works fine.
3. Change the country back to 'FR' (check your web developer console to ensure that `x-mask` has been updated again) and type a number in the input. It won't show up. If you type the same number again, it will show up. If you type another number, the input is still limited to 4 numbers (instead of the expected 5).

- This behaviour doesn't occur if the field doesn't have `wire:model`.
- Bissecting shows this bug was introduced between v4.2.2 and v4.2.3 (presumably the Alpine update from v3.15.8 to v3.15.9)

```php
<?php

use Livewire\Attributes\Computed;
use Livewire\Component;

new class extends Component {
    public string $country = 'FR';
    public string $postalCode = '12345';

    #[Computed]
    public function mask()
    {
        return $this->country === 'FR' ? '99999' : '9999';
    }
};

?>
```
```blade
<form wire:submit="save">
    <select wire:model.live="country">
        <option value="FR">France</option>
        <option value="BE">Belgium</option>
    </select>
    <input wire:model="postalCode" x-mask="{{ $this->mask }}">
    <button type="submit">Save</button>
</form>
```

_No response_

`x-mask` is supposed to work properly after changing between renders

joshhanley:
# The Scenario

When Livewire morphs an input that uses both `wire:model` and `x-mask`, changing the server-rendered `x-mask` attribute can leave the input using a previous mask.

In this example, the input starts with a five-digit French postal code mask, changes to a four-digit Belgian mask, then changes back to the five-digit French mask. The `x-mask` attribute updates, but the input remains limited by the older four-digit mask.

```php
use Livewire\Attributes\Computed;
use Livewire\Component;

new class extends Component {
    public string $country = 'FR';
    public string $postalCode = '12345';

    #[Computed]
    public function mask()
    {
        return $this->country === 'FR' ? '99999' : '9999';
    }
};
```

```blade
<form>
    <select wire:model.live="country">
        <option value="FR">France</option>
        <option value="BE">Belgium</option>
    </select>

    <input wire:model="postalCode" x-mask="{{ $this->mask }}">
</form>
```

# The Problem

The mask directive wraps `el._x_forceModelUpdate` so programmatic model updates are formatted before they are applied to the input.

When an existing `x-mask` attribute changes, Alpine treats that as directive cleanup followed by directive initialisation. The cleanup aborts the directive's event listeners, but it does not restore `el._x_forceModelUpdate`.

That means every reinitialised mask wraps the updater currently on the element. Previous masks stay in the update chain, so an older four-digit mask can still trim a value after the element has changed back to a five-digit mask.

# The Solution

The solution is to let the mask directive unregister the model updater wrapper it installs.

The directive now keeps a reference to its wrapper and restores the previous updater during cleanup, but only when the element is still using that exact wrapper. This avoids removing a newer wrapper that may have been installed after another directive initialisation.

Fixes livewire/livewire#10377

### DIFF
--- a/packages/mask/src/index.js
+++ b/packages/mask/src/index.js
@@ -3,8 +3,12 @@ export default function (Alpine) {
     Alpine.directive('mask', (el, { value, expression }, { effect, evaluateLater, cleanup }) => {
         let templateFn = () => expression
         let lastInputValue = ''
+        let undoModelUpdate = () => {}
+        let cleanedUp = false
 
         queueMicrotask(() => {
+            if (cleanedUp) return
+
             if (['function', 'dynamic'].includes(value)) {
                 // This is an x-mask:function directive.
 
@@ -51,7 +55,7 @@ export default function (Alpine) {
                 }
 
                 let updater = el._x_forceModelUpdate
-                el._x_forceModelUpdate = (value) => {
+                let update = (value) => {
                     // If the model value was cleared (e.g. the parent object was
                     // removed), just clear the input — don't format and write back
                     // as that would resurrect the model path with an empty value.
@@ -69,13 +73,23 @@ export default function (Alpine) {
                     updater(value)
                     el._x_model.set(value)
                 }
+
+                el._x_forceModelUpdate = update
+
+                undoModelUpdate = () => {
+                    if (el._x_forceModelUpdate === update) {
+                        el._x_forceModelUpdate = updater
+                    }
+                }
             }
         })
 
         const controller = new AbortController()
 
         cleanup(() => {
+            cleanedUp = true
             controller.abort()
+            undoModelUpdate()
         })
 
         el.addEventListener('input', () => processInputValue(el), {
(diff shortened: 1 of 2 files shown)


calebporzio:
Traced this through `mutation.js` (removed-attrs cleanup runs before added-attrs init, so the wrapper chain stays depth-1) and verified the Cypress test fails on main and passes with the fix. The identity guard on restore handles the x-model-reinit case nicely, and dropping the `cleanedUp` microtask guard is safe since any dead wrapper gets clobbered by x-model re-init. Ready to merge.

PR #4865, opened 2026-07-28
Fix stale effects after `x-data` replacement

### LINKED ISSUE #4864: x-text stops reacting to data changes after Livewire morph (data stays correct, binding goes dead)
### Description

I'm seeing an intermittent bug where an `x-text`-bound element stops reacting to changes in its own component's data after a Livewire morph, while the underlying reactive data itself remains correct. It happens inside a Livewire v3 app, so the DOM updates are driven by Livewire's `morph2()` (which wraps `Alpine.morph()` + `Alpine.transaction()`), not `x-if`/`x-for`.

Setup: a small Alpine component (`x-data`) with a `select(id, value)` method that sets `selectedId`/`selectedValue`, and a `<div x-text="selectedValue != null ? selectedValue : placeholder">` inside a `<button>`. The component root has a `wire:key` that stays stable across ordinary selections (it only changes when a separate, unrelated field changes server-side).

### Symptom

After enough interaction (typically: select an option, wait for the resulting Livewire round-trip/morph to settle showing the correct value, then select a different option), the button's displayed text sometimes stops updating and stays on the previous value, permanently, not just for one tick. Confirmed via direct inspection, not just visual:

```js
const el = document.querySelector('[wire\\:key^="dropdown-sizeId"]');
const data = Alpine.$data(el);
data.selectedValue; // "12" - correct, reactive data updated fine
el.querySelector("button div").textContent; // "3" - stale

// proving it's not just a stale read - the binding itself is dead:
data.selectedValue = "PROBE";
el.querySelector("button div").textContent; // still "3", no reaction at all
```

The element still has registered effects (`el.querySelector('button div')._x_effects.size === 2`), and manually invoking them (`_x_runEffects()`) does correctly write the current value - so the effect closure itself is fine, but it's no longer being triggered by writes to the reactive property it's supposedly watching.

### What I ruled out

- Not stale data - `Alpine.$data()` is correct throughout.
- Not something already fixed - Version is Alpine v3.15.12 (bundled with Livewire v3.8.1, the latest release on Livewire's still-maintained v3.x branch). This already includes the reactive-effect-transactions work in #4731 and the morph/`$wire` fixes in livewire/livewire#9951. Livewire v4.3.3 (the latest overall) bundles the same Alpine v3.15.12, so this isn't specific to the v3.x branch.
- Not our own app-specific mitigations - I tried debouncing the client → server save call (to reduce morph frequency) and forcing a full rebuild via a dynamic `wire:key` suffix keyed to a related field; neither prevented the desync.
- Tried to isolate in a minimal `jsdom` harness using the exact bundled `alpinejs@3.15.12` + `@alpinejs/morph@3.15.12`, mirroring Livewire's `Alpine.transaction(() => Alpine.morph(...))` wrapping and a direct `Alpine.cloneNode(fromEl, toEl)` call on the matched label-div pair (since that's what `context.patch()` does for every matched node during a morph, and by the time it reaches a nested element its `to` ancestor already shares the same live `_x_dataStack` as `from` - so the throwaway effect created by `cloneNode`'s `dontRegisterReactiveSideEffects` reads the _same_ reactive property as the real persistent effect). Neither overlapping transactions nor the direct `cloneNode` collision reproduced the sticky variant in isolation, only a transient one-tick lag that self-heals. So the real trigger seems to depend on something about the larger tree / nested component structure / real network timing that I haven't isolated yet.

### Versions

- alpinejs: 3.15.12 (via livewire/livewire v3.8.1's bundled build, also present in v4.3.3)
- @alpinejs/morph: 3.15.12
- Laravel: 12, PHP 8.4

### What I'm asking

I don't have a minimal, deterministic reproduction yet - this writeup is the evidence I have have, in case the symptom (an effect surviving as a registered-but-no-longer-triggered subscription after a morph) rings a bell, or points at where else to look. Happy to instrument further / share more of our app-level repro steps if useful.

joshhanley:
# The Scenario

When a Livewire update changes values rendered into a keyed Alpine component's `x-data` expression, the component state resets correctly but a descendant `x-text` binding can continue displaying its previous value.

In this example, selecting the first option displays `One`. Replacing the available options causes Livewire to morph the existing element with a new `x-data` expression. Although `selectedValue` resets to `null`, the label can remain stuck on `One` and ignore subsequent changes to the current scope.

```blade
<?php

use Livewire\Component;

new class extends Component {
    public array $options = ['One', 'Two'];

    public function replaceOptions(): void
    {
        $this->options = ['Three', 'Four'];
    }
};

?>

<div>
    <button wire:click="replaceOptions">
        Replace options
    </button>

    <div
        wire:key="options"
        x-data="{
            options: {{ Js::from($options) }},
            selectedValue: null,
        }"
    >
        <button
            type="button"
            x-on:click="selectedValue = options[0]"
        >
            Select first option
        </button>

        <span x-text="selectedValue ?? 'No option selected'"></span>
    </div>
</div>
```

# The Problem

When Alpine observes a changed `x-data` attribute, it cleans up the existing directive and initialises the replacement.

The cleanup removes the original reactive object from the component's data stack. The replacement expression is then wrapped in a new reactive object.

Descendant directives are not reinitialised because their elements and attributes have not changed. Their evaluators still reference the original reactive object, while `Alpine.$data()` returns the replacement. Effects therefore remain registered but no longer react to writes made through the component's current scope.

Repeated attribute changes can also initialise the final expression more than once, producing duplicate scope entries.

# The Solution

The solution is to update the existing reactive `x-data` object instead of replacing it with a new one.

This keeps descendant bindings connected to the same object they were originally watching. Alpine applies values from the new `x-data` expression to that object and removes properties that are no longer present.

Getters and setters are copied without converting them into plain values.

Alpine now distinguishes between replacing an `x-data` attribute and removing it entirely. Replacements preserve the reactive object, while removals continue to destroy and release the component normally.

Multiple changes to the same attribute in one mutation batch are combined so the final expression is only initialised once.

Effects created by `$persist` are released when their owning `x-data` expression is replaced. This prevents old persistence effects from accumulating.

PR #4759 previously explored updating the existing reactive object in place. This implementation was developed independently and differs in several important ways.

It only retains the reactive object during a confirmed attribute replacement. Removing `x-data` does not leave the old component data attached to the element.

It also preserves getters and setters, removes obsolete lifecycle properties, prevents duplicate scopes during repeated mutations, and cleans up interceptor effects created by `$persist`.

Related to PR #4759
Fixes #4864

### DIFF
--- a/packages/alpinejs/src/directives.js
+++ b/packages/alpinejs/src/directives.js
@@ -119,7 +119,7 @@ export function getElementBoundUtilities(el) {
         evaluate: evaluate.bind(evaluate, el),
     }
 
-    let doCleanup = () => cleanups.forEach(i => i())
+    let doCleanup = (...args) => cleanups.forEach(i => i(...args))
 
     return [utilities, doCleanup]
 }
--- a/packages/alpinejs/src/directives/x-data.js
+++ b/packages/alpinejs/src/directives/x-data.js
@@ -10,6 +10,8 @@ import { evaluate } from '../evaluator'
 
 addRootSelector(() => `[${prefix('data')}]`)
 
+let dataToReconcile = new WeakMap
+
 directive('data', ((el, { expression }, { cleanup }) => {
     if (shouldSkipRegisteringDataDuringClone(el)) return
 
@@ -27,18 +29,46 @@ directive('data', ((el, { expression }, { cleanup }) => {
 
     injectMagics(data, el)
 
-    let reactiveData = reactive(data)
+    let reactiveData = dataToReconcile.get(el)
 
-    initInterceptors(reactiveData)
+    dataToReconcile.delete(el)
+
+    if (reactiveData) {
+        Object.keys(data).forEach(key => {
+            let descriptor = Object.getOwnPropertyDescriptor(data, key)
+            let existingDescriptor = Object.getOwnPropertyDescriptor(reactiveData, key)
+
+            if (descriptor.get || descriptor.set || existingDescriptor?.get || existingDescriptor?.set) {
+                if (existingDescriptor) delete reactiveData[key]
+                if (! existingDescriptor) reactiveData[key] = undefined
+
+                descriptor.get || descriptor.set
+                    ? Object.defineProperty(reactiveData, key, descriptor)
+                    : reactiveData[key] = data[key]
+            } else {
+                reactiveData[key] = data[key]
+            }
+        })
+
+        Object.keys(reactiveData)
+            .filter(key => ! Object.prototype.hasOwnProperty.call(data, key))
+            .forEach(key => delete reactiveData[key])
+    } else {
+        reactiveData = reactive(data)
+    }
+
+    initInterceptors(reactiveData, cleanup)
 
     let undo = addScopeToNode(el, reactiveData)
 
     reactiveData['init'] && evaluate(el, reactiveData['init'])
 
-    cleanup(() => {
+    cleanup((isReplaced) => {
         reactiveData['destroy'] && evaluate(el, reactiveData['destroy'])
 
         undo()
+
+        if (isReplaced) dataToReconcile.set(el, reactiveData)
     })
 }))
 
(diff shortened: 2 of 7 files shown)


joelwmale (approved):

calebporzio:
closing in favor of this PR: https://github.com/alpinejs/alpine/pull/4868

PR #4881, opened 2026-08-23
Fix Unicode escapes in CSP string literals

### LINKED ISSUE #4879: Can't pass an escaped string literal into a component method with CSP version
### Alpine.js version

v3.16.1

Firefox on Linux

When passing an escaped string literal into an alpine component method, it is not parsed in the CSP version of alpine, but is parsed in the regular version.

The below example demonstrates this with two buttons: one that passes `test` and the other that passes `tes\u0027t`:

- Passing `test` works fine for both versions.
- Passing `tes\u0027t` results in `tes't` in the regular version, but `tesu0027t` in the CSP version.

This is particularly problematic when using laravel's `Js::from()` such as `x-data="component({{ Js::from($test) }})"`function, as it escapes apostrophes and other common symbols which get misread inside the component.

```html
<!--<script src="https://unpkg.com/alpinejs" defer></script>-->
<script defer src="https://cdn.jsdelivr.net/npm/@alpinejs/csp@3.x.x/dist/cdn.min.js"></script>

<div x-data="test">
    <button x-on:click="set('test')">Without apostrophe (`test`)</button>
    <button x-on:click="set('tes\u0027t')">With apostrophe (`tes\u0027t`)</button>

    <input x-ref="input" />
</div>

<script>
    Alpine.data('test', () => ({
      set(value) {
        this.$refs.input.value = value;
      }
    }))
  })
</script>
```

_No response_

I expect both the CSP and the non-CSP versions to handle escapes within literal strings passed into component constructors and methods.

dawNotPoi:
Fixes #4879.

The CSP tokenizer now decodes four-digit Unicode escapes inside string literals. This makes escaped values such as `tes\\u0027t` match the regular Alpine build while preserving the existing handling for other escapes.

Added a parser regression assertion for the escaped apostrophe case reported in the issue.

Testing:

- `npx vitest run tests/vitest/csp-parser.spec.js` (100 tests passed)
- `node --check packages/csp/src/parser.js`

### DIFF
--- a/packages/csp/src/parser.js
+++ b/packages/csp/src/parser.js
@@ -139,6 +139,18 @@ class Tokenizer {
                     case 'r': value += '\r'; break;
                     case '\\': value += '\\'; break;
                     case quote: value += quote; break;
+                    case 'u': {
+                        const codeUnit = this.input.slice(this.position + 1, this.position + 5);
+
+                        if (/^[0-9a-fA-F]{4}$/.test(codeUnit)) {
+                            value += String.fromCharCode(parseInt(codeUnit, 16));
+                            this.position += 4;
+                        } else {
+                            value += char;
+                        }
+
+                        break;
+                    }
                     default: value += char;
                 }
                 escaped = false;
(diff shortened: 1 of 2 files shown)


calebporzio:
thanks!

PR #4883, opened 2026-08-26
Docs: clarify how Alpine effects track reactive dependencies

MuslumYilmaz:
## Summary

- Clarify that `Alpine.effect()` collects dependencies from reactive property reads, while later changes trigger the effect.
- Explain the synchronous dependency-collection boundary with a `setTimeout` example.
- Update the stale `vuejs/vue-next` source link to the current `vuejs/core` repository.
- Remove the duplicated `alpine-effect` anchor and correct a small possessive typo.
- Add an optional conceptual explanation of the Vue reactivity model Alpine inherits.

## Why

The current wording describes both reads and writes as tracked interactions. This can obscure the distinction between dependency collection and triggering.

The proposed explanation connects that distinction to the existing counter example and documents an important boundary: reactive reads performed after an effect callback has returned are not dependencies of that effect.

## Disclosure

I maintain FrontendAtlas, which hosts the external learning resource proposed in the final paragraph. I have no affiliation with Alpine.js or Vue.

I included the link because it explains the proxy, dependency-map, and `track`/`trigger` model used by Vue's reactivity system. The documentation clarification and official source-link correction stand independently, and I am happy to remove the third-party link if external learning resources are outside the documentation policy.

## Verification

Not run; this is a documentation-only Markdown change.

### DIFF
--- a/packages/docs/src/en/advanced/reactivity.md
+++ b/packages/docs/src/en/advanced/reactivity.md
@@ -10,7 +10,7 @@ Alpine is "reactive" in the sense that when you change a piece of data, everythi
 Every bit of reactivity that takes place in Alpine, happens because of two very important reactive functions in Alpine's core: `Alpine.reactive()`, and `Alpine.effect()`.
 
 > Alpine uses VueJS's reactivity engine under the hood to provide these functions.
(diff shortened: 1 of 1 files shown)


calebporzio:
makes sense 👌

PR #4884, opened 2026-08-26
Add dwell timing to the intersect plugin

cyppe:
## Problem

`x-intersect` evaluates as soon as an element crosses its configured threshold. View qualification and similar interactions sometimes need the threshold state to remain continuous for a short interval, which currently requires each application to maintain its own observer, timer, page-visibility, and cleanup logic.

## Approach

Add a `.dwell` modifier to the intersect plugin:

```html
<div x-intersect.half.dwell.500ms.once="trackView()"></div>
```

- waits 250 milliseconds by default and accepts an adjacent millisecond duration
- composes with `.half`, `.full`, custom thresholds, margins, `.parent`, and `.once`
- treats `:leave` as continuously below the configured threshold
- cancels pending dwell when the threshold breaks or the page becomes hidden
- reconciles queued observations and requires a fresh observation before evaluating, so stale entries cannot complete a dwell
- leaves the existing non-dwell path unchanged

## Validation

- focused intersect Cypress spec: 17 passed
- complete Cypress suite passed in Chromium 151
- Vitest: 185 passed, 1 skipped
- production build passed
- standalone intersect CDN bundle: +1,058 minified bytes / +365 Brotli bytes

### DIFF
--- a/packages/docs/src/en/plugins/intersect.md
+++ b/packages/docs/src/en/plugins/intersect.md
@@ -110,6 +110,35 @@ Sometimes it's useful to evaluate an expression only the first time an element e
 <div x-intersect.once="shown = true">...</div>
 ```
 
+<a name="dwell"></a>
+### .dwell
+
+Sometimes an element should remain in view for a short period before the expression evaluates. The `.dwell` modifier waits for 250 milliseconds of continuous intersection by default:
+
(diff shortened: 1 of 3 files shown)


calebporzio:
Great! ty!

PR #4897, opened 2026-09-08
Change x-for to allow loops like `(foo)in(bar)`

ColonelThirtyTwo:
Currently, the loop in x-for must have a space between the in/of and both the bindings and the source expression. However, JS doesn't require the space - only that they are separate tokens - so `(item, index) in list` may be minified to `(item,index)in list` which `x-for` no longer parses.

This patch alters the regular expression in `parseForExpression` to split the bindings and expression on word boundries (`\b`) rather than spaces, to support this syntax.


ekwoka:
I get it...but like why?

ColonelThirtyTwo:
In my app, I have a Rust macro that runs some basic minification over passed-in javascript, and it minifies `(foo, bar) in baz` to `(foo,bar)in baz` since it's technically legal. I could special case it, but I figure since it was a fairly straightforward change to Alpine to accept it, I'd submit it as a PR.

PR #4908, opened 2026-09-14
Fix mutation observer recovery after errors

### LINKED ISSUE #4907: mutateDom stops observing future mutations when its callback throws
### Alpine.js version

v3.17.3

Chrome on macOS (the issue is browser-independent)

`Alpine.mutateDom()` stops the global mutation observer before invoking its callback and restarts it only after the callback returns. If that callback throws synchronously, control skips the restart call and Alpine remains permanently disconnected from DOM mutations.

After the error is caught, dynamically inserted Alpine components are therefore never initialized. This can also amplify unrelated directive errors: one exception from code running inside `mutateDom()` prevents Alpine from observing all subsequent DOM changes.

The current implementation in `packages/alpinejs/src/mutation.js` is effectively:

```js
stopObservingMutations()

let result = callback()

startObservingMutations()
```

Because observation is global state, it should be restored even when the callback fails.

```html
<div id="target"></div>

<script type="module">
    import Alpine from 'https://cdn.jsdelivr.net/npm/alpinejs@3.17.3/+esm'

    window.Alpine = Alpine
    Alpine.start()

    try {
        Alpine.mutateDom(() => {
            throw new Error('expected test error')
        })
    } catch (_) {}

    document.querySelector('#target').innerHTML = `
        <div x-data="{ message: 'initialized' }">
            <span id="result" x-text="message"></span>
        </div>
    `

    setTimeout(() => {
        console.log(document.querySelector('#result').textContent)
        // Actual: ""
        // Expected: "initialized"
    })
</script>
```

Steps:

1. Start Alpine.
2. Call `Alpine.mutateDom()` with a callback that throws, and catch the expected error.
3. Insert a new `x-data` component into the document.
4. Observe that its directives are never initialized.

Not applicable.

`mutateDom()` should preserve the callback's existing return and error behavior while always restoring mutation observation. A `try`/`finally` around the callback would ensure that one failed DOM mutation does not disable Alpine's handling of future DOM changes.

BillyNoyes:
## What changed

- Restore Alpine's mutation observer in a `finally` block when a `mutateDom` callback throws.
- Preserve the callback's existing return value and error propagation behavior.
- Add a browser regression test proving later DOM additions are still initialized after an error.

Fixes #4907.


BillyNoyes (commented):
My agent review looks good

jeffwhansen:
Thanks for this fix

PR #4910, opened 2026-09-15
Add viewport modifier to resize plugin

BillyNoyes:
## What

Add an `x-resize.viewport` modifier that reports the visual viewport dimensions and reacts to `window.visualViewport` resize events. This keeps viewport-aware layouts responsive to mobile browser controls and on-screen keyboards without adding visual viewport behavior to the generic `x-on` directive.

When the Visual Viewport API is unavailable, the modifier falls back to independently observing the document element so every subscriber receives an initial measurement.

This follows the direction discussed in #4350 to house visual viewport resize handling in the Resize plugin.


calebporzio:
looks perfect! thanks!

PR #4916, opened 2026-09-21
Add reactive $media magic

BillyNoyes:
## What

Add a reactive `$media(query)` magic that reports whether a browser media query currently matches.

Alpine shares one `MediaQueryList` listener for identical queries, updates dependent expressions when the match changes, and removes the listener after its final consuming element is cleaned up. The implementation also supports the legacy `MediaQueryList` listener API.

## Why

CSS media queries should remain the default for responsive styling and visibility. This proposal is for media-query-dependent **JavaScript behavior**: focus management, component state, timers, and other effects that CSS cannot control.

Today, that requires creating a `MediaQueryList`, copying its initial match into Alpine data, subscribing to changes, and cleaning up the listener. `$media` makes the match available directly as a reactive dependency in Alpine expressions.

## Examples

These are integration examples, not complete components. They assume Alpine is initialized and any named data providers are registered before `Alpine.start()`.

### Trap focus only when navigation behaves as a mobile drawer

A navigation panel can behave as a modal drawer on smaller screens and an ordinary sidebar on larger screens. Within the navigation component's existing `x-data` scope, where `open` controls the drawer, the Focus plugin could use:

```html
<nav x-trap.noscroll="open && !$media('(min-width: 768px)')">
    <a href="/account">Account</a>
    <a href="/settings">Settings</a>
</nav>
```

While `open` is true, matching the desktop query makes the trap expression false, releasing the focus trap and scroll lock. Returning to a smaller viewport re-enables them if `open` is still true. CSS can handle layout and visibility, but CSS rules alone cannot enable or disable the Focus plugin.

This requires `@alpinejs/focus`. The panel must be visible while trapped; its trigger, dismissal controls, and accessibility semantics remain the existing component's responsibility. `$media` itself does not depend on the Focus plugin.

### Respect reduced motion in JavaScript-driven autoplay

A carousel component could react to changes in the user's reduced-motion preference:

```html
<div
    x-data="carousel"
    x-effect="autoplayEnabled && !$media('(prefers-reduced-motion: reduce)')
        ? startAutoplay()
        : stopAutoplay()"
>
    <!-- Carousel markup -->
</div>
```

Here, `carousel` is an application-defined component providing `autoplayEnabled`, `startAutoplay()`, and `stopAutoplay()`. Its start method must avoid creating duplicate timers, its stop method must tolerate an already-stopped timer, and its `destroy()` should stop any remaining timer. `$media` only supplies the reactive query result; it does not implement those methods.

The effect runs initially and reacts to the preference and `autoplayEnabled`. Reduced motion stops autoplay; removing that preference allows autoplay only if `autoplayEnabled` remains true. An explicit pause represented by setting that boolean to false is therefore preserved. CSS can disable CSS animations, but CSS rules alone cannot stop a JavaScript timer.

### Reset transient state when leaving a mobile layout

Consider a mobile navigation drawer whose visibility is controlled by `mobileMenuOpen`. CSS can hide the drawer at a desktop breakpoint, but the boolean remains `true`. Returning to a narrow viewport can then reopen the drawer when the intended behavior is to return with it closed.

On the existing navigation component, an effect can reset that state when the desktop query matches:

```html
<div x-effect="if ($media('(min-width: 768px)')) mobileMenuOpen = false"></div>
```

This runs within the component's existing `x-data` scope. It checks the query immediately and again when the match changes. Entering the desktop layout resets the flag; returning to mobile does not reopen it. It is a breakpoint-driven reset, not a rule that continuously prevents other code from setting the flag while already on desktop. Layout and visibility stay in CSS.

### Mount an optional expensive enhancement only when it is useful

A product comparison page might always show an accessible data table, with an additional interactive chart for wider layouts:

```html
<div x-data>
    <template x-if="$media('(min-width: 1024px)')">
        <div x-data="comparisonChart">
            <canvas x-ref="chart" aria-hidden="true"></canvas>
        </div>
    </template>
</div>
```

Here, `comparisonChart` is an application-defined Alpine component. Its `init()` creates the chart and its `destroy()` releases the chart instance and any subscriptions. The always-available table contains the same information; the supplementary canvas is hidden from assistive technology to avoid duplicating it.

Below the breakpoint, the chart component is not mounted and its `init()` does not run. Crossing above it mounts the component; crossing back below removes it through `x-if` and invokes its `destroy()` hook. Hiding an already-mounted chart with CSS does neither.

This defers component initialization, not necessarily downloading or parsing the chart library. Code-splitting would be a separate application concern, as would cancelling any asynchronous work started by the chart.

## Scope

The change adds the magic, documentation, and Cypress coverage. It does not introduce breakpoint aliases, a media-query parser, or component-specific behavior; queries are passed to the browser's `matchMedia()` API.


browner12:
Do you have another use case that isn't easily handled by CSS media queries?

BillyNoyes:
Good point @browner12, the original visibility example is better handled by CSS. I’ve updated the description to focus on JavaScript behavior instead

The proposed benefit is exposing its result as a reactive Alpine dependency, with shared listeners and element-lifecycle cleanup

ekwoka:
I made one for this before https://github.com/ekwoka/newthink-ui/blob/main/src/magics/screen.ts

I think you can just forgo actually cleaning up the MatchMedia and listeners since it just is so little work in the first place, that the effort of managing and cleaning up the stuff is more troublesome than it's even worth.

Not that my example is the best implementation in other ways (like a Map is better than an object).

Supporting all media is also nice for a more general plugin

BillyNoyes:
Thanks @ekwoka, that’s useful precedent! I can see the argument for keeping one shared listener per query for the page’s lifetime, especially when the same handful of queries are reused.

My reasoning for cleanup was to avoid retaining one-off queries after their consuming elements are removed, rather than because the listeners themselves are expensive. I’m happy to simplify that if a page-lifetime cache is the preferred tradeoff.

Supporting arbitrary media queries was deliberate, so the same API covers reduced motion, pointer capabilities, and other preferences without introducing breakpoint configuration.

Do you see the general `$media(query)` API as a good fit for core?

